PR #8, opened 2017-11-05
Minor changes to improve integration

Mayankm96:
* Added a liquid tag for google-analytics
* The previous instruction to set the baseurl to make `<username>.github.io` was not working when I tried out. However leaving it black does the charm.


alshedivat:
Looks good to me. Thanks for the contribution!

PR #15, opened 2018-06-11
Alignfix

samfearn:
There's a bug in the `_pages/_about.md` file, in that there's a `profile: align:` property defined in the front matter, but changing it doesn't do anything. I realised it's because there's an `align: right !important;` line in the `_sass/_profile.scss` file, which shouldn't be there. The `_layout/about.html` file creates a div for the profile image with a class given by the `profile: align:` property of the about.md page, and these classes are defined in `_sass/_positions.scss` simply to set the alignment, so the alignment should be set from this class, rather than separately in the profile class. If one changes the `profile: align:` property to left instead of right as it is in the template file, then nothing happens, as the `important` property of profile overrides the left class. If one just removes the alignment from the profile class, then there's still an error, as the col class in `_sass/_gallery.scss` is also defined to be left aligned, and as the profile image is also set to have the col class, then there is still a conflict. My fix is therefore to also remove the left align property from the col class seeing as a left/right class exists which makes all alignment explicit. Finally, this requires the project pages to have the images also have the left class in order to get the nice 2/1 or 1/1/1 images. I also updated the code snippet in the projects to match the changed classes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,12 +39,16 @@ By default, the script uses the `master` branch for the source code and deploys 
 The optional flag `--user` tells it to deploy to `master` and use `source` for the source code instead.
 Using `master` for deployment is a convention for [user and organization pages](https://help.github.com/articles/user-organization-and-project-pages/).
 
-**Note:** when deploying your user or organization page, make the `_config.yml` has `url` and `baseurl` fields as follows.
+**Note:** when deploying your user or organization page, make sure the `_config.yml` has `url` and `baseurl` fields as follows.
 
 ```
 url: # should be empty
 baseurl:  # should be empty
 ```
+
+### Usage ###
+
+Note that `_pages/about.md` is built to index.html in the published site. There is therefore no need to have a separate index page for the project. If an index page does exist in the root directory then this will prevent `_pages/about.md` from being added to the built site.
 
 ## Features
 
@@ -100,4 +104,4 @@ Style improvements and bug fixes are especially welcome.
 
 ## License
 
-MIT
+The theme is available as open source under the terms of the [MIT License](https://opensource.org/licenses/MIT).
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ url: # should be empty
 baseurl:  # should be empty
 ```
 
-### Usage ###
+### Usage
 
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -39,13 +39,17 @@ By default, the script uses the `master` branch for the source code and deploys
...
 
-**Note:** when deploying your user or organization page, make the `_config.yml` has `url` and `baseurl` fields as follows.
+**Note:** when deploying your user or organization page, make sure the `_config.yml` has `url` and `baseurl` fields as follows.
 
 ```
 url: # should be empty
 baseurl:  # should be empty
 ```
 
+### Usage ###
```
alshedivat: Let's keep # symbols at the beginning of the header only.

README.md
```diff
@@ -100,4 +104,4 @@ Style improvements and bug fixes are especially welcome.
 
 ## License
 
-MIT
+The theme is available as open source under the terms of the [MIT License](https://opensource.org/licenses/MIT).
```
alshedivat: I think the 3-letter version was descriptive enough, but thanks for the update.

_config.yml
```diff
@@ -11,7 +11,7 @@ footer_text: >
   Photos from <a href="https://unsplash.com" target="_blank">Unsplash</a>.
 
 url:  # the base hostname & protocol for your site
-baseurl: /al-folio # the subpath of your site, e.g. /blog/
+baseurl: # the subpath of your site, e.g. /blog/
```
alshedivat: Please keep it `/al-folio`, so that I can use https://alshedivat.github.io/al-folio/ for the demo webpage.

alshedivat:
Thanks a lot for spotting the bugs and contributing!! Happy to merge after some minor fixes.

samfearn:
Hey, yes sorry I shouldn’t have had the readme or config files in this pull request. I’ll change those files back how you say later.

samfearn:
I've changed the markdown in the README to only use the hash symbol before the title name, and changed the baseurl property back in the config file as requested. I've left the license change as-is in order to match the default you get if you use the `jekyll new-theme` command. I can change it back to the simple 'MIT' string if you'd rather?

alshedivat:
Wonderful, thanks again for the contribution!

samfearn:
No worries, thanks for the lovely theme :)

PR #25, opened 2018-08-24
Do not delete CNAME file 

sahirbhatnagar:
The step in the deploy script that deletes almost all files except for _site/ was causing issues if you have a custom domain name. GitHub pages requires a CNAME, so I added this to the list of files not to delete.


alshedivat:
Thanks for the fix.

PR #39, opened 2019-02-11
Vary gallery row height and spacing with viewport

### LINKED ISSUE #11: On mobile, images crop instead of scale
The image gallery structure works great on wide devices, but does not scale well, as images are cropped to fit into a narrower box (to stay 3 wide) instead of scaling down or shifting to the next line. 

Here's an example of what three images that I have (on collections) look like on browser and mobile:

*on web (browser)*
screen shot 2018-01-19 at 16 41 50

*on mobile*
screen shot 2018-01-19 at 16 37 40

I think there are a couple of options.
1. modify the base image gallery such that it includes responsive square containers that "contain" the image -> shrink and grow
2. use a media query to make breaks between images for mobile

Perhaps other options?

monkeywithacupcake:
Move the img-row height and col padding based on the viewport size. This closes #11


alshedivat:
Awesome, thanks for the contribution!

Issue #26, opened 2018-09-09
Content jumps horizontally whenever browser adds a scrollbar


ayuroman:
Page content  'jumps' a little on Safari/Chrome/IE when navigating between short and long pages:
https://alshedivat.github.io/al-folio/blog/ vs https://alshedivat.github.io/al-folio/projects/

alshedivat:
What do you mean by "jumps when navigating?" When I switch between the two pages (in Chrome), the behavior seems normal, nothing jumps anywhere.

ayuroman:
![jump](https://user-images.githubusercontent.com/42931270/46260307-15a84a80-c4dc-11e8-8878-6373a9902fe4.jpg)

alshedivat:
No idea how to fix this.

samfearn:
Pull request #48 created to resolve this.

PR #48, opened 2019-07-18
Added css to take account of vertical scroll bar and avoid content jumps

samfearn:
Fixed issue #26, as explained [here](https://css-tricks.com/elegant-fix-jumping-scrollbar-issue/). I've added an overall left margin to the main html tag, so that margin:auto works as if the scrollbar is invisible. Since the header has a fixed position, I've also changed the left position of the header. The horizontal alignment of the header now doesn't change between 'short' and 'long' pages.

The other way to fix this would be to make the vertical scrollbar always appear, regardless of whether or not the content requires it. This would achieve the same visual consistency, but I think is much less aesthetically pleasing.

### DIFF
--- a/_sass/_base.scss
+++ b/_sass/_base.scss
@@ -1,4 +1,8 @@
 @charset "UTF-8";
+html{
+    margin-left: calc(100vw - 100%);
+    margin-right: 0;
+}
 
 html, body {
   height: auto;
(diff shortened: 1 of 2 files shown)


alshedivat:
@samfearn, this is an elegant fix! Thank you!

PR #63, opened 2019-11-01
Make google analytics conditional

g-oikonomou:
Thanks for sharing this work. I just started using it for g-oikonomou.github.io

This pull makes the inclusion of google analytics conditional.


_includes/hemline.html
```diff
@@ -25,6 +25,7 @@
 <link rel="stylesheet" href="{{ '/assets/css/academicons.min.css' | prepend: site.baseurl | prepend: site.url }}">
 
 <!-- Google Analytics -->
```
alshedivat: Please move the `<!-- Google Analytics -->` comment inside the if statement.

g-oikonomou: That's done now

PR #66, opened 2019-11-01
Enable preview images on social media

g-oikonomou:
This commit adds support for serving Open Graph meta tags inside the HTML head. When enabled, links to the page on social media will display a preview.

NB: twitter behaves somewhat differently and may require twitter card meta tags in addition to Open Graph.

https://ogp.me/

Draft for now: Works nicely on skype. The twitter validator likes it, but I failed to make it work inside a tweet. Need to also test a couple more sites/apps. Success/failure reports welcome!

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -15,6 +15,14 @@ baseurl: /al-folio # the subpath of your site, e.g. /blog/
 last_updated: # leave blank if you don't want to display last updated
 
 # -----------------------------------------------------------------------------
+# Open Graph
+# -----------------------------------------------------------------------------
+# Display links to the page with a preview object on social media.
+# To achieve this, change serve_og_meta to true and then provide the URL of the
+# preview image as the value of og_image.
+serve_og_meta: false # Include Open Graph meta tags in the HTML head
+og_image: # The Open Graph preview image
+# -----------------------------------------------------------------------------
 # Social integration
 # -----------------------------------------------------------------------------
 github_username: # your GitHub user name
--- a/_config.yml
+++ b/_config.yml
@@ -21,7 +21,7 @@ last_updated: # leave blank if you don't want to display last updated
 # To achieve this, change serve_og_meta to true and then provide the URL of the
 # preview image as the value of og_image.
 serve_og_meta: false # Include Open Graph meta tags in the HTML head
-og_image: # The Open Graph preview image
+og_image: # The site-wide (default for all links) Open Graph preview image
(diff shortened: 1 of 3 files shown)


g-oikonomou:
Is what it looks like on twitter

https://twitter.com/oikonomou_geo/status/1190072892593188864?s=20

_config.yml
```diff
@@ -14,6 +14,14 @@ url:  # the base hostname & protocol for your site
...
 last_updated: # leave blank if you don't want to display last updated
 
+# -----------------------------------------------------------------------------
+# Open Graph
+# -----------------------------------------------------------------------------
+# Display links to the page with a preview object on social media.
+# To achieve this, change serve_og_meta to true and then provide the URL of the
+# preview image as the value of og_image.
+serve_og_meta: false # Include Open Graph meta tags in the HTML head
+og_image: # The Open Graph preview image
```
alshedivat: If I understand correctly, `og_image` will be used for all pages of the website shared on social media, correct? This seems suboptimal for people who use al-folio for personal blogs, where each post is supposed to have a separate featured image.

Maybe there is a way to have `og_image` to be a page variable rather than the site variable? Or if the page does not have `og_image` specified, it could fall back to using `site.og_image`.

g-oikonomou: > If I understand correctly, og_image will be used for all pages of the website shared on social media, correct? This seems suboptimal for people who use al-folio for personal blogs, where each post is supposed to have a separate featured image.

Yes, that's good thinking. Not my use case so didn't occur to me. Let me have a stab at that, it shouldn't be a problem.

Arguably, we should do the same for some of the other fields too (e.g. send the blog page title to the preview, instead of sending the entire site title)?

g-oikonomou:
So:

* Added documentation in the README
* Added support for page-specific preview images as requested. Site-wide image used as fallback.
* `og:description` is now the page-specific `page.title`, instead of `site.name`. This assumes silently that the title will not contain HTML tags. If it does then we will likely end up serving invalid HTML, but I cannot think of a good way to work-around this other than ask the user to explicitly provide values for multiple OG properties. Which I don't think we want to do.
* `og:url` now points to the individual page linked, not simply to the site's home page.

This is a tad difficult to test, but I tested using:

* [Twitter card validator](https://cards-dev.twitter.com/validator) (need to be logged in twitter)
* [Open Graph Check](https://opengraphcheck.com/)

PR #81, opened 2020-03-22
Improve support for deploying in out-of-tree locations

g-oikonomou:
This pull improves our configuration in order to better support deploying a website to an out-of-tree location by using:

```
bundle exec jekyll build --destination <destination>
```
The proposed configuration changes do not harm (maybe they even improve) the current default approach of deploying in a separate branch of the `al-folio` repo.

The pull also documents this method in the `README.md`, explicitly flagging it as "for advanced users".

Motivation: We are currently assuming that users host the website's sources and the deployed version as different branches of the same repo. An alternative use-case is for the sources to be in one repo and the deployment version under a totally different repo. This could be the case for example if a project's website is hosted under a subdir of the project's main repo (for example a `docs` subdir under the `master` branch of the project's main repo).

In terms of `keep_files`, this pull makes the following assumptions about `<destination>`:
* For user github pages, it is almost certain that there will be a `.nojekyll` file in the top-level dir.
* For user github pages, a `CNAME` may be in use if the user is using a custom domain.
* It is very likely (almost certain) that `<destination>` will be a git-versioned repo. We therefore need to leave `.git/`, `.gitignore` alone.

In terms of the `exclude` list:
* We remove `vendor`, since there does not appear to be a file/directory with that name in al-folio's source tree
* We add `bin` and the `Gemfile`s, since they don't really need to be in the deployment location


alshedivat:
Great! Thanks for your contribution. I'm happy to merge once Travis CI build works.

alshedivat:
I'm happy to merge this PR. Could you please resolve conflicts with `master` (it was just upgraded to `v0.3.0`) and I'll take another look. Thanks!

g-oikonomou:
Rebased

PR #89, opened 2020-05-30
Document that blog_name changes the navbar

jkmacc-LANL:
I love the new design, but was confused why the blog link didn't show up on the nav bar.  It was because the blog needs a name for it to appear.  Just added a comment in the config that documents this behavior.  

Thank you for a great template!


alshedivat:
Great, thanks for the contribution!

PR #102, opened 2020-07-14
changes allow for more than first and last name to be displayed.  

jasongfleischer:
If more than 2 elements in _config.yml variable title, any middle names or initials will be displayed.

This doesn't handle things like a Jr. or III yet.  The lastname variable will always catch the last thing.  

An alternative would be adding  either
- a separate variable in _config.yml for name suffix.  
- Or perhaps revamp all this with separate config variables explicitly set for Personal name, Middle names, Family name, Suffix

### DIFF
--- a/_layouts/about.html
+++ b/_layouts/about.html
@@ -6,8 +6,16 @@ layout: default
 
   <header class="post-header">
     {% assign name = site.title | split: ' ' %}
-    <h1 class="post-title"><span class="font-weight-bold">{{ name | first }}</span> {{ name | last }}</h1>
-    <p class="post-description">{{ page.description }}</p>
+    {% assign firstname = name | first %}
+    {% if name.size > 2 %}
+       {% assign middlenames =  name | slice:1, name.size-1 | join:" " %}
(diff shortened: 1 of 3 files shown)


alshedivat:
Fantastic! Thanks again for the contribution!

_layouts/about.html
```diff
@@ -6,8 +6,16 @@
...
+    {% assign firstname = name | first %}
+    {% if name.size > 2 %}
+       {% assign middlenames =  name | slice:1, name.size-1 | join:" " %}
+    {% else %}
+       {% assign middlenames = '' %}
+    {% endif %}
+    {% assign lastname = name | last %}
+    <a class="navbar-brand title font-weight-lighter" href="{{ site.baseurl | prepend: site.url }}/">
+     <span class="font-weight-bold">{{ firstname }}</span> {{ middlenames }}  {{ lastname }}
+    </a>
```
alshedivat, line 18: Looks like this change substituted `<h1>` with `<a>` in the about page template, which is a bug.

PR #111, opened 2020-08-30
FIX add first/middle/last name var in config

### LINKED ISSUE #109: A bug in the about.html template introduced in #102
The post header defined in `about.html` must contain the first, middle, and last names within `<h1>` tags instead of an `<a>`. Also, the description is missing in the template. The bug was introduced in #102.

### LINKED ISSUE #110: Incorrectly handling last names with more than one word
**Acknowledge the following**
- [X] The issue I am raising is a potential bug in al-folio and not just a usage question. <br> [For usage questions, please use [gitter chat](https://gitter.im/alshedivat/al-folio) instead of raising an issue.]

**Describe the bug**
Hi, first of all, thanks for making & sharing this awesome theme. I'm excited to use it for my own website.

I think that the changes introduced by #102 do not correctly parse last names with more than one word (like mine, made up by four words). Clearly, there is no general way to automatically parse a name into first, middle, last name in a way that generalizes to all possible names. I would think that adding explicit variables in the config file would solve the issue.

This is the first time I'm using jekyll, but I will address this issue (because I really want to use this theme!). It will take me a while most likely, but I'm happy to submit a PR when I solve it.

**EDIT**

Short example showing the problem. Changing the name in `_config.yml` to "You R. Long Last Name" renders as follows

image

mvdoc:
OK well that was faster than expected.

Example rendering now

image

image

closes #110 
closes #109

### DIFF
--- a/_layouts/about.html
+++ b/_layouts/about.html
@@ -6,7 +6,7 @@ layout: default
 
   <header class="post-header">
     <h1 class="post-title">
-     {% if site.title %}{{ site.title }}{% else %}<span class="font-weight-bold">{{ site.first_name }}</span> {{ site.middle_name }}  {{ site.last_name }}{% endif %}
+     {% if site.title == blank %}{{ site.title }}{% else %}<span class="font-weight-bold">{{ site.first_name }}</span> {{ site.middle_name }}  {{ site.last_name }}{% endif %}
     </h1>
   </header>
 
--- a/_layouts/about.html
+++ b/_layouts/about.html
@@ -6,7 +6,7 @@ layout: default
 
   <header class="post-header">
     <h1 class="post-title">
-     <span class="font-weight-bold">{{ site.first_name }}</span> {{ site.middle_name }}  {{ site.last_name }}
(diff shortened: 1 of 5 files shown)


_layouts/about.html
```diff
@@ -5,17 +5,9 @@
...
-       {% assign middlenames =  name | slice:1, name.size-1 | join:" " %}
-    {% else %}
-       {% assign middlenames = '' %}
-    {% endif %}
-    {% assign lastname = name | last %}
-    <a class="navbar-brand title font-weight-lighter" href="{{ site.baseurl | prepend: site.url }}/">
-     <span class="font-weight-bold">{{ firstname }}</span> {{ middlenames }}  {{ lastname }}
-    </a>
+    <h1 class="post-title">
+     <span class="font-weight-bold">{{ site.first_name }}</span> {{ site.middle_name }}  {{ site.last_name }}
```
alshedivat, line 9:
```suggestion
     {% if site.title %}{{ site.title }}{% else %}<span class="font-weight-bold">{{ site.first_name }}</span> {{ site.middle_name }}  {{ site.last_name }}{% endif %}
```
Same here, could you apply the above change?

alshedivat (approved):

PR #118, opened 2020-09-27
Issue 105 : Dark Mode + Theming

Jongwoo-Shim:
* Themes scss file created with default color scheme and dark color scheme.

* Logic for managing dark mode toggle added.

* Modified existing scss/html to work with theme toggling.

* https://github.com/alshedivat/al-folio/issues/105

PR #119, opened 2020-09-29
 Issue 105 : Dark Mode + Theming #118 

Jongwoo-Shim:
Created separate branch for dark mode, as I maybe submitting another PR.


alshedivat (changes requested):
Thanks for your contribution! Looks great, especially the theming feature.

One request before I merge: could you make the dark mode an optional feature (off by default)?
To do that, simply add a variable to `_config.yml` called `enable_dark_mode` (similar to [other optional features](https://github.com/alshedivat/al-folio/blob/master/_config.yml#L146-L154)) and make sure that dark-mode specific code is loaded only when `enable_dark_mode=true`.

Jongwoo-Shim:
Sounds good I'll make the change by Wednesday latest.

Jongwoo-Shim:
Made the requested change. If you require a rebased version I can make another pull request using the master branch : https://github.com/Jongwoo-Shim/al-folio/tree/master

alshedivat (approved):
Great, thanks for making the changes!

PR #132, opened 2020-10-22
Add keybase badge

shpark:
This commit allows to add keybase social url.


alshedivat:
Thanks for your contribution!

PR #136, opened 2020-11-06
Adding back the description on the About page.

swabhs:
Added back the missing description on `about.html`, similar to `page.html`. This could additionally be added back to the `post.html` template as well.


alshedivat:
Nice catch. Apparently, it was mistakingly removed in #102. Thanks!

PR #137, opened 2020-11-06
Update _base.scss for the `post-description` style.

swabhs:
Description text size should match whether or not there is a link.


alshedivat:
Thanks for the fix! :)

Issue #129, opened 2020-10-12
Google Analytics Code block is going to body


saeed1262:
**Describe the bug**
After deploying the Google analytics code blocks goes to the body. However, it is better to go the head so that it can be verified by google search console.

**Expected behaviour**
Copying the Google analytics code block in the head of index.html when deploying

louis1642:
This is marked as completed but the Google Analytics code blocks are still in the body

PR #140, opened 2020-11-13
Move GA script to head.html

zt-chen:
Related to #129 
Google analytics code is moved to head.html


alshedivat:
Looks great! Thanks for your contribution.

PR #153, opened 2020-12-15
Update README, change $theme-color in _themes.scss

huyen-nguyen:
`$theme-color` value should be updated from `_sass/_themes.scss` instead of `_sass/variables.scss`


alshedivat:
Thanks for the fix!

Issue #100, opened 2020-07-06
Better author identification in /publications


pclucas14:
Hi, 

My brother and I do a lot of research together. The current theme only asks for my last name in order to underline my name in my publication list. However, since my brother has the same last name, his name gets underlined as well.

I'm super n00b when it comes to website stuff, I played around but couldn't come up with a solution. Is there any way to fix this ?

btw thanks for the free theme, it's very friendly :)

-Lucas

![Screen Shot 2020-07-06 at 7 56 54 PM](https://user-images.githubusercontent.com/20862882/86670714-f0a30e80-bfc2-11ea-9c73-c3637604db3f.png)

pclucas14:
I ended up just manually fixing the html file

alshedivat:
Thanks for raising this issue! Proper author disambiguation is important. Let me re-open this issue, so I can come back and address it later.

sangheek16:
@pclucas14 Hi, I also had the same issue with my co-authors. Can I ask how you fixed the html file? Which file did you fix?

pclucas14:
The `index.html` file in the `publications/` folder. I added / removed the `<em>` html tags where needed

gkthiruvathukal:
Just wanted to clarify something. I was looking at issue #156. I believe this issue is focused on the "self" author but does not address the co-authors disambiguation for this issue. Is this correct?

We need something along these lines to allow for disambiguation.

Suppose you had names like Mike Adams, John Adams, and John Quincey Adams (sometimes written as Q), this would preserve the current (nicely) straightforward design:

```
Adams:
   John:
      url: <url for John Adams>

      Quincey:
          url: <url for John Quincey Adams>
      Q:
          url: <url for John Quincey Adams>
    Mike:
       url: <url for Mike Adams>
```

There are obviously some other edge cases to consider (people with multiple middle names) or the unlikely case that someone's middle name is `url`, although we could require that every name is capitalized so we know names from other properties.

Another possibility could be to make a more "flat" design using a YAML list:

```
- surname: Adams
  given: John
  middle1: ["Q", "Quincey"]
  url: ...

- surname: Adams
  given: John
  url: ...
```

Perhaps we could discuss this a bit on Gitter? I'd be interested in pairing to see if we could get it done. I don't do much Ruby but it doesn't seem like an impossible task, once we have the right data model.

alshedivat:
@gkthiruvathukal, nice catch, thanks for pointing this out!

[...]

PR #156, opened 2020-12-17
Support more accurate author annotation

ziruizhuang:
added a list `scholar: first_name` in `_config.yml`
updated `_layouts/bib.html` template
supports more accurate author annotation by both matching the last name and one form of first name in the `_config.yml/scholar: first_name`

possible related issue #100


alshedivat:
LGTM. Thanks for this contribution!

Issue #143, opened 2020-11-23
Support for wide screen


deng-cy:
**Is your feature request related to a problem? Please describe.**
Currently the website is narrow even on a wide screen. 

**Describe the solution you'd like**
I hope there is a feature to extend the width of the pages.

alshedivat:
Thanks for the suggestion, @deng-cy.

Could you provide more details on how you would want the webpage to behave on wide screens? Right now, the theme is designed to have max-width (similar to [distill.pub](https://distill.pub/)), since it would look weird if it was to span the full screen on wide displays.

If you have links to other websites with similar layouts as al-folio which extend to the width of the page on wide displays, please let me know and I'll think if we can add that as an optional layout to al-folio.

deng-cy:
Some page types look good even if it spans the full screen. One option is to set a maximum width in `_config.yml`.

ziruizhuang:
One workaround is to manually adjust the value `$max-content-width: 800px;` in `_sass\_variables.scss`. 
But yes it would be nice if there is an option in `_config.yml` as well.

rohandebsarkar:
`max_width` configuration in the `_config.yml` file does not apply to distill themed blog posts.

PR #157, opened 2020-12-18
add wide screen configuration

ziruizhuang:
add wide screen configuration by adding a `max_width` config in `_config.yml` 

possibly related issue #143


alshedivat:
Looks good. I would prefer to keep `assets/css/main.scss` contain only imports, but for now it's an okay workaround.

PR #163, opened 2020-12-31
Github action: add deploy workflow

junghans:
This will automatically deploy the page when pushing to source branch (i.e. `master` and `source`).
That basically allows editing through the GitHub web interface, no more need to install bundler locally.

This also supports pull requests, which can be useful to collaboratively make a conference page. For PRs the whole build is still done except the actual `git push` at the end. To achieve this I added a `--dry-run` option to `bin/deploy`.

Side notes:
- added `--verbose` option to make debugging of `bin/deploy` easier
- changed the exit code for unknown options to 1, before it would end gracefully (`exit 0`)
- fixed `--user` option calling `shift` twice
- using actions/cache to cache bundle
- the current branch is always used as the source branch, however the deploy action is only triggered for `master` / `source` (and PR to these, but PR are not deployed).
- in case of a `.github.io` repo, the deploy branch is switched from `gh-pages` to `master`.
- all the GitHub action stuff is pretty easy to customize
- all changes to `bin/deploy` are backward compatible

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,52 @@
+name: Deploy
+
+on:
+  push:
+    branches:
+      - master
+      - source
+  pull_request:
+    branches:
+      - master
+      - source
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+    - name: Setup Ruby
+      uses: actions/setup-ruby@v1
+    - name: Enable bundler cache
+      uses: actions/cache@v2
+      with:
+        path: vendor/bundle
+        key: ${{ runner.os }}-gems-${{ hashFiles('**/Gemfile.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-gems-
+    - name: Install deps
+      run: |
+        gem install bundler
+        bundle config path vendor/bundle
+        bundle install --jobs 4 --retry 3
+    - name: Setup deploy options
+      id: setup
+      run: |
+        git config --global user.name "GitHub Action"
+        git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
+        if [[ ${GITHUB_REF} = refs/pull/*/merge ]]; then # pull request
+          echo "::set-output name=SRC_BRANCH::${GITHUB_BASE_REF}"
+          echo "::set-output name=DRY_RUN::--dry-run"
+        elif [[ ${GITHUB_REF} = refs/heads/* ]]; then # branch, e.g. master, source etc
+          echo "::set-output name=SRC_BRANCH::${GITHUB_REF#refs/heads/}"
+        fi
+        if [[ ${{ github.repository }} = *.github.io ]]; then # user/org repo
+          echo "::set-output name=DEPLOY_BRANCH::master"
+        else
+          echo "::set-output name=DEPLOY_BRANCH::gh-pages"
+        fi
+    - name: Deploy website 
+      run:  yes | bin/deploy --verbose ${{ steps.setup.outputs.DRY_RUN }}
+                    --src ${{ steps.setup.outputs.SRC_BRANCH }} 
+                    --deploy ${{ steps.setup.outputs.DEPLOY_BRANCH }} 
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -36,7 +36,7 @@ jobs:
         git config --global user.name "GitHub Action"
         git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
         if [[ ${GITHUB_REF} = refs/pull/*/merge ]]; then # pull request
-          echo "::set-output name=SRC_BRANCH::${GITHUB_BASE_REF}"
+          echo "::set-output name=SRC_BRANCH::${GITHUB_HEAD_REF}"
           echo "::set-output name=DRY_RUN::--dry-run"
         elif [[ ${GITHUB_REF} = refs/heads/* ]]; then # branch, e.g. master, source etc
           echo "::set-output name=SRC_BRANCH::${GITHUB_REF#refs/heads/}"
--- a/bin/deploy
+++ b/bin/deploy
@@ -7,7 +7,7 @@
 SRC_BRANCH="master"
 DEPLOY_BRANCH="gh-pages"
 
-USAGE_MSG="usage: deploy [-h|--help] [-u|--user] [-s|--src SRC_BRANCH] [-d|--deploy DEPLOY_BRANCH]"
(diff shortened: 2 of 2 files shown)


alshedivat:
@junghans, this is awesome, thanks for the contribution!

Overall, LGTM. Before merging this PR, let me play with the config to make sure it takes care of all / most of use cases.

bin/deploy
```diff
@@ -20,7 +20,6 @@ while [[ $# > 0 ]]; do
         -u|--user)
         SRC_BRANCH="source"
         DEPLOY_BRANCH="master"
-        shift
```
alshedivat, line 23: Why was this `shift` deleted? Wouldn't this introduce a bug?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,52 @@
...
+    - name: Install deps
+      run: |
+        gem install bundler
+        bundle config path vendor/bundle
+        bundle install --jobs 4 --retry 3
+    - name: Setup deploy options
+      id: setup
+      run: |
+        git config --global user.name "GitHub Action"
+        git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
alshedivat, line 37: Where is this email coming from? Does it correspond to a specific GitHub action?

bin/deploy
```diff
@@ -30,10 +29,16 @@ while [[ $# > 0 ]]; do
         DEPLOY_BRANCH="$2"
         shift
         ;;
+        --verbose)
+        set -x
+        ;;
```
alshedivat, line 34: Same here (and on line 37 below as well): it looks like some options above have `shift` while others don't. Why and what's the expected behavior?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,52 @@
...
+        if [[ ${GITHUB_REF} = refs/pull/*/merge ]]; then # pull request
+          echo "::set-output name=SRC_BRANCH::${GITHUB_BASE_REF}"
+          echo "::set-output name=DRY_RUN::--dry-run"
+        elif [[ ${GITHUB_REF} = refs/heads/* ]]; then # branch, e.g. master, source etc
+          echo "::set-output name=SRC_BRANCH::${GITHUB_REF#refs/heads/}"
+        fi
```
alshedivat, line 43: Could you explain the logic of this step:

1) Do I understand correctly that on PR, it will set `SRC_BRANCH` to `${GITHUB_BASE_REF}` (which must be the PR target) and also set the `--dry-run` option on. Why is dry-run necessary? What is the intended behavior on PR?
2) On push, it sets `SRC_BRANCH` to the branch name the push was made to, correct?

alshedivat (commented):
LGTM.

Could you please explain and document the logic of this workflow a bit more by answering a couple of questions I left inline? I'd be happy to merge after that. Thanks.

bin/deploy
```diff
@@ -30,10 +29,16 @@ while [[ $# > 0 ]]; do
...
         *)
-        echo "Option $1 is unknown."
-        echo $USAGE_MSG
-        exit 0
+        echo "Option $1 is unknown." >&2
+        echo $USAGE_MSG >&2
+        exit 1
         ;;
     esac
     shift
```
junghans, line 44: @alshedivat There is already a shift here.

bin/deploy
junghans: Line 44 has one `shift` for each option already, hence there is only a `shift` instead of a `shift 2` for `--src` as well.

.github/workflows/deploy.yml
junghans: You can use any email here, `<some_number>+github-actions[bot]@users.noreply.github.com` is the Github Actions default email. See https://github.community/t/github-actions-bot-email-address/17204 for discussion.

.github/workflows/deploy.yml
junghans: Re:  1. I was actually mistaken here, `GITHUB_BASE_REF` is the target branch, so we need to use `GITHUB_HEAD_REF` here (see https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions). The `--dry-run` is needed, so that a PR from an internal branch doesn't get deployed. PR from a fork won't have write access to the repo, so there it won't be an issue.

Re: 2. Yes, the action is only triggered when pushed to `master` or `source` and then `GITHUB_REF` will be something like `refs/heads/master`. It is worth mentioning that, in the user/org case, a push to the `source` branch will trigger the deploy action onto master, but the then automatic commit to the `master` branch won't trigger another Github action execution. GitHub doesn't allow a push triggered by an action to trigger another action.

bin/deploy
junghans: Try `bin/deploy --user --src master` to reproduce the issue in the original deploy script.

.github/workflows/deploy.yml
alshedivat: Thanks for explaining!

Re: 1. In the dry-run regime, I guess base vs. head branch doesn't matter. Just to double-check the logic, on internal PR merge, the action will get triggered twice: (1)  due to PR (which will be a dry-run) and (2) due to the resulting commit from the PR in the base branch (which will trigger the actual deploy), right?

alshedivat (approved):

.github/workflows/deploy.yml
junghans: Re 1. I still think it needs to be `GITHUB_HEAD_REF`, let's assume you have PR from `new_feature` to `master`. The action is supposed to test the commits on `new_feature` (`GITHUB_HEAD_REF`) and not `master` (`GITHUB_HEAD_REF`). And this PR won't trigger twice as a push to `new_feature` doesn't trigger an action, only an open PR will trigger an action (see the selection of branches in the top of the file `on: push: branches:`)

junghans: Plus I will submit a bunch of unrelated other changes as PR afterwards to test this ;-)

alshedivat: > Re 1. I still think it needs to be `GITHUB_HEAD_REF`, let's assume you have PR from `new_feature` to `master`. The action is supposed to test the commits on `new_feature` (`GITHUB_HEAD_REF`) and not `master` (`GITHUB_HEAD_REF`). And this PR won't trigger twice as a push to `new_feature` doesn't trigger an action, only an open PR will trigger an action (see the selection of branches in the top of the file `on: push: branches:`)

Got it, so the following adjustment should make it work as expected?
```suggestion
        if [[ ${GITHUB_REF} = refs/pull/*/merge ]]; then # pull request
          echo "::set-output name=SRC_BRANCH::${GITHUB_HEAD_REF}"
          echo "::set-output name=DRY_RUN::--dry-run"
        elif [[ ${GITHUB_REF} = refs/heads/* ]]; then # branch, e.g. master, source etc
          echo "::set-output name=SRC_BRANCH::${GITHUB_REF#refs/heads/}"
        fi
```

> Plus I will submit a bunch of unrelated other changes as PR afterwards to test this ;-)

Awesome, thanks! :)

Issue #167, opened 2021-01-03
Deploy script deletes untracked files in working directory without asking


LoryPack:
**Acknowledge the following**
- [x] I carefully read and followed the [Getting Started](https://github.com/alshedivat/al-folio#getting-started) guide.
- [x] I read through [FAQ](https://github.com/alshedivat/al-folio#faq) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [x] The issue I am raising is a potential bug in al-folio and not just a usage question. <br> [For usage questions, please use [gitter chat](https://gitter.im/alshedivat/al-folio) instead of raising an issue.]

**Describe the bug**
When using the script for deploying, it checks for uncommited changes to tracked files but it does not check if the working directory contains untracked files. Those files are then deleted when running the deploy script, which may be annoying (it may for instance remove a blog post on which you are working on but which you are not ready to publish).

**To Reproduce**
1. Create a new file in the directory and do not add to git
2. Run the deploy script

**Expected behavior**
Similar to the case of uncommited changes to tracked files, it would be nice to raise an error prompting the user to either commit or stash untracked (and not ignored) files.

**System (please complete the following information):**
-    OS: Ubuntu 20.04
-   Browser: Firefox 84.0, Chrome 81.0.4044.122
-    Jekyll version: 3.9.0
-    Ruby version: ruby 2.7.0p0 (2019-12-25 revision 647ee6f091) [x86_64-linux-gnu]

Issue #131, opened 2020-10-18
deploy should switch back to original branch if it fails


eamanj:
Currently if ./bin/deploy fails, it won't switch back to the branch it was on when the command was called. It will stay on gh-pages branch.

alshedivat:
For future reference: to address this issue, need to switch from `set -e` in the `./bin/deploy` to a cleanup using [trap](https://www.linuxjournal.com/content/bash-trap-command).

PR #168, opened 2021-01-03
Check for local untracked files when deploying

LoryPack:
Way to fix issue #167.

### DIFF
--- a/bin/deploy
+++ b/bin/deploy
@@ -62,6 +62,16 @@ if ! git diff-index --quiet HEAD --; then
     [[ "$0" = "$BASH_SOURCE" ]] && exit 1 || return 1
 fi
 
+# Check if there are any untracked files
+u="$(git ls-files --exclude-standard --others)"
+if ! test -z "$u"; then
+    echo "There are untracked files:"
+    git ls-files --exclude-standard --others
+    echo "Please commit those files or stash them before proceeding."
+    echo "Aborting."
+    [[ "$0" = "$BASH_SOURCE" ]] && exit 1 || return 1
+fi
+
 # Switch to source branch (creates it if necessary from the current branch)
 if [ `git branch | grep $SRC_BRANCH | tr ' ' '\n' | tail -1` ]
 then


alshedivat:
Awesome, thank you!

bin/deploy
```diff
@@ -62,6 +62,16 @@ if ! git diff-index --quiet HEAD --; then
     [[ "$0" = "$BASH_SOURCE" ]] && exit 1 || return 1
 fi
 
+# Check if there are any untracked files
+u="$(git ls-files --exclude-standard --others)"
+if ! test -z "$u"; then
```
alshedivat, line 67:
```suggestion
if ! test -z "$(git ls-files --exclude-standard --others)"; then
```

alshedivat (approved):
Looks great, thanks for the fix! Please make a tiny stylistic change and I'll be happy to merge.

bin/deploy
LoryPack: Hi, thanks so much for that. There is actually a reason why I kept that separate (see here https://stackoverflow.com/a/2659808/13516418). Basically if you do not split the line there is a small chance that errors from git ls-files are translated as “no untracked files” reports, so that the deploy script may behave in weird ways which may be tricky for the user to understand. 

But if you think that is not an issue in practice I am happy to modify as you suggested!

alshedivat: I see, thanks for clarifying.

It looks like separating the command into 2 lines makes the script treat the `git ls-files` error as if there were no untracked files. I would prefer the script to abort in such edge cases instead of proceeding silently.

LoryPack: That makes sense. The above stackoverflow question seems to provide a way to address that: 

```
git ls-files --others --error-unmatch --exclude-standard . >/dev/null 2>&1; ec=$?
if test "$ec" = 0; then
    echo some untracked files
elif test "$ec" = 1; then
    echo no untracked files
else
    echo error from ls-files
fi
```

What do you think of that solution?

alshedivat: In case when `git ls-files --others --error-unmatch --exclude-standard` returns an error code (which is an unlikely edge case), I believe the script will abort because of `set -e` on line 43 and won't get to the if-then-else part (to test this, we would need to come with a situation that results in `git ls-files` erroring out).

Btw, using `set -e` is not the best way to handle exceptions in bash (e.g., see #131). I plan to move away from `set -e` to proper exception handling. For now, I would prefer your original solution compressed to 1 line as I suggested above.

LoryPack: Ok, thanks for explaining. I'll commit your previous suggestion then.

PR #176, opened 2021-01-06
Add webpage + add panelbear analytics

AbstractGeek:
Minor commits:
1. Adding my webpage to the list of academic pages
2. Adding panelbear analytics option. Changes to the config file (analytics id entry, enable entry). Script added to head.html.


alshedivat:
Thanks for your contribution!

PR #182, opened 2021-01-12
Issue 180 - Distill Bugfix

Jongwoo-Shim:
* dark_mode.js was not being loaded on distill blog pages.
* Moved dark_mode.js and jquery.html to head.html to simplify/standardize dark mode/theming for other pages.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -33,9 +33,14 @@
 
 <link rel="canonical" href="{{ page.url | replace:'index.html','' | relative_url }}">
 
+<!-- JQuery -->
+{% include scripts/jquery.html %}
+
 <!-- Theming-->
 {% if site.enable_darkmode %}
   <script src="{{ '/assets/js/theme.js' | relative_url }}"></script>
+  <!-- Load DarkMode JS -->
+<script src="{{ '/assets/js/dark_mode.js' | relative_url }}"></script>
 {% endif %}
 
 {% if site.enable_google_analytics %}
--- a/_includes/scripts/misc.html
+++ b/_includes/scripts/misc.html
@@ -8,6 +8,3 @@ $(function () {$('[data-toggle="tooltip"]').tooltip()})
 
 <!-- Load Common JS -->
 <script src="{{ '/assets/js/common.js' | relative_url }}"></script>
-
-<!-- Load DarkMode JS -->
-<script src="{{ '/assets/js/dark_mode.js' | relative_url }}"></script>
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -24,7 +24,6 @@
 
   </body>
 
-  {% include scripts/jquery.html %}
   {% include scripts/bootstrap.html %}
   {% include scripts/mansory.html %}
   {% include scripts/misc.html %}


Jongwoo-Shim:
Bug fix can be checked here : https://jongwoo-shim.github.io/al-folio/blog/2018/distill/

alshedivat:
Thanks for the bug fix, really appreciate it!

While we are at it, do you mind adding some minor dark mode style fixes (1) the text color of `authors` in the blog post header (in the dark mode, should be white), (2) the color of horizontal lines (should be white in dark mode), and (3) same for the color of `footnotes` and `references`?

Thanks again for all your contributions to the theme!!

Jongwoo-Shim:
@alshedivat No problem. I actually have made most of the fixes and was planning on making a desperate branch for it.
The distill template was really problematic, due to the fact that the css was written into the template.v2.js file.

Just need your input on the code blocks : https://github.com/alshedivat/al-folio/issues/183, and I'll make a pull request with the fixes.

alshedivat:
@Jongwoo-Shim, thank you! Let me take a look at the issue you linked.

Re: distill styles and `template.v2.js` -- I maintain a fork of the [distillpub-template](https://github.com/alshedivat/distillpub-template/tree/al-folio) which has `al-folio` branch. It might be easier to make significant style changes there and then re-generate `template.v2.js`, which is programmatically generated. However, dark mode style overrides do not require significant changes (only adjusting colors in a few places), so the preferred solution is just to override CSS without touching `template.v2.js`.

Jongwoo-Shim:
@alshedivat Part of the reason that I started to modify the template.v2.js file was due to the fact that there are several css attributes that are currently hard coded into it. I was finding that it was making it difficult/impossible to modify certain attributes of the distill template. 

Then again, I'm not particularly great at web-dev so I might be implementing the css incorrectly.

Regardless, I personally think that the css should be stripped from the distill template, and either added to the _base.scss file or added to it's own scss file. 

This is due to the fact that : 
* It's unintuitive where to look to make css changes for distill pages. 
* It forces additional changes to be custom tailored for distill, which I think would be otherwise unnecessary.

Unless, there's a specific reason that would restrict the movement of the css from the template. The template that you linked could just be used to create the scaffold of the page, while the css is managed in the al-folio project.

We could merge in the changes to the template.v2.js file as a temporary bug-fix and migrate the css files later down the line.

Jongwoo-Shim:
@alshedivat Additionally, are you looking for all of the fixes to be in a single merge? As previously mentioned I was planning to make a separate branch/merge request for it. But I can adjust it if need be.

alshedivat:
> We could merge in the changes to the template.v2.js file as a temporary bug-fix and migrate the css files later down the line.

That sounds good. Let's discuss this in the corresponding issue before making any changes.

> @alshedivat Additionally, are you looking for all of the fixes to be in a single merge?

No need to, it's a good idea to stage these changes through multiple PRs. I'll also think more what's the best way to go about altering CSS in `template.v2.js`. Since that file is generated from `distillpub-template`, if we are to upgrade to a later version of distillpub styles in the future, it will have to be re-generated, so any changes we make will be lost or have to be re-applied.

PR #204, opened 2021-02-21
Fix titles and headings

rogerthat94:
The If statements to decide the page title and header seemed incorrect. This fixes these.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -2,7 +2,7 @@
 <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
 <meta http-equiv="X-UA-Compatible" content="IE=edge">
 
-<title>{% if site.title == blank %}{{ site.title }}{% else %}{{ site.first_name }} {{ site.middle_name }} {{ site.last_name }}{% endif %}{% if page.title and page.url != "/" %} | {{ page.title }}{% endif %}</title>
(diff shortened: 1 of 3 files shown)


alshedivat:
Thanks for the contribution. Which exact issue does this PR fix?

rogerthat94:
`{{ site.title }}` was never getting used as the title and heading. The condition `if site.title == blank` was never true, because blank was not in quotes. In addition, {{ site.title }} is the wrong value to use in this case. 

To see the problem, try using the theme with title set as blank.

alshedivat:
@rogerthat94, thanks for pointing this out. Looks like the bug was introduced in [this commit](https://github.com/alshedivat/al-folio/commit/619bb2df74a79d551785d3c478fb7cc6fe2c028d). If that's the case, then `_layouts/about.html` probably has the same issue. Could you please confirm that?

On a stylistic note, how about we switch from `{% if site.title == "blank" %}` to just `{% if site.title %}` and flip the true and false branches? I just find it weird that liquid evaluates `"" == "blank"` to true.

And thanks again for spotting this!

rogerthat94:
We don't want to look for site.title being undefined. We want to check if it's specifically the string "blank"

See https://github.com/alshedivat/al-folio/blob/9229f7d881d215969378fdc287fdad51936037f9/_config.yml#L5

If you'd rather check `{% if site.title %}`, we should change _config.yml to comment out the title and say "the website title (if undefined, full name will be used instead)"

PR #223, opened 2021-03-20
GDPR cookie consent

### LINKED ISSUE #199: GDPR compilent cookie consent dialog
Hi,

since I am from the Europe and want my site to be accessed by EU citizens I need to ask users if they are ok with the use of cookies. Adding a dialogue by default that that asks the user before any cookies / localStorage are used would be nice. 

I have integrated the dialog from [https://www.cookieconsent.com/](https://www.cookieconsent.com/) on [my website](http://hschwane.github.io) and am happy to file a PR if you are interested. However It is not shown directly at the top, but below the menu bar for some reason. Not sure how to fix that.

hschwane:
Adds a cookie consent notice as discussed in #199.

Currently the notice is scrolled up and down, even if the header is set to fixed, which looks a bit strange. 
Also we would somehow need to block the embedded twitter tweets in the twitter example, as they generate cookies as well.
I added a new example on how to block a youtube video based on the cookie settings, but I do not use the twitter plugin myself.

resolves #199

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,6 +307,24 @@ Easily create beautiful grids within your blog posts and project pages:
   </a>
 </p>
 
+### Cookie Consent
+Includes a cookie consent dialog to allow your user to opt out of non-necessary cookies. 
+To block scripts based on the user's cookie settings, include the javascrip as:
+```
+<script type="text/plain" cookie-consent="<level>">
+   // only executed if user agreed to cookies
+</script>
+```
+Where `<level>` is one of:
+- strictly-necessary
+- functionality
+- tracking
+- targeting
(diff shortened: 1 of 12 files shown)


alshedivat:
closing this PR. ideally, we should implement a more light-weight GDPR cookie consent as suggested in https://github.com/alshedivat/al-folio/issues/199#issuecomment-2081572832.

alshedivat:
closing this PR. ideally, we should implement a more light-weight GDPR cookie consent as suggested in https://github.com/alshedivat/al-folio/issues/199#issuecomment-2081572832.

PR #234, opened 2021-04-02
Change gitcdn link in 

jkmacc-LANL:
`gitcdn.link` was quite slow, so I tried the alternate link at `gitcdn.xyz` and there are no issues with lag anymore.
`gitcdn` was used to quickly get pygments code highlighting css themes inside `_includes/head.html` from [this](https://github.com/jwarby/jekyll-pygments-themes) repo.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -25,7 +25,7 @@
 <link rel="stylesheet" type="text/css" href="https://fonts.googleapis.com/css?family=Roboto:300,400,500,700|Roboto+Slab:100,300,400,500,700|Material+Icons">
 
 <!-- Code Syntax Highlighting -->
-<link rel="stylesheet" href="https://gitcdn.link/repo/jwarby/jekyll-pygments-themes/master/{{ site.highlight_theme }}.css" />
+<link rel="stylesheet" href="https://gitcdn.xyz/repo/jwarby/jekyll-pygments-themes/master/{{ site.highlight_theme }}.css" />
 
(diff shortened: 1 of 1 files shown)


alshedivat:
Interesting, I didn't know about the issues with `gitcdn.link` (it works fine on my end). Thanks for the fix though!

jkmacc-LANL:
I wonder if it was on my end.  My apologies.  A quick search suggested the `.link` address had also given others some trouble.  Thanks!

ashafaei:
I have been experiencing a consistently slow load on my homepage at [shafaei.ai](https://www.shafaei.ai) after I updated my copy of the repository to the latest changes. After digging in a bit I found that gitcdn.xyz is the culprit, which brought me to this page. I ended up undoing the changes introduced in this pull-request and the problem was resolved.

willgearty:
I've been having the same issues for the last few days. Switching back to gitcdn.link also worked for me (my website took more than a minute to load before, now loads in a second).

alshedivat:
I'm going to revert this change. At least in North America, gitcdn.xyz takes forever to load (not sure about other locales).

jkmacc-LANL:
Sorry for the trouble.  Since this seems to be region-dependent, is it worth leaving this open so others can find it?

ashafaei:
@jkmacc-LANL isn't there a standard way of pointing to that file? An alternative might be to put that small CSS file in this repository directly if the original authors permit.

PR #270, opened 2021-05-11
add dblp to social links

pazz:
Pretty self-explanatory I guess.
Many thanks for the awesome theme!


junghans:
Field should be added to `_config.yml` as well.

PR #284, opened 2021-05-21
Initial Atom (RSS-like) feed support

### LINKED ISSUE #263: RSS Feed for Blog part
It is so useful to have a RSS XML generated file, suitable for theoldreader or your favourite RSS reader.

In this way, someone could easily read your contents.

@alshedivat the only "problem" I am thinking about is the standardization of certain elements: in my blog for example, images and blockquotes are treated with bootstrap (raw html identities), since they are not in standard markdown, due to the inability of centering them with markdown, this could be a problem while parsing, because each of us could easily treat them in a different way, so we should deal with a lot of tags.

But it seems a great PoC to me, even if it would "just work with text", it should be a giant leap.

dag7dev:
Close #263

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -24,10 +24,7 @@ impressum_path:  # set to path to include impressum link in the footer, use the 
 # RSS Feed
 # -----------------------------------------------------------------------------
 # will use title and url fields
-description:  # Higly suggested: FILL THIS FIELD! If blank, this could lead to errors. A longer description of what your site is about, e.g., "Where I blog about Jekyll and other awesome things"
-author:  # your name and surname
-feed:
-  excerpt_only: false # set it to have all posts in feed will be without <content> tags.
+# Take a look to https://github.com/jekyll/jekyll-feed for more customization
 
 # -----------------------------------------------------------------------------
 # Layout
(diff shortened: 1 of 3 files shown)


_config.yml
```diff
@@ -20,6 +20,15 @@ baseurl: /al-folio # the subpath of your site, e.g. /blog/
 last_updated: false # set to true if you want to display last updated in the footer
 impressum_path:  # set to path to include impressum link in the footer, use the same path as permalink in a page, helps to conform with EU GDPR
 
+# -----------------------------------------------------------------------------
+# RSS Feed
+# -----------------------------------------------------------------------------
+# will use title and url fields
+description:  # Higly suggested: FILL THIS FIELD! If blank, this could lead to errors. A longer description of what your site is about, e.g., "Where I blog about Jekyll and other awesome things"
```
alshedivat, line 27: `description` field is already given on line 10.

_config.yml
```diff
@@ -20,6 +20,15 @@ baseurl: /al-folio # the subpath of your site, e.g. /blog/
...
+author:  # your name and surname
+feed:
+  excerpt_only: false # set it to have all posts in feed will be without <content> tags.
```
alshedivat, line 30: These fields seem to be optional. I'd prefer to remove these fields and refer users to `jekyll-feed` docs instead, since there many more options. Basically, simply substitute all fields in this subsection with a brief comment with a link to https://github.com/jekyll/jekyll-feed.

alshedivat (changes requested):
Thanks for the contribution! Looks great. I left just a couple of comments.

dag7dev:
@alshedivat I've been busy in these days, now it should be fine!

alshedivat (approved):

Issue #298, opened 2021-06-04
Distill post darkmode


ivanpuhachov:
**Is your feature request related to a problem? Please describe.**
Citations, footnotes, and code blocks do not display correctly in the dark mode since distill does not support the dark mode by default. I have found a solution and opening this issue to submit a pull request.

I fixed `_distill.scss` and `js/distillpub/template.v2.js`, putting `var(--global-text-color)` here and there. Check out working dark mode on [my page](https://puhachov.xyz/blog/2018/distill/).

Issue #280, opened 2021-05-17
Blockquote Appears to Be Broken


alexandermichels:
**Acknowledge the following**
- [x] I carefully read and followed the [Getting Started](https://github.com/alshedivat/al-folio#getting-started) guide.
- [x] I read through [FAQ](https://github.com/alshedivat/al-folio#faq) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [x] The issue I am raising is a potential bug in al-folio and not just a usage question. <br> [For usage questions, please use [gitter chat](https://gitter.im/alshedivat/al-folio) instead of raising an issue.]

**Describe the bug**
It appears that blockquotes are indistinguishable from normal text. You can see it on this blog post on the demo site (the second-to-last paragraph is a blockquote): https://alshedivat.github.io/al-folio/blog/2015/formatting-and-links/ and the problem exists on [my site](https://alexandermichels.github.io/blog/2015/formatting-and-links/) as well. The problem seems to exist for both light and dark modes and when using the markdown `>` syntax and html blockquote tag. Using inspect confirms that it is a blockquote, just doesn't appear different at all.

**To Reproduce**
See the demo site: https://alshedivat.github.io/al-folio/blog/2015/formatting-and-links/ or you can test it locally by using the `>` markdown syntax such as:

```
> this is a blockquote test
```
> this is a blockquote test

**Expected behavior**
I would expect (and believe the old behavior was) that the text looks different from regular text. I believe it used to be indented with a line to the left of the text.

**System (please complete the following information):**
Obviously this only relevant to my site, not sure what the demo site is using:
 - OS: Linux Mint
 - Browser (and its version): checked on Chrome (Version 90.0.4430.212 (Official Build) (64-bit)) and Firefox 88.0.1
- Jekyll version [e.g. 3.8.7]: 4.2.0
- Ruby version [e.g. 2.6.5]: ruby 2.7.0p0 (2019-12-25 revision 647ee6f091) [x86_64-linux-gnu

dag7dev:
I can confirm this, I solved this by manually styling a div but I can confirm this issue.

ivanpuhachov:
Confirming this. As a temporary fix I added this to `_base.scss`:

```
blockquote {
  background: #f9f9f9;
  border-left: 2px solid #000;
  margin: 1.5em 10px;
  padding: 0.5em 10px;
  font-size: 1.2rem;
}

```

[...]

PR #299, opened 2021-06-04
Fix distill darkmode

ivanpuhachov:
Fixing the issue with distill darkmode #298

I found out that the problem emerges with numerated lists, table entries, inline math equations, blockquotes (see also #280 ), footnotes, citations, hovering boxes and code highlights. Seems to be fixed now, take a look on `al-folio/blog/2021/distill/` when building locally, or online on my page: [here](https://puhachov.xyz/blog/2018/distill/)

Let me know if something is missing. Cheers!


alshedivat:
Thanks so much for fixes!

ivanpuhachov:
Fixes #298 

(comment to link pr to issue)

alshedivat:
@ivanpuhachov, just realized you made changes to `assets/js/distillpub/template.v2.js`. Thanks for taking the effort!

I'm wondering, would it be hard to factor out these style changes into `_sass/_distill.scss` instead of directly editing the `template.v2.js`? The reason is that `template.v2.js` is currently auto-generated from the [al-folio branch of my fork of distillpub template](https://github.com/alshedivat/distillpub-template/tree/al-folio) (later, I'm thinking of switching to loading `template.v2.js` directly from https://distill.pub/template.v2.js). So, I'd strongly prefer treating `template.v2.js` as read-only and make changes by overriding functionality.

ivanpuhachov:
The problem is that `template.v2.js` automatically generates `style` properties for created items and stores them right in the cell (see, i.e. line 1374 in `template.v2.js`). I tried to override them with scss file and `!important` flag but I did not succeed. Is there another method to do so? I can try

alshedivat:
You are right, it looks like most of distill is written using web components that encapsulate styles using shadow DOM. The styles inside shadow DOM cannot be overridden from outside using CSS, unfortunately (see [discussion here](https://stackoverflow.com/questions/37352637/shadow-dom-styling-from-the-outside)). But, these styles can be overridden using javascript without directly modifying `template.v2.js`.

I added `assets/js/distillpub/overrides.js` that overrides `d-footnote` and `d-cite` styles. Would you mind reverting your changes in `template.v2.js` and using this method instead?

ivanpuhachov:
Great. Will do!

ivanpuhachov:
Done!

alshedivat (approved):

PR #341, opened 2021-07-08
Fix for project category text color in light mode

alexandermichels:
This is a minor fix for the Projects page's project category headers. Currently, the headers are hard-coded to light grey which makes them difficult to read in light-mode. This can be seen on the [demo website](https://alshedivat.github.io/al-folio/projects/) and in the screenshot below

al-folio-light-theme

This fix simply changes the color to the `--global-text-color` allowing the color of the headers to change based on whether the site is in light or dark mode. You can see screenshots for both light and dark mode below:

al-folio-pr-dark-theme
al-folio-pr-light-theme

### DIFF
--- a/_sass/_base.scss
+++ b/_sass/_base.scss
@@ -337,8 +337,8 @@ footer.sticky-bottom {
   }
 
   h2.category {
-    color: $grey-color-light;
-    border-bottom: 1px solid $grey-color-light;
+    color: var(--global-text-color);
+    border-bottom: 1px solid var(--global-text-color);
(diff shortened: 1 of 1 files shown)


alshedivat:
Thanks for bringing this up. The headers are intentionally light grey to put less emphasis on that detail (similar for the color of the years on the publications page). I agree that readability is not ideal, but changing them to `--global-text-color` makes it detail too conspicuous. I'd be okay with making these headers slightly darker, but not all the way to `--global-text-color`.

Issue #342, opened 2021-07-09
Deployment workflow enhancement


neerajgangwar:
The deployment workflow currently uses two sets of branches:

1. For the project pages: `master` branch for the source code and `gh-pages` branch for the generated website.
2. For personal or organization webpages: `source` branch for the source code and `gh-pages` branch for the generated website.

**Describe the solution you'd like**
The above-mentioned behavior can be unified. For both use cases, `master` can be used for the source code and `gh-pages` can be used for deployment.

**Describe alternatives you've considered**
NA

**Additional context**
NA

neerajgangwar:
I have raised a PR for this - https://github.com/alshedivat/al-folio/pull/343. I am still unsure about the manual deployments. Happy to hear suggestions.

neerajgangwar:
I have raised a PR for this - https://github.com/alshedivat/al-folio/pull/344.

ivanpuhachov:
Hi! I was following the previous instruction and `source` branch, and after sync with upstream, github actions are no longer triggered. What should I do to fix it?

neerajgangwar:
@ivanpuhachov Can you share your repo link?

ivanpuhachov:
Fixed it.

1. `.github/workflows/deploy` add `source` to the branches:
```
on:
  push:
    branches:
      - master
      - source
```
2. On Github - repo - Settings - Pages select `gh-pages` as a source branch – as it is written in current readme file. In old instructions it was `master`

neerajgangwar:
Yes, this is one way. But the idea was to have source code in the `master` branch. So you could also do the following:

1. Delete the `master` branch.
2. Rename `source` to `master`.
3. Push all the future changes to `master`. The generated site will be pushed to `gh-pages` which should be selected as the source branch in the publishing source.

ivanpuhachov:
Got it, thank you!

PR #344, opened 2021-07-09
Use master for the source code for personal/organization webpages

neerajgangwar:
This PR is for https://github.com/alshedivat/al-folio/issues/342 and makes the following changes:

1. Always push the generated website code to `gh-pages`.
2. Update the documentation for the personal/organization webpages.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -4,11 +4,9 @@ on:
   push:
     branches:
       - master
-      - source
   pull_request:
     branches:
       - master
-      - source
 
 jobs:
   deploy:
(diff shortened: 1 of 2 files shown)


alshedivat:
Thank you for the PR! I'm 100% in favor of simplifying deployment.

Can you please somehow confirm that deploying user and organization pages to `gh-pages` actually works? A while back, it used to be the case that user and organization pages had to be deployed to `master`. Did something change recently?

neerajgangwar:
@alshedivat Yes, even I remember this restriction. There is an option to configure a publishing source:
> https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#choosing-a-publishing-source

I am not sure if this was always there or recently introduced. I have set up my page using the same method - https://github.com/neerajgangwar/neerajgangwar.github.io

PR #404, opened 2021-09-24
add jekyll target blank plugin

pourmand1376:
pull request for issue #408 

This is also working perfectly on my own [website](https://pourmand1376.github.io/).

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -143,6 +143,9 @@ plugins:
   - jekyll-twitter-plugin
   - jekyll-sitemap
   - jemoji
+  #if you want to uncomment next line, you have to set base_url beforehand
+  #- jekyll-target-blank
+  # for more information: https://github.com/keithmifsud/jekyll-target-blank
 
 # Extras
 github: [metadata]
--- a/_config.yml
+++ b/_config.yml
@@ -15,7 +15,7 @@ footer_text: >
   Photos from <a href="https://unsplash.com" target="_blank">Unsplash</a>.
 
 icon: 🔥 # the emoji used as the favicon
-url:  # the base hostname & protocol for your site
+url: https://alshedivat.github.io # the base hostname & protocol for your site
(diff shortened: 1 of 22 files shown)


_config.yml
```diff
@@ -143,6 +143,9 @@ plugins:
   - jekyll-twitter-plugin
   - jekyll-sitemap
   - jemoji
+  #if you want to uncomment next line, you have to set base_url beforehand
```
alshedivat, line 146: how exactly should `base_url` be set? would be great to have this plugin enabled by default.

alshedivat (changes requested):
thank you for contributing this PR! manually setting `target='_blank'` was a pain point.

do you mind enabling this plugin by default and removing all manually added `target='_blank'` lines to html and markdown? would love to merge it right after this change!

_config.yml
pourmand1376: Sorry, that was a typo. I meant `url` should be set. Something like: `url: pourmand1376.github.io`. Otherwise, it would throw an error. 

As the [site](https://github.com/keithmifsud/jekyll-target-blank) says:
> By default. all anchor tags and markdown links pointing to an external host, other than the one listed as the url in Jekyll's _config.yml will automatically be opened in a new browser tab once the site is generated.

So, if we want to make this default, I think we would have to make `url` it mandatory. That would also cause some compatibility issues. Does this sound right to you?

alshedivat: I see, will have to investigate how to avoid compatibility issues. let's keep the PR open for the time being.

maskedvillain:
CRITICAL! 404 it's completely broken. image
@pourmand1376 @alshedivat @george-gca @

george-gca:
If you try the default repo example with a wrong url (https://alshedivat.github.io/al-folio/test), it works as expected. How have you tried this? Are you with the latest code base?

PR #431, opened 2021-10-18
Optimize dark theme

Edsuns:
### The original border was too bright and unattractive on dark theme. I made some changes and now it looks much better.

Note: I also changed the `footer_fixed` configuration, which probably shouldn't be merged.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -31,7 +31,7 @@ impressum_path:  # set to path to include impressum link in the footer, use the 
 # -----------------------------------------------------------------------------
 
 navbar_fixed: true
-footer_fixed: true
+footer_fixed: false
 
 # Dimensions
 max_width: 800px
--- a/_config.yml
+++ b/_config.yml
@@ -31,7 +31,7 @@ impressum_path:  # set to path to include impressum link in the footer, use the 
 # -----------------------------------------------------------------------------
 
 navbar_fixed: true
-footer_fixed: false
+footer_fixed: true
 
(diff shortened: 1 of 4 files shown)


_sass/_variables.scss
```diff
@@ -36,3 +36,4 @@ $black-color: #000000 !default;
 $code-bg-color-light:     rgba($purple-color, 0.05);
 $code-bg-color-dark:      #2c3237 !default;
 
+$dark-bg-color: #1C1C1D;
```
alshedivat, line 39: minor: could you please indent this the same way as `$code-bg-color-*` variables and add a new line at the end of the file?

_config.yml
```diff
@@ -31,7 +31,7 @@ impressum_path:  # set to path to include impressum link in the footer, use the
 # -----------------------------------------------------------------------------
 
 navbar_fixed: true
-footer_fixed: true
+footer_fixed: false
```
alshedivat, line 34: please revert this change

alshedivat (changes requested):
looks great, thank you for the contribution!

happy to merge if you could make a couple of small changes requested below. thanks.

_sass/_variables.scss
Edsuns: OK. The variable `$dark-bg-color` is supposed to be named with `$grey-color-dark`. I will make changes later.

alshedivat (approved):

PR #435, opened 2021-10-23
Optimize Website Load Time

pourmand1376:
Hi, 
As it is more efficient in term of loadtime speed, I placed the Javascript files after the body tag. 

Google search console is also showing promising results (15 percent better).

Before the test Google search console was showing **39%** for mobile, and after that, we are getting **54%**. Still, a long way to optimize it completely. 
image


alshedivat:
this is awesome! thanks for making these small optimizations.

PR #436, opened 2021-10-27
Update deploy.yml

jmrplens:
‘action’ is deprecated, change to ‘ruby’
https://github.com/ruby/setup-ruby


alshedivat:
Awesome, thanks for the fix!

PR #440, opened 2021-10-30
Delete assets/resized directory from GIT

pourmand1376:
The current configuration wastes a lot of space, and these files are unrelated to git.


alshedivat:
could you confirm that the resized files are correctly regenerated when the webpage is built by the github actions? I haven't had a chance to test that and hesitant to merge until this is tested.

rohandebsarkar:
This will not commit the `assets/resized` folder to GitHub pages.

A workaround could be to remove the `.gitignore` file by adding `rm .gitignore` in `bin/deploy`:
https://github.com/alshedivat/al-folio/blob/1e4e147d8fa346804bc36f71ca546185300c7576/bin/deploy#L97-L101

pourmand1376:
Sorry, I can not test that. I am using `Netlify` instead of `Github actions`.

jmrplens:
With PR https://github.com/alshedivat/al-folio/pull/498 that folder is no longer needed.

PR #465, opened 2021-11-20
jQuery should be load in head

jmrplens:
If jQuery load after footer get this error (chrome developer):

Screenshot of Google Chrome (20-11-21, 01-23-17)

Moved to head.

PR #466, opened 2021-11-20
Unnecesary .js

jmrplens:
Is included in misc.html

I had made [this pull request](https://github.com/alshedivat/al-folio/pull/465), but I was correcting it wrong. This is the correct solution.


jmrplens:
Today is not my day, I had an old version of the misc.html file, sorry!

PR #469, opened 2021-11-21
Fix broken read_time in blog/index.html

zhxchd:
Previously the `read_time` shown in blogs page is broken and always shows 2 minutes. This is due to the wrong variable name in `blog/index.html`. Simply changing `page` to `post` would fix it.


alshedivat:
thank you for the fix, appreciate it!

PR #516, opened 2021-12-30
Fix distill dark citation

rohandebsarkar:
Applies a minor fix for better readability of distill post citations in dark theme.

### DIFF
--- a/assets/js/distillpub/template.v2.js
+++ b/assets/js/distillpub/template.v2.js
@@ -2131,7 +2131,7 @@ d-appendix > distill-appendix {
 }
 
 .citation {
-  color: hsla(206, 90%, 20%, 0.7);
+  color: grey;
 }
 
 .citation-number {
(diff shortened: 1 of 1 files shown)


alshedivat (changes requested):
thank you for this enhancement!

I would strongly prefer not to edit `template.v2.js` directly (treat it as an imported library). instead, I'd suggest overriding anything distill-related using either via [`_sass/_distill.scss`](https://github.com/alshedivat/al-folio/blob/master/_sass/_distill.scss) or (for styles that distill loads dynamically) use [`assets/js/distillpub/overrides.js`](https://github.com/alshedivat/al-folio/blob/master/assets/js/distillpub/overrides.js).

alshedivat:
also, most of the issues with distill dark mode have been fixed in #299. looks like something got broken in one of the subsequent PRs. let me figure out and fix the breaking change.

rohandebsarkar:
Yes, #516 fixes it. Thank you.

PR #521, opened 2022-01-05
Update library versions (bootstap, jquery, fontawesome, academicicons, masonry, mdbootstrap)

rohandebsarkar:
Also, update the links to jsdelivr CDN, as they are more efficient in my experience.


rohandebsarkar:
**Note:** Could not update bootstrap v5 as it is incompatible with bootstrap v4 since it has some breaking changes.
More information: https://getbootstrap.com/docs/5.0/migration/

PR #524, opened 2022-01-08
Fix distill collapsible Navbar

junliang-lin:


rohandebsarkar:
Thanks for the fix! I was wondering, why was it not working.

PR #531, opened 2022-01-16
fixed title for og preview, otherwise its "blank"

alik-git:
So for the open graph preview, the metadata.html file is currently displaying `site.title` which is "blank" instead of `title` which is a variable that's created in lines 16-20, which sets the title to the name from the config.yml file, which is what should be used instead. So this change should fix the issue, I've tested it for my wesbite and it works. You can test the issue yourself using https://www.opengraph.xyz/

### DIFF
--- a/_includes/metadata.html
+++ b/_includes/metadata.html
@@ -33,9 +33,9 @@
 {%- if site.serve_og_meta %}
 
     <!-- OpenGraph -->
-    <meta property="og:site_name" content="{{ site.title }}" />
+    <meta property="og:site_name" content="{{ title }}" />
     <meta property="og:type" content="website" />
-    <meta property="og:title" content="{%- if page.title -%}{{ site.title }} | {{ page.title }}{%- else -%}{{ site.title }}{%- endif -%}" />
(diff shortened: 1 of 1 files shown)


_includes/metadata.html
```diff
@@ -33,9 +33,9 @@
 {%- if site.serve_og_meta %}
 
     <!-- OpenGraph -->
-    <meta property="og:site_name" content="{{ site.title }}" />
+    <meta property="og:site_name" content="{{ title }}" />
     <meta property="og:type" content="website" />
-    <meta property="og:title" content="{%- if page.title -%}{{ site.title }} | {{ page.title }}{%- else -%}{{ site.title }}{%- endif -%}" />
+    <meta property="og:title" content="{%- if page.title -%}{{ title }} | {{ page.title }}{%- else -%}{{ site.title }}{%- endif -%}" />
```
alshedivat, line 38:
```suggestion
    <meta property="og:title" content="{%- if page.title -%}{{ title }} | {{ page.title }}{%- else -%}{{ title }}{%- endif -%}" />
```

alshedivat (commented):
thanks for the fixes! looks like `site.title` wasn't changed to `title` in one of the places, but other than that LGTM and happy to merge.

PR #540, opened 2022-01-24
minor changes

vanshaj18:
As per https://developer.mozilla.org/en-US/docs/Web/CSS/@media/-webkit-transition, webkit-transition feature is no longer recommended. Though some browsers might still support it, it may have already been removed from the relevant web standards, may be in the process of being dropped, or may only be kept for compatibility purposes.

changed "webkit-transition" to "transition-property" in _base.scss


alshedivat:
Thank you for spotting this!!

PR #628, opened 2022-04-05
Add support for CV

rohandebsarkar:
The initial code was taken from: https://github.com/eaplatanios/eaplatanios.github.io/blob/00409c61fd173d2cab1c2d25188122a443f9e4b4/_pages/cv.md

I made some changes, mostly to make it work with dark mode.

I also added demo data for the CV.

This may need more work for formatting, etc.


buerviper (commented):
This is a real great addition, as it adds the CV in a similar style as the publication list. I have implemented some small changes which results in such a page:
https://buerviper.github.io/cv/

I just wonder - shouldn't the contents of cv.html and cv.md be exchanged? The *.md file now contains the layout which I think is not intended?

_sass/_base.scss
```diff
@@ -282,6 +282,22 @@ footer.sticky-bottom {
   font-size: 0.9rem;
 }
 
+// CV
+
+.cv {
+  .card {
+    background-color: var(--global-code-bg-color);
```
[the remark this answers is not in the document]
rohandebsarkar: This was intentional, since it looks better with the `code-bg-color`.

And also, the cards are indistinguishable with the background if both have the same color.

_pages/cv.md
```diff
@@ -0,0 +1,105 @@
...
+					<li class="list-group-item">
+						<div class="row">
+							{% if content.year %}
+								<div class="col-xs-2 cl-sm-2 col-md-auto text-left" style="width: 75px;">
+									<span class="badge font-weight-bold danger-color-dark text-uppercase align-middle" style="width: 75px;">
+										{{ content.year }}
+									</span>
+								</div>
+							{% endif %}
+							<div class="col-xs-10 cl-sm-10 col-md mt-2 mt-md-0">
```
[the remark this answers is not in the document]
rohandebsarkar: Thank you for your suggestion!

I was struggling to solve this particular issue.

_sass/_base.scss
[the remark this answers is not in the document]
buerviper: The cards still have a shaded frame, which sets them apart.

At least in al-folio's default color scheme,I don't think it works (it has black text on a dark blue background). In dark mode, it kinda pops. Maybe it comes down to taste in the end and how you set up your color scheme, but to me, it looked faulty.

rohandebsarkar:
> I just wonder - shouldn't the contents of cv.html and cv.md be exchanged? The *.md file now contains the layout which I think is not intended?

Yes, I agree. But in that case, the **empty** `cv.md` file must be there in `_pages` folder.

_sass/_base.scss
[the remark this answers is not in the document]
rohandebsarkar: I found that there was another variable named `--global-card-bg-color`, I felt it was more suited, so I changed the card colors to that.

alshedivat:
@rohandebsarkar, just wanted to say thank you for implementing support for CV, this is awesome! 🙌 
I'll try to review this PR as soon as possible and will hopefully have it merged shortly after that.

_data/cv.yml
```diff
@@ -0,0 +1,97 @@
...
+  contents:
+    - name: Full Name
+      value: Albert Einstein
+    - name: Date of Birth
+      value: 14th March 1879
+    - name: Languages
+      value: English, German
+
+- title: Education
+  type: table
```
alshedivat, line 12: could we change the name of this type to `time_table`? this would indicate that time information (the `year` field) has some special treatment.

_layouts/cv.html
```diff
@@ -0,0 +1,119 @@
...
+            
+            <div class="cv">
+              {% for entry in site.data.cv %}
+                <div class="card mt-3 p-3">
+                  <h3 class="card-title font-weight-medium">{{ entry.title }}</h3>
+                  <div>
+                  {% if entry.type == "list" %}
+                    <ul class="card-text font-weight-light list-group list-group-flush">
+                    {% for content in entry.contents %}
+                      <li class="list-group-item">{{ content}}</li>
```
alshedivat, line 22: nit
```suggestion
                      <li class="list-group-item">{{ content }}</li>
```

_layouts/cv.html
```diff
@@ -0,0 +1,119 @@
...
+                  {% elsif entry.type == "map" %}
+                    <table class="table table-sm table-borderless table-responsive">
+                      {% for content in entry.contents %}
+                        <tr>
+                          <td class="p-1 pr-2 font-weight-bold"><b>{{ content.name }}</b></td>
+                          <td class="p-1 pl-2 font-weight-light text">{{ content.value }}</td>
+                        </tr>
+                      {% endfor %}
+                    </table>
+                  {% elsif entry.type == "nested_list" %}
```
alshedivat, line 34: quick question: do you think it's possible to implement nested list (or any nested structure) using [recursion](https://jekyllrb.com/tutorials/navigation/#scenario-9-nested-tree-navigation-with-recursion)?

_layouts/cv.html
```diff
@@ -0,0 +1,119 @@
+---
+layout: default
+---
+<!-- _layouts/cv.html -->
```
alshedivat, line 4: this is a great page layout! thank you for contributing it. I noticed that you had to use a long list of if-then-else statements to determine the sub-layout block based on the entry type. I suggest splitting these sub-layouts into separate files that can be included, and keep these sub-components of the cv in `_includes/cv/*`. imo, this would improve readability and maintainability of this code. but what do you think?

(btw, some other long-ish layouts, such as `bib.html`, could benefit from the same refactor.)

alshedivat (commented):
thanks again for this awesome contribution! the cv page looks great both in light and dark modes. I left a few (mostly minor) comments / questions on the organization of the code.

rohandebsarkar:
@alshedivat thanks for your suggestions. I have resolved a couple of them, but I will be away for a week. Will try to take a look at them after I am back.

_layouts/cv.html
rohandebsarkar: I tried implementing this, but was unsuccessful, each list sub item has different classes and formatting. May be, to implement this we may have to change the style a bit.

alshedivat: thank you for giving it a shot! let's leave it as a future improvement.

alshedivat (approved):

PR #481, opened 2021-12-01
Better SEO, OpenGraph, schema.org and clean generated code

jmrplens:
I have made an html that generates all the information for [OpenGraph](https://ogp.me/) and also the [Schema.org](https://schema.org/) in head.html.

Right now, schema it is intended for personal web use, it would be necessary to add an option if the web is from a laboratory or group to get correct Schema.org.

[Schema test (my web).](https://validator.schema.org/#url=https%3A%2F%2Fjmrplens.github.io%2F)

[OpenGraph test (my web).](https://www.opengraph.xyz/url/https%3A%2F%2Fjmrplens.github.io/)

**UPDATE 1:**

Added HTML compression to remove blank lines generated by liquid code. Thanks to [jekyll-compress-html](https://github.com/penibelst/jekyll-compress-html).

**UPDATE 2:**

- Added: google and bing site verification
- Minor corrections

### [CHANGELOG](https://github.com/alshedivat/al-folio/pull/481#issuecomment-991958126)

PR #668, opened 2022-05-11
Add `jekyll-minifier`

rohandebsarkar:
[`jekyll-minifier`](https://github.com/digitalsparky/jekyll-minifier) provides a better way to minify HTML, CSS and JS files.

It reduces the file sizes by a considerable amount, resulting in faster loading of the webpages.


bin/deploy
```diff
@@ -91,6 +91,9 @@ then
 fi
 git checkout -b $DEPLOY_BRANCH
 
+# Export JEKYLL_ENV=production
+export JEKYLL_ENV=production
```
alshedivat, line 95: why is this necessary?

alshedivat:
super cool! thanks for adding `jekyll-minifier`.  do you know if this also solves issues with having to add white spaces to ensure the rendered html looks nicely formatted? (see relevant discussion here: https://github.com/alshedivat/al-folio/pull/481#issuecomment-1001252779.)

bin/deploy
rohandebsarkar: `jekyll-minifier` only works when `JEKYLL_ENV=production` is set.

This tells the plugin that this site is build for production, that way this won't work during testing but will work while deploying. This will save a lot of time during testing, as it takes a bit of time to minify all the files.

rohandebsarkar:
> super cool! thanks for adding `jekyll-minifier`. do you know if this also solves issues with having to add white spaces to ensure the rendered html looks nicely formatted? (see relevant discussion here: [#481 (comment)](https://github.com/alshedivat/al-folio/pull/481#issuecomment-1001252779).)

This plugin will not format the code to look good. It will delete all the whitespace and newline characters and also remove the comments. This will reduce the file size by a lot, but the code after being minified will not be in human-readable form. Said that, the browsers will auto format the code upon receiving, therefore it will not be a problem while inspecting the code. Moreover, this will only run while deploying and not while testing.

PR #687, opened 2022-05-27
Add a new webpage

brsata:


alshedivat:
thanks, but the link you are adding here leads to a modified version of the template, not to a personal webpage (while all links in that list correspond to personal webpages). I'm going to close this PR.

btw, the nordic color scheme in the modified looks very cool. would be awesome to add it as an option to al-folio :)

Issue #633, opened 2022-04-10
Ability to change default dark/light mode. 


ssprakhar:
**Is your feature request related to a problem? Please describe.**
The default display mode is Dark by default. There is a toggle to switch to Light, but I would prefer the Light to be default (still keeping the toggle feature intact) 

**Describe the solution you'd like**
A config flag to control the default mode. 

**Describe alternatives you've considered**
I disabled dark mode in config, so, website is now Light, but now the toggle button is gone. I would like that to still be present.

alshedivat:
I think there's a bit of a confusion here. the dark mode is not default. the theme "memorizes" the last setting of whoever views the webpage (storing it in the user's browser cache) and uses that setting whenever the website is visited again.

rohandebsarkar:
The default display is not dark mode. The website queries for the preferred color scheme by through the [`prefers-color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) option.

If `prefers-color-scheme` is set to `dark` then only the default mode is dark, otherwise the default mode is light.

https://github.com/alshedivat/al-folio/blob/f7274c0dab0aed6b0b1c0d7534779546478d1910/assets/js/theme.js#L55-L57

rohandebsarkar:
There is a minor bug. The default theme was actually always set to light theme without querying about the user preferences.

This should be fixed by #697.

PR #697, opened 2022-06-05
Fix theme

rohandebsarkar:
There was a minor bug. The default theme was actually always set to light theme without querying about the user preferences as discussed it was pointed out in #633.

Now it should be fixed.


MartensCedric:
Where can the user preferences be set as to change the default theme to dark?

PR #727, opened 2022-06-26
Adds support for GA4 analytics

rohandebsarkar:
Seems like GA4 analytics uses the same scripts as previous Google Analytics. 

Therefore, I don't think we need to make any more changes. Just added a note in the `_config.yml` and changed the ID to `G-XXXXXXXXXX` to make it clearer.

As Google will [stop supporting previous Universal Analytics](https://support.google.com/analytics/answer/11583528?hl=en&ref_topic=10094551), I did not mention it in code.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -79,7 +79,7 @@ contact_note: >
 # Analytics and search engine verification
 # -----------------------------------------------------------------------------
 
-google_analytics:  # your GA4 measurement ID (format: G-XXXXXXXXXX)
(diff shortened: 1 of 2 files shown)


rohandebsarkar:
I do not use Google Analytics, therefore cannot verify if this is working or not. If someone can confirm, I'll merge this PR.

pourmand1376:
I can confirm that this is working properly. Thanks!

Issue #701, opened 2022-06-10
Shorten the (very) long author lists


Randl:
For a very long author lists, the common solutions are either print all the names or only the first couple of authors. For personal website both are not very suitable.

I suggest adding an option to config with maximal number of authors. If the number exceed this number, the author list would look like
`A, B, C and 15 others including self` (where self is usual self for the publication page).

rohandebsarkar:
PR #732

PR #732, opened 2022-06-27
Adds support `max_author_limit`

rohandebsarkar:
As requested in #701.


alshedivat (commented):
could you add an example bib item that demonstrates `max_author_limit`?

rohandebsarkar:
In the hover, if there are more than tree authors, the authors are not distinguishable with one another. Will it be better to add a semicolon or an “and” and the end of all authors?

alshedivat (approved):
LGTM. added some minor improvements. feel free to merge if looks good.

Randl:
Thanks for the feature implementation!

I'd like to note that rendering of the full list is pretty slow for me for large amount of authors -- you can see it here https://evgeniizh.com/publications/

Also it would be nice if expanded list was formatted as the regular one and not with dashed underlining.

alshedivat:
@Randl, "428 more authors"... damn! 😄

- to speed up rendering, feel free to reduce the delay between rendering characters by editing line 115 of `_layouts/bib.html` and changing 15 ms to something like 0.1 ms, which should speed up rendering for you >100x.

- feel free to adjust the styling of the additional authors however you want by editing `_sass/_base.scss` lines 522-530.

hope this helps!

PR #734, opened 2022-06-28
Adds support for tag-list

### LINKED ISSUE #707: Blog archive / category tag list
With so many blog posts (currently 8 pages with 5 per page), it's become very difficult for casual readers to sort through and find posts which would be relevant to them. The newer tags/categories feature helps this a bit, but it's not so helpful if the latest post in a category is 4 pages in.
To fix this, initially, I converted all of my posts to projects. That way, every post on my blog is easily accessible, and readers can browse through all the different categories. The problem with that is that it's nigh impossible to tell when a new post is added, since the projects are sorted by importance, not date. So to fix that, I copied all the projects to blog posts as well (changing every file name and header...). Allowing users to see the newest posts on the blog, and all the posts in projects. Basically letting the projects page function as a blog archive. However, this means I need two copies of my posts with slightly different headers. Meaning, to make a new post I need to make two files, and if I want to change any of them, I have to change both files, which is inconvenient.

The simplest/most manageable solution I can think of might be to add a list of categories/tags which is always visible somewhere on the blog page so that users can click on a category even if the latest post for it was a while ago. 

Having a blog archive which lists every blog post on one page (sorted by categories rather than date) would be a more straightforward and helpful (but harder to implement) solution. But, since the projects page can already do this pretty well, just a copy of the projects page which is slightly modified to read from _posts rather than projects might not be too much more difficult?

Alternatively, an option to automatically put blog posts to projects (or vice versa) would be nice.  Or any changes to make blog posts/projects more compatible/interchangeable would go a long way. Such as removing dates from the url when they're in the file name, having category/categories function more identically, etc.

Also, a search bar and last page option (such as 1 2 3 4 ... 8) for example, would be helpful too (but now I'm just putting a wishlist of features...)

Thank you!

rohandebsarkar:
As requested in #707.

### DIFF
--- a/_sass/_base.scss
+++ b/_sass/_base.scss
@@ -332,6 +332,7 @@ footer.sticky-bottom {
 
   ul {
     justify-content: center;
+    display: flow-root;
 
     p, li {
       list-style: none;
--- a/_sass/_base.scss
+++ b/_sass/_base.scss
@@ -318,10 +318,26 @@ footer.sticky-bottom {
   border-bottom: 1px solid var(--global-divider-color);
   text-align: center;
   padding-top: 2rem;
-  padding-bottom: 5rem;
+  padding-bottom: 3rem;
   h1 {
     color: var(--global-theme-color);
     font-size: 5rem;
+  }
+}
+
+.tag-list {
+  border-bottom: 1px solid var(--global-divider-color);
+  text-align: center;
+
+  ul {
+    justify-content: center;
+
+    p, li {
+      list-style: none;
+      display: inline-block;
+      padding: 1rem 0.5rem;
+      color: var(--global-text-color-light);
+    }
   }
 }
 
--- a/_sass/_base.scss
+++ b/_sass/_base.scss
@@ -328,6 +328,7 @@ footer.sticky-bottom {
 .tag-list {
   border-bottom: 1px solid var(--global-divider-color);
   text-align: center;
+  padding-top: 1rem;
 
   ul {
     justify-content: center;
--- a/blog/index.html
+++ b/blog/index.html
@@ -20,6 +20,18 @@ pagination:
     <h2>{{ site.blog_description }}</h2>
   </div>
 
+  <div class="tag-list">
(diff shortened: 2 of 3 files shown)


blog/index.html
```diff
@@ -20,6 +20,18 @@ <h1>{{ site.blog_name }}</h1>
...
+      {% for tag in site.display_tags %}
+        <li>
+          <i class="fas fa-hashtag fa-sm"></i> <a href="{{ tag | prepend: '/blog/tag/' | relative_url }}">{{ tag }}</a>
+        </li>
+        {% unless forloop.last %}
+          <p>&bull;</p>
+        {% endunless %}
+      {% endfor %}
+    </ul>
+  </div>
```
alshedivat, line 34: could you make this block optional? if `site.display_tags` is not provided, this entire block should not be shown.

_sass/_base.scss
```diff
@@ -318,13 +318,29 @@ footer.sticky-bottom {
...
+  padding-bottom: 3rem;
   h1 {
     color: var(--global-theme-color);
     font-size: 5rem;
   }
 }
 
+.tag-list {
+  border-bottom: 1px solid var(--global-divider-color);
+  text-align: center;
```
alshedivat, line 330: let's add a bit more white space
```suggestion
  text-align: center;
  padding-top: 1rem;
```

_sass/_base.scss
```diff
@@ -318,13 +318,29 @@ footer.sticky-bottom {
...
   padding-top: 2rem;
-  padding-bottom: 5rem;
+  padding-bottom: 3rem;
   h1 {
     color: var(--global-theme-color);
     font-size: 5rem;
   }
 }
 
+.tag-list {
```
alshedivat, line 328: if you change the window size, the tag list does not behave responsively. could you make it responsive?

alshedivat (approved):
LGTM overall. thank you for adding tag-list support! left just a couple of small comments. would be great if you could make tag-list responsive before merging.

Issue #537, opened 2022-01-23
Images do not display on macOS Safari with 100% zooming level


HaoranLiao:
**Acknowledge the following**
- [x] I carefully read and followed the [Getting Started](https://github.com/alshedivat/al-folio#getting-started) guide.
- [x] I read through [FAQ](https://github.com/alshedivat/al-folio#faq) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [x] The issue I am raising is a potential bug in al-folio and not just a usage question. <br> [For usage questions, please post in the [Discussions](https://github.com/alshedivat/al-folio/discussions) instead of raising an issue.]

**Describe the bug**
My profile and project images do not display on macOS Safari (Version 13.1.3). The display was fine on Chrome and Firefox. And it was also fine on iOS Safari.

**Expected behavior**
Profile images and project card images displayed on the webpage

**Screenshots**
A screenshot of the profile image not displayed on macOS Safari: 
<img width="256" alt="Screen Shot 2022-01-23 at 11 15 49 AM" src="https://user-images.githubusercontent.com/16676497/150694622-bc22c718-9a5f-4bf3-b2d8-69478d7dac83.png">
A screenshot of the project card image not displayed on macOS Safari: 
<img width="603" alt="Screen Shot 2022-01-23 at 11 22 16 AM" src="https://user-images.githubusercontent.com/16676497/150694644-176a70d1-9197-4957-b4a0-4fdb83dc8ead.png">

**System (please complete the following information):**
 - OS: macOS
 - Browser (and its version) Safari 13.1.3
 - Jekyll version: 4.2.1
- Ruby version: 3.0.3p157 (2021-11-24 revision 3fb7d2cadc) [x86_64-darwin19]

**Additional context**

My profile and project images do not display on macOS Safari (Version 13.1.3). The display was fine on Chrome and Firefox. And it was also fine on iOS Safari. Whenever I use the macOS Safari to open my page, say the about page, I get this error:
```
[2022-01-23 11:00:17] ERROR `/assets/img/my_pic-480.webp' not found.
[2022-01-23 11:00:18] ERROR `/assets/img/my_pic-1400.webp' not found.
[2022-01-23 11:00:23] ERROR `/assets/img/my_pic-480.webp' not found.
```
[...]

PR #748, opened 2022-07-08
Fix responsive image

rohandebsarkar:
This finally addresses the problem in issue #537. This PR adds the support to remove responsive image blocks (`<source>` elements) if it encounters any error by using using jQuery.


alshedivat:
@all-contributors please add @rohandebsarkar for code

PR #750, opened 2022-07-09
Better bundler caching for GitHub Actions

rohandebsarkar:
References:
1. https://github.com/actions/cache/blob/main/examples.md#ruby---bundler
2. https://github.com/ruby/setup-ruby#caching-bundle-install-automatically


rohandebsarkar:
This PR will decrease GitHub Action consecutive build times by **~80%** with the same `Gemfile`.

PR #758, opened 2022-07-11
fix more authors bug

lufficc:
If more than one bib items trigger more author(s) action, it will effect each other due to  `$('span.more-authors')`.  Simply it change to `$(this)`  can solve this.


rohandebsarkar:
@lufficc, thanks for your contribution. Made a few quick changes to your branch. Hope you don't mind.

PR #770, opened 2022-07-16
Fix commas in author list

### LINKED ISSUE #749: Support equally contributed sign ^* on publications.
Hi, 

Thanks for providing this awesome persnal page template. I am wondering whether it is possble to add equally contributed sign ^* on publications, when it is needed to be shown on the list of authors. For example, Name1$^{\*}$, Name2$^{\*},$ .... 

Thanks!

gzwongkk:
The current author list has an issue with commas and spaces. 
For example, the first paper in publications

>Letters on wave mechanics
Einstein, Albert, [Schrödinger, Erwin](https://en.wikipedia.org/wiki/Erwin_Schr%C3%B6dinger),  [Planck, Max](https://en.wikipedia.org/wiki/Max_Planck) and 2 more authors
1967

There are actually two spaces in front of 'Planck, Max' and a missing comma before 'and'.
Moreover, the expanded author list after clicking on the '2 more authors' does not contain the 'and' conjunction.

This PR fixes the above problems and simplifies some of the redundant codes.
In particular, it prevents liquid from printing any arbitrary whitespaces and puts the 'and' into latter phrase.
A weird boundary case (i.e., author_array_limit is zero) is also considered and handled.

Last but not least, I still think it is a bit inconvenient to have the author's last name before the first name. 
Conventionally, it should be 'A. Einstein, B. Podolsky, and N. Rosen' instead of the current 'Einstein, A., Podolsky, B., and Rosen, N.' 
Although I have not changed this in the PR, it is recommended to review the current name order.

### DIFF
--- a/_layouts/bib.html
+++ b/_layouts/bib.html
@@ -11,13 +11,9 @@
           {%- endif -%}
         {%- elsif entry.abbr -%}
           {%- if site.data.venues[entry.abbr] -%}
-            {%- assign venue_style = "" -%}
-            {%- if site.data.venues[entry.abbr].color -%}
-            {%- assign venue_style = venue_style | append: 'style="background-color:' | append: site.data.venues[entry.abbr].color | append: '"' -%}
-            {%- endif -%}
-            <abbr class="badge" {{venue_style}}><a href="{{site.data.venues[entry.abbr].url}}">{{entry.abbr}}</a></abbr>
+          <abbr class="badge"><a href="{{site.data.venues[entry.abbr].url}}">{{entry.abbr}}</a></abbr>
           {%- else -%}
-            <abbr class="badge">{{entry.abbr}}</abbr>
+          <abbr class="badge">{{entry.abbr}}</abbr>
           {%- endif -%}
         {%- endif -%}
         </div>
--- a/_layouts/bib.html
+++ b/_layouts/bib.html
@@ -11,9 +11,13 @@
           {%- endif -%}
         {%- elsif entry.abbr -%}
           {%- if site.data.venues[entry.abbr] -%}
-          <abbr class="badge"><a href="{{site.data.venues[entry.abbr].url}}">{{entry.abbr}}</a></abbr>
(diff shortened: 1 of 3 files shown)


alshedivat (commented):
thank you so much for bug fixes + improvements!!

re: the "and" before "more authors" — this is a matter of style preference, but I would actually prefer having that "and" to keep the first K authors always separated from "more authors".

re: "last name, first name" ordering — this has come up a couple of times in the discussions. I have not strong preference here, so let's switch to more natural "first last" name ordering that you and others have requested. it would be awesome if you could make this change in this PR directly.

gzwongkk:
Got it. I have added a new global variable "max_author_separated" in _config.yml to address the style preference. When it is set to true, "and" will be shown before "more authors"; otherwise, "and" will be shown before the last author.

The "first last" name ordering is changed in the new commit as well.

_layouts/bib.html
```diff
@@ -29,76 +29,65 @@
...
           {%- if site.max_author_limit and author_array_size > site.max_author_limit %}
             {% assign author_array_limit = site.max_author_limit %}
-          {% else %}
-            {% assign author_array_limit = author_array_size %}
           {% endif %}
 
           {%- for author in entry.author_array limit: author_array_limit -%}
             {%- assign author_is_self = false -%}
-            {%- if author.last == site.scholar.last_name %}
+            {%- assign author_last_name = author.last | remove: "*" -%}
```
rohandebsarkar, line 39: I think other symbols such as `^` should also be removed.

gzwongkk: Thank you for the reminder. I took the ^ symbol as a math sign unconsciously. 
I have referred to the guideline by [PLOS ONE](https://journals.plos.org/plosone/s/submission-guidelines), in particular this [PDF document](https://journals.plos.org/plosone/s/file?id=3fac/PLOS%20Affiliations%20Formatting%20Guidelines.pdf).
The guideline suggested five common symbols for indicating special authorships.

PR #797, opened 2022-08-01
Support different colors for different kinds of publications in publication list

### LINKED ISSUE #541: Different colors for different kinds of publications in publication list
Currently, the journal/conference abbreviation is always the same (e.g., purple or blue in the standard light/dark schemes). I would however like to make a visual differentiation between, e.g, journal articles, preprints, conference contributions, or book chapters, and I think color would be the easiest.

I would like to define colors for certain contribution types. For instance, I would like to underly preprints in red. If I did not define a specific color for a specific content type, I'd like to use the default theme.

**Additional context**
The only thing I managed was changing the entire color scheme, but I was not able to change individual colors for individual contributions.

gzwongkk:
This PR attempts to solve the issue #541. 

The current `site.data.venues` used for `<abbr>` surprisingly comes without an example. I have added two in this PR.
I also make use of this venue mechanism and the liquid syntax to provide a simple modification for anyone in need for different color badges. This change simply overrides the class with `style="background-color: #XXXXXX"`.

Another work-around using bootstrap is described in the issue #541. 

This PR has a drawback that the color changes will not change accordingly with the light/dark themes.
Feel free to change anything or comment on whether this is needed.


alshedivat:
thank you for contributing this feature, it looks great!

one way to address the drawback is to add a bunch of variables for different color tags to https://github.com/alshedivat/al-folio/blob/master/_sass/_themes.scss (eg, call them `--global-badge-color-<color-name>`). can be done in a follow up PR.

PR #801, opened 2022-08-02
Add Pre-commit Hook

pourmand1376:
It is nice to have some pre-commit hooks for checking of spelling and things like this. If you want to see more information take a look at [here](https://calmcode.io/pre-commit/the-problem.html).

Notice that I have just created one file called `.pre-commit-config.yaml` and run `pre-commit install` and then created a commit. All other files are staged automatically for changes. 

There is also a [bot in github marketplace](https://github.com/marketplace/pre-commit-ci) which takes care of this pre-commit hooks in every pull request. So, we do not need to check for spelling errors and formatting styles. However, I can not install it to this repo, the owner himself should install it. To see an example of how it works take a look at one of my PRs at [yolov5](https://github.com/ultralytics/yolov5/pull/8766). 

[Here](https://pre-commit.com/hooks.html) is the list of pre-commit hooks we can add. We can add spelling and lots of others if you want.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,10 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.2.0
+    hooks:
+    -   id: trailing-whitespace
+    -   id: end-of-file-fixer
+    -   id: check-yaml
+    -   id: check-added-large-files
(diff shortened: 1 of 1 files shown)


pourmand1376:
Up

pourmand1376:
Hi, I figured out that I forgot to mention how to install this thing. Perhaps it is because It should be done automatically after each commit. 

To install it:
```bash
pip install pre-commit
pre-commit install
```

pourmand1376:
Hi, @alshedivat 
Thanks for merging. I just wanted to mention that this pre-commit hook is not useful unless you assign the bot to automatically check for every pull request. 

Also, It would be useful if you do:
```
pip install pre-commit
pre-commit install
```
And then do a dummy commit. This way this pre-commit hook gets activated on your local system and you see the changes it makes. For the first time, it would be a lot of files to commit. But after that, it would be less.

PR #823, opened 2022-08-16
Fix typos of about page circular p.p.

alpsencer:
Changed "cicular" to "circular"


alpsencer:
I put paranthesis on previous PR's name then I think it crashed the deployment. I fixed the name.

PR #853, opened 2022-08-30
A suggest to set right margin for profile picture

parhammrd:
setting profile picture in left side, cause touching content box and the picture without padding. 
I found no better why that doesn't effect on the right state option, but it could apply by condition.

thanks for your amazing project.

### DIFF
--- a/_sass/_base.scss
+++ b/_sass/_base.scss
@@ -94,6 +94,7 @@ blockquote {
 
 .profile {
   margin-left: 1rem;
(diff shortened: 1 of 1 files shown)


pourmand1376:
@parhammrd 
Can you take a screenshot to show the change?

rohandebsarkar:
Sorry but I am unable to understand your concern, can you elaborate the issue a bit more.

PR #871, opened 2022-09-06
Display booktitle for publications of type incollection.

beauby:
Until now, only publications of type `inproceedings` have the `booktitle` attribute shown, although `incollection` would benefit from it as well.


alshedivat:
thanks for this fix!

PR #875, opened 2022-09-11
Add Docker-compose file for windows compatibility 

### LINKED ISSUE #829: Cannot see the local version page using docker in README
OS: window 10 using MINGW

I want to check the edited page running local server with using `./bin/dockerhub_run.sh`.

After I run `./bin/dockerhub_run.sh` and connect the server, I see the second image below. (this means that cannot connect to site)

This is my process about I described above.
image
image

Thanks in advance !

pourmand1376:
This file makes it easier for windows users to use docker. (Closes #829)

Previous to this commit, those who used Windows had to install Ubuntu inside windows (via WSL) and run our commands. Now they can run it by just typing `docker-compose up`. 

> The main problem was that `./bin/dockerhub_run.sh` command was written with `Bash` in mind and you had to change it a little bit to make it compatible with windows `Powershell`. We shouldn't have two scripts. This is why adding a `docker-compose.yml` file is necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,12 +139,12 @@ $ cd <your-repo-name>
 Finally, run the following command that will pull a pre-built image from DockerHub and will run your website.
 
 ```bash
-$ ./bin/dockerhub_run.sh
+$ docker-compose up
 ```
 
 Note that when you run it for the first time, it will download a docker image of size 300MB or so.
 
-Now, feel free to customize the theme however you like (don't forget to change the name!). After you are done, you can use the same command (`bin/dockerhub_run.sh`) to render the webpage with all you changes. Also, make sure to commit your final changes.
+Now, feel free to customize the theme however you like (don't forget to change the name!). After you are done, you can use the same command (`docker-compose up`) to render the webpage with all you changes. Also, make sure to commit your final changes.
 
 <details><summary>(click to expand) <strong>Build your own docker image (more advanced):</strong></summary>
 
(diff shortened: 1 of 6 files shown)


pourmand1376:
This commit changes the command users should type if they want to use docker. I've understood that using `docker-compose` is pretty much standard and it works with all operating systems (specially Windows). 

> Note: we still keep previous commands for compatibility. 

With this commit, we can use docker pre-built image using:
```bash
docker-compose up
```

we can build a custom image with: 
```bash
docker-compose -f docker-local.yml build
```

and we can use locally built image using:
```bash
docker-compose -f docker-local.yml up
```

pourmand1376:
I added two parameters `--verbose` and `--livereload`. I have to see if those parameters add any problems here.

pourmand1376:
I have tested this commit in Windows, and it is working well. Any comments? @alshedivat @rohandebsarkar

rohandebsarkar (approved):

README.md
```diff
@@ -139,12 +139,12 @@ $ cd <your-repo-name>
 Finally, run the following command that will pull a pre-built image from DockerHub and will run your website.
 
 ```bash
-$ ./bin/dockerhub_run.sh
+$ docker-compose up
```
alshedivat, line 142: does this mean the user has to install `docker-compose`? if yes, let's add "install docker compose" step to the list on lines 131-132 above.

alshedivat (approved):
LGTM. thanks for adding `docker-compose` support. please see a small question about documentation below. once addressed, we can go ahead and merge.

README.md
pourmand1376: Yes. I also added `docker-compose` installation step.

Issue #313, opened 2021-06-14
Semantic Scholar Social Intergration


XuhuiZhou:
**Is your feature request related to a problem? Please describe.**
I think it would be good to have the option to add semantic scholar to the webpage.

**Describe the solution you'd like**
Have a place to fill in the semantic scholar ID in the social integration part in `_config.yml` like Google scholar.

**Describe alternatives you've considered**
Folks could modify their own version, but since semantic scholar is very common nowadays, it would be nice to have an existing option. 

**Additional context**

alshedivat:
definitely would be nice to have. don't use personally semantic scholar, but seems like this feature would require just a couple of lines of changes (adding a flag to `_config.yml` similar to google scholar and one line in `_includes/social.html` with a correct icon and url). would appreciate if someone could send a PR!

PR #879, opened 2022-09-13
Add Semantic Scholar social integration

dominikstrb:
This PR addresses issue #313 according to the suggestion by @alshedivat 

In my tests with my local build, the icon got pulled automatically from the Semantic Scholar site.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -72,6 +72,7 @@ gitlab_username: # your GitLab user name
 twitter_username: # your Twitter handle
 linkedin_username: # your LinkedIn user name
 scholar_userid: # your Google Scholar ID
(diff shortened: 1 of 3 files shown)


kargaranamir:
In this case, `http://www.semanticscholar.org/author/NUMERIC_CODE` redirects to `http://www.semanticscholar.org/author/AUTHOR_NAME/NUMERIC_CODE`.

dominikstrb:
I added semantic scholar also to metadata.html

About the author name and numeric code: I thought about adding a hint, but it works with just the id as well and redirects to the author page.

alshedivat (approved):
thanks for adding semantic scholar, this is great!

Issue #912, opened 2022-10-01
inproceedings bib items don't render fully


jveitchmichaelis:
**Acknowledge the following**
- [ X] I carefully read and followed the [Getting Started](https://github.com/alshedivat/al-folio#getting-started) guide.
- [ X] I read through [FAQ](https://github.com/alshedivat/al-folio#faq) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [ X] The issue I am raising is a potential bug in al-folio and not just a usage question. <br> [For usage questions, please post in the [Discussions](https://github.com/alshedivat/al-folio/discussions) instead of raising an issue.]

**Describe the bug**

Proceedings bib items are not correctly displayed. This seems to be a syntax error or a version mismatch somewhere.

The bib item template has the following code:

```
{%- elsif ["inproceedings", "incollection"] contains entry.type -%}
            {%- capture entrytype -%}<em>In {{entry.booktitle}}</em> {%- endcapture -%}
```

which seems to be valid Liquid syntax. However this doesn't render locally, nor does it render on a published site (e.g. built with the github pages action).

**To Reproduce**

Have a bibitem with an inproceedings entry, the `booktitle` is not shown. e.g. 

```
@inproceedings{mateo2019flood,
  abbr={NeurIPS},
  title={Flood Detection On Low Cost Orbital Hardware},
  author={Mateo-Garcia, Gonzalo and Oprea, Silviu and Smith, Lewis and Veitch-Michaelis, Josh and Baydin, At{\i}l{\i}m G{\"u}ne{\c{s}} and Backes, Dietmar and Gal, Yarin and Schumann, Guy},
  booktitle={Artificial Intelligence for Humanitarian Assistance and Disaster Response Workshop at NeurIPS},
  year={2019},
  doi={10.48550/arXiv.1910.03019},
  website={https://www.hadr.ai/previous-years/2019/home-2019}
}
```

which fails to render fully on my website: https://josh.veitchmichael.is/publications/

**Expected behavior**

Proceedings venue should be shown. This works with the following as a test, so it seems to be a syntax error with the list membership check.

```
{%- elsif entry.type == "inproceedings" -%}
```

Works if we assign the type list first, e.g.

```
{% assign proceedings = "inproceedings, incollection" | split: ','%}
```

and then check for membership in that.

**System (please complete the following information):**
 - OS: OS X
 - Browser (and its version): Not relevant, this is not present in the generated source.
[...]

PR #913, opened 2022-10-01
fix bug where proceedings do not render

jveitchmichaelis:
Aims to fix: https://github.com/alshedivat/al-folio/issues/912


alshedivat:
thank you for the fix!

PR #920, opened 2022-10-05
Improve bib layout for thesis

sim642:
### Changes
1. Use the nicer layout for thesis entries in bib layout by supporting the `school` field.
2. Add support for the `note` field.
3. Fix proceedings type lookup split.
4. Add space between journal/book/school and year if both present.

### DIFF
--- a/_layouts/bib.html
+++ b/_layouts/bib.html
@@ -104,10 +104,13 @@
 
         <!-- Journal/Book title and date -->
         {% assign proceedings = "inproceedings,incollection" | split: ','%}
+        {% assign thesis = "thesis,mastersthesis,phdthesis" | split: ','%}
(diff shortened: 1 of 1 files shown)


MortenHannemose:
I just tested this and it works as expected by printing the school for theses, and the rest of my bibliography also looks good. 
For item number 4, I assume you mean "comma" instead of "space".

Would it make sense to also use this pull request to fix the bug where there is a comma before the "and" when there is exactly two authors, or should I make a separate PR for that?

MortenHannemose:
@alshedivat should I open an issue to go along with this PR?

alshedivat (approved):

alshedivat:
thanks for the layout improvements. lgtm, merging!

Issue #927, opened 2022-10-07
`align:left` does not work as intended


subhodeeps:
**Acknowledge the following**
- [x] I carefully read and followed the [Getting Started](https://github.com/alshedivat/al-folio#getting-started) guide.
- [x] I read through [FAQ](https://github.com/alshedivat/al-folio#faq) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [x] The issue I am raising is a potential bug in al-folio and not just a usage question.

**Describe the bug**
Using `align:left` in the frontmatter of `about.md` generates a page where there is no gap/space between the profile picture and the text body.

**To Reproduce**
1. Open `about.md`
2. Set `align: left` in the front matter
3. See error

**Expected behavior**
There should be a space between the profile picture and text just like when one uses `align:right`

**Screenshots**

<img width="1025" alt="Screenshot 2022-10-07 at 5 54 58 PM" src="https://user-images.githubusercontent.com/49411080/194552725-1f2aa500-8cd3-49b1-9513-2c4a0f462901.png">

**System (please complete the following information):**
 - OS: macOS Monterey 12.3.1
 - Browser (and its version) : Google Chrome Version 106.0.5249.91 (Official Build) (arm64) and Safari Version 15.4 (17613.1.17.1.13)
 - Jekyll version : jekyll 4.2.2
- Ruby version: ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [arm64-darwin21]

**Additional context**

As  a work around, I have set 

```
// Profile
.profile {
  img {
    width: 88%;
 }
```
in  `_layout.scss` but I am sure that there is a better way to fix this.

PR #928, opened 2022-10-08
Set the margin according to the alignment selected

cmoralesmx:
As reported in #927, the style of the profile class only considers a left-margin regardless of the alignment option selected.
Thus, when the user changes the default alignment there is no corresponding margin to match their selection.
My commit addresses this by enabling the corresponding margin for the left or right alignment options

### DIFF
--- a/_sass/_base.scss
+++ b/_sass/_base.scss
@@ -93,7 +93,6 @@ blockquote {
 // Profile
 
 .profile {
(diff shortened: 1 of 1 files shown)


rohandebsarkar (approved):

cmoralesmx:
May I please have this PR tagged as `hacktoberfest-accepted` too?

PR #934, opened 2022-10-11
Add a progress bar to show the scroll position

cmoralesmx:
This feature adds a horizontal bar under the top menu which tracks the vertical scroll position. Such a feature can be useful to represent how much is left to read on the current page more aesthetically.

As this is an optional feature, `enable_progressbar` must be set to `true` in `_config.yml` to activate the functionality.

I am not the original author of this code. I just made it compatible with the current version of the template at the time of this commit. The original code was most likely authored by Pankaj Parashar in this [post](https://css-tricks.com/reading-position-indicator/) made a few years before the first inclusion in an `al-folio` site. Then, the code was adapted for compatibility with the template at Anthony Plantanios' site. Finally, I did
the last updates to have the code fit the new conventions used in the project.

This was discussed in #557

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -308,7 +308,7 @@ enable_navbar_social:       false  # enables displaying social links in the
 enable_project_categories:  true   # enables categorization of projects into
                                    # multiple categories
 enable_medium_zoom:         true   # enables image zoom feature (as on medium.com)
-
+enable_progressbar:         false  # enables a horizontal progress bar linked to the vertical scroll position
 
 # -----------------------------------------------------------------------------
 # Library versions
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -108,4 +108,12 @@
           </div>
         </div>
       </nav>
+{% if site.enable_progressbar %}
+      <!-- Scrolling Progress Bar -->
+      <progress id="progress" value="0">
+        <div class="progress-container">
+          <span class="progress-bar"></span>
+        </div>
+      </progress>
+{%- endif %}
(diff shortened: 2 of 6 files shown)


rohandebsarkar:
Great work, congratulations!

This works on most of the pages, the only page I saw this was not working perfectly was the `/projects` page. The progress bar stops about a quarter way through, though I was unable to deduce why this happens.

cmoralesmx:
Alright, thanks for your feedback. Regrettably, I did not check the functionality across all the default pages, sorry for that. I will fix this

cmoralesmx:
I identified the issue and updated the commit not to pollute the project's history with several commits for this single feature. The gist of the problem and the solution follow.
Using images larger than needed for the project's page causes a resize after loading to fit the target size.
However, as the distance to scroll is computed right after loading is completed, the values are higher than needed after the resize.
To fix that, I added a minimal delay to allow for the resize before computing the values.
Plus, I added better comments and improved the overall code quality.

rohandebsarkar:
Hey, can you also mention the article from where you took the code as a comment in the javascript file.

cmoralesmx:
Done, please let me know if the changes are OK

rohandebsarkar (approved):
LGTM

alshedivat:
apologies for the slow reply here and thanks for adding this feature!

my only suggestion would've been to reduce the height of the progress bar to 1px and turn it on by default in `_config.yml`. I'm going to go ahead and push these changes to master directly.

PR #945, opened 2022-10-14
Add WhatsApp to social integration

nopsqi:


alshedivat:
thanks for adding this social integration!

PR #946, opened 2022-10-15
Right to left writting is now possible

arashgmn:
This branch helps the template's internationalization by providing right-to-left (rtl) scripting. 

The `default` layout and all its derivates now automatically change their direction and alignment if a language with right-to-left scripting is specified in the front matter. The feature detects the following right-to-left languages (can be extended in `_config.yml`):

- Arabic (ar)
- Aramaic (arc)
- Azeri (az)
- Dhivehi/Maldivian (dv)
- Hebrew (he)
- Kurdish (ku)
- Persian (fa)
- Urdu (ur)

An example post is also provided in the Blog.

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -16,7 +16,8 @@
     {%- include header.html %}
 
     <!-- Content -->
-    <div class="container mt-5">
+    <div class="container mt-5" {% if site.rtl_langs contains page.lang %}dir="rtl" align="right"{% endif %}
+>
       {{ content }}
     </div>
 
--- a/_posts/2022-10-15-rtl.md
+++ b/_posts/2022-10-15-rtl.md
@@ -1,0 +1,74 @@
+---
+layout: post
+title: Right to left scripting
+date: 2022-10-15 
+description: you can also write from right to left
+lang: fa
+tags: internationalization formatting math code
+category: internationalization
+---
+
+# عنوان
+این یک پست نمونه برای سنجش نوشتارهای راست‌به‌چپ است.
+
+
+## فرمول‌نویسی
+فرمول‌ها هم مانند کد به درستی به نمایش در‌خواهند آمد:
+
+
+$$
+\sum_{k=1}^\infty |\langle x, e_k \rangle|^2 \leq \|x\|^2
+$$
+
+## کد
+کدها هم‌چنان مانند قبل از چپ‌به‌راست خواهند بود.
+
+```python
+if lang=="rtl"
+    print("I am a right-to-left script!")
+```
+
+## بلاک
+بلاک‌ها همواره چپ‌چین خواهند ماند:
+
+    ---
+    layout: page
+    title: project
+    description: a project with a background image
+    img: /assets/img/12.jpg
+    ---
+    
+حتی اگر به زبان‌های راست‌به‌چپ نوشته شوند:
+    
+    ---
+    صفحه‌بندی: صفحه
+    عنوان: پروژه
+    توضیحات: یک پروژه با تصویر زمینه
+    ---
+    
+## نقل‌قول
+> نقل‌قول‌ها به این شکل به نمایش درخواهند آمد.
+
+زیباتر خواهد بود اگر خط عمودی به راست انتقال داده شود. این کار با ویرایش _base.scss امکان‌پذیر است:
+
+```scss
+blockquote {
+  background: var(--global-bg-color);
+  border-left: 2px solid var(--global-theme-color); // change to border-right:
+  margin: 1.5em 10px;
+  padding: 0.5em 10px;
+  font-size: 1.2rem;
+}
+
+...
+
+
+.post-content{
+  blockquote {
+    border-left: 5px solid var(--global-theme-color); // change to border-right: ...
+    padding: 8px;
+  }
+}
+```
+
+منطقا باید بتوان با توجه به جهت‌گیری متن (rtl یا ltr) صفت مناسب را تعریف کرد. اما من چندان به sass مسلط نیستم. امیدوارم فرد مسلط‌تری این مشکل کوچک را رفع کند.
--- a/_sass/_layout.scss
+++ b/_sass/_layout.scss
@@ -6,7 +6,9 @@ body {
   padding-bottom: 70px;
   color: var(--global-text-color);
   background-color: var(--global-bg-color);
-
+  unicode-bidi: bidi-override !important;
+  direction: unset !important;
+  
   h1, h2, h3, h4, h5, h6 {
     scroll-margin-top: 66px;
   }
--- a/_config.yml
+++ b/_config.yml
@@ -228,7 +228,7 @@ jekyll-archives:
     tag: '/blog/tag/:name/'
     category: '/blog/category/:name/'
 
-display_tags: ['formatting', 'images', 'links', 'math', 'code'] # this tags will be dispalyed on the front page of your blog
+display_tags: ['formatting', 'images', 'links', 'math', 'code', 'internationalization'] # this tags will be dispalyed on the front page of your blog
 
 # -----------------------------------------------------------------------------
 # Jekyll Scholar
--- a/_config.yml
+++ b/_config.yml
@@ -96,6 +96,12 @@ discord_id: # your discord id (18-digit unique numerical identifier)
 contact_note: >
   You can even add a little note about which of these is the best way to reach you.
 
+
+# -----------------------------------------------------------------------------
+# Right-to-left scripts
+# -----------------------------------------------------------------------------
+rtl_langs: ["ar","arc","az","dv","he","ku","fa","ur"]
+
 # -----------------------------------------------------------------------------
 # Analytics and search engine verification
 # -----------------------------------------------------------------------------
--- a/_sass/_base.scss
+++ b/_sass/_base.scss
@@ -605,6 +605,8 @@ pre {
   background-color: var(--global-code-bg-color);
   border-radius: 6px;
   padding: 6px 12px;
+  text-align: left;
+  direction: ltr;
   pre, code {
     background-color: transparent;
     border-radius: 0;


_layouts/default.html
```diff
@@ -16,7 +16,8 @@
     {%- include header.html %}
 
     <!-- Content -->
-    <div class="container mt-5">
+    <div class="container mt-5" {% if site.rtl_langs contains page.lang %}dir="rtl" align="right"{% endif %}
```
alshedivat, line 19: I think it's also worth adding `"rtl"` class to this div, so that we can conditionally change CSS of some inner elements (eg, the border location of the blockquote).

_posts/2022-10-15-rtl.md
```diff
@@ -0,0 +1,74 @@
...
+\sum_{k=1}^\infty |\langle x, e_k \rangle|^2 \leq \|x\|^2
+$$
+
+## کد
+کدها هم‌چنان مانند قبل از چپ‌به‌راست خواهند بود.
+
+```python
+if lang=="rtl"
+    print("I am a right-to-left script!")
+```
```
alshedivat, line 29: what's the purpose of this?

_posts/2022-10-15-rtl.md
```diff
@@ -0,0 +1,74 @@
...
+
+## بلاک
+بلاک‌ها همواره چپ‌چین خواهند ماند:
+
+    ---
+    layout: page
+    title: project
+    description: a project with a background image
+    img: /assets/img/12.jpg
+    ---
```
alshedivat, line 39: same question here: does this showcase something?

_posts/2022-10-15-rtl.md
```diff
@@ -0,0 +1,74 @@
...
+حتی اگر به زبان‌های راست‌به‌چپ نوشته شوند:
+    
+    ---
+    صفحه‌بندی: صفحه
+    عنوان: پروژه
+    توضیحات: یک پروژه با تصویر زمینه
+    ---
+    
+## نقل‌قول
+> نقل‌قول‌ها به این شکل به نمایش درخواهند آمد.
```
alshedivat, line 50: I don't think blockquote renders correctly (the border is still on the left side)

_posts/2022-10-15-rtl.md
```diff
@@ -0,0 +1,74 @@
...
+...
+
+
+.post-content{
+  blockquote {
+    border-left: 5px solid var(--global-theme-color); // change to border-right: ...
+    padding: 8px;
+  }
+}
+```
```
alshedivat, line 72: instead of providing a suggestion on how to update CSS here, it's better the add rtl-conditional style overrides directly to `_scss/_base.scss`, like this (I haven't tested the code below, so please adjust as necessary):
```scss
rtl blockquote {
    border-left: none;
    border-right: 2px solid var(--global-theme-color);
}
.post-content {
    rtl blockquote {
        border-left: none;
        border-right: 2px solid var(--global-theme-color);
    }
}
```

_sass/_layout.scss
```diff
@@ -6,7 +6,9 @@ body {
   padding-bottom: 70px;
   color: var(--global-text-color);
   background-color: var(--global-bg-color);
-
+  unicode-bidi: bidi-override !important;
+  direction: unset !important;
```
alshedivat, line 10: sorry, I'm not as familiar with multi-directional language support. what does this do?

alshedivat (commented):
thanks for this contribution! really important step towards internationalization. please see some feedback and questions below. once addressed, I'd be happy to merge.

_layouts/default.html
arashgmn: Happy that you find it useful :) . I have to clarify one important fact: I'm by no means dominant on HTML, SCSS, and I'm afraid, even after going through Jekyll's documentation several times, I haven't fully comprehend its philosophy. So please take my response with a grain of salt, i.e., modify it as you feel it's better.

Coming back to your comment, I actually don't know how to add it to `div` here. If you provide me with some hints, I would be happy to work on it.

_posts/2022-10-15-rtl.md
arashgmn: It's merely a demonstration that shows code blocks stay left-to-right even if the language script is not. Now that I think of it, the code is a bit misleading... :)

_posts/2022-10-15-rtl.md
arashgmn: The same. Just for demonstration.

_posts/2022-10-15-rtl.md
arashgmn: I agree. As a matter of fact, the content of this post addresses this issue (although in Persian -- if you use chrome or firefox you may want to translate the whole page to English. Translation is decent in my experience). That said, I was unsuccessful in fixing this issue. Maybe adding `rtl` to `div` resolves this.

_posts/2022-10-15-rtl.md
arashgmn: I'll check this locally, and push a new commit if it worked on my side.

_sass/_layout.scss
arashgmn: TBH, I don't know much rather than `bidi` handles bi-directional scripting in the correct way. Often right-to-left scripts are not equal to a right-aligned version of a left-to-right script. Text becomes more shattered if there are numbers or special characters involved as well. Yet, with smart conditioning, one can prevent the misplacement of phrases. I guess, what bidi does is take care of all those cases.

Why these particular key-values are necessary to fix this, I don't know. I had [this](https://stackoverflow.com/q/19997153/7106957) problem and (almost) followed the accepted answer. In my case, it worked.

arashgmn:
The most recent push, fixes the location of the vertical line of blockquote with `&`: 
```
[dir=rtl] & {
        border-right: 5px solid var(--global-theme-color);
        border-left: none;
```

**Note**: It seems to me that the lateral border highlighting is repeated in the `_base.scss`. Namely, there's a command at [line 41](https://github.com/arashgmn/al-folio/blob/ff6f4e5728dccb519be655be865c496757865c8f/_sass/_base.scss#L41) and another one at the very end [line 661](https://github.com/arashgmn/al-folio/blob/ff6f4e5728dccb519be655be865c496757865c8f/_sass/_base.scss#L661). I'm not sure why is that. So, I added ```[dir=rtl] & {...` to both.

alshedivat:
@arashgmn, thanks again for the implementation of `rtl` support and for pushing a fix for the blockquote, it looks great! I'll take another pass, perhaps make some small adjustments to the code, and merge.

arashgmn:
Happy to help. And thanks for the awesome template!

_layouts/default.html
pourmand1376: Hi, Thanks for this great idea. 

It is a common practice ([+](https://stackoverflow.com/a/47376504/4201765)) to add `dir=rtl` in a separate div. So, maybe it is better to add it like this:
```html
<div {% if site.rtl_langs contains page.lang %} dir="rtl" endif %>
Stuff
</div>
```

Also, I see that you have added `align=right`. Why do you think that is necessary? I haven't seen it anywhere.

pourmand1376:
Also, we have a great font for writing `Persian (Farsi)` called `Vazirmatn`. It would be very good if we can somehow add custom fonts to this site. I'll just leave it as an idea.

_layouts/default.html
arashgmn: Hi, 
I was not aware of this practice, but it seems to work on my local machine. 

`align=right` is to decouple the text of each post from the body of the page which have some use cases. E.g., one might want to have the header and navigation in, say English, but the posts in a rtl script. An example post is [here](https://arashgmn.github.io/blog/2022/dialogue-prerequisites/).

arashgmn:
> Also, we have a great font for writing `Persian (Farsi)` called `Vazirmatn`. It would be very good if we can somehow add custom fonts to this site. I'll just leave it as an idea.

In the same blog I posted earlier, I use `Noto Naskh Arabic` (check `_include/head.html`). It can be changed to `Vazirmatn` easily.

The tricky part is to ensure `ltr` and `rtl` posts have the correct fonts. I used some layman's solutions to what didn'work, and it nearly works fine. The only exception is that the codes-block font is modified to the default `ltr` font except `mono`. This I couldn't fix, unfortunately. Would be happy to fix this one if you have any ideas.

arashgmn:
Any update on this? Should I modify anything?

george-gca:
I am not the best one to verify this, since I have no idea what are the rules when using left-to-right writing. @alshedivat @pourmand1376

arashgmn:
Hi,
I synced my fork to include the nice sidebar feature and adapted my repo. My new commits include the followings:

1. Adding "Vazirmatn" font for `rtl` content
2. ensuring the post-tags are always `ltr` (it looks better this way)
3. removing deprecated attributes defined for the earlier version 
4. Correct highlighting of the sidebar headers for active title or hover (by default, the vertical bar is placed on the right)
5. Ensuring that the code blocks are always `ltr`, left-aligned and in `monospace`

In addition, I added one new post to demonstrate that the sidebar indeed works an `rtl` language. 

Cheers, Arash

_posts/2022-10-15-rtl.md
arashgmn: It is entirely fixed in the latest commit.

_posts/2022-10-15-rtl.md
arashgmn: Please review my new commit (lines 71-74 and 971-977).

arashgmn (commented):
Please review my new commits and my recent reviews. And as before, thanks for the awesome template!

alshedivat:
Thanks for the contribution. Under v1 modularization, this feature should be developed as a standalone plugin rather than in the starter repo.

Tracking proposal issue opened: https://github.com/alshedivat/al-folio/issues/3544

I am closing this PR in favor of the plugin proposal path.

PR #978, opened 2022-11-11
Add backlink to verify mastodon

brunovianna:


rohandebsarkar:
Configuration for `jekyll-link-attributes` will need to be added into `_config.yml`: https://github.com/twinsunllc/jekyll-link-attributes#usage

Also, I think it would be better if the username and instance were included in the same variable, as discussed in https://github.com/alshedivat/al-folio/pull/974#issuecomment-1311447226.

brunovianna:
> Configuration for `jekyll-link-attributes` will need to be added into `_config.yml`: https://github.com/twinsunllc/jekyll-link-attributes#usage
> 
> Also, I think it would be better if the username and instance were included in the same variable, as discussed in [#974 (comment)](https://github.com/alshedivat/al-folio/pull/974#issuecomment-1311447226).

Done.

PR #986, opened 2022-11-15
New update to GH Actions: Deprecating `save-state` and `set-output` commands

adehghanzadeh:
Hi there

First of all, thanks for this amazing complete theme

Due to this warning which has been recently showed by Github, `::set-output` will be deprecated and must be replaced with a new method

image

topic: [GitHub Actions: Deprecating save-state and set-output commands](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)

Thanks


rohandebsarkar:
Thanks for the fix.

PR #992, opened 2022-11-18
Adds Telegram to social integrations

andrea-gasparini:
Hi, I realized Telegram is not present among the social integrations yet, I added a `telegram_username` entry and accordingly edited `social.html` and `metadata.html` and it correctly works on my side.

Anyway, thanks for the great work with this theme!


alshedivat:
thanks for adding support for telegram social link!

PR #1002, opened 2022-11-27
Hide the other opened box (Bib, Abs), if any

qomhmd:
The problem with the previous code is that whenever a box (especially abstract) is open, it is difficult to see that the other one (in the case, Bib) becomes open when one clicks on the related button.


alshedivat:
thanks for this improvement!

Issue #1005, opened 2022-11-29
Add youtube as a social option


p2635:
**Is your feature request related to a problem? Please describe.**
Not related to a problem, just in case people want to add youtube as a social.

**Describe the solution you'd like**
Implement in social.html.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #1007, opened 2022-11-29
Add youtube as a social option

p2635:
Relates to #1005


p2635:
I actually have no idea how the code is pulling through the icon in social.html, I just followed the examples for the other social icons. It's my first time doing these changes, so let me know if I did anything wrong!

alshedivat:
looks like it works. thanks for this contribution to al-folio!

PR #1081, opened 2023-01-06
Installation issues for Ubuntu>=20.04

rodosingh:
Are you facing issues with installing `Ruby`, `rubygems`, and `Bundler` of specific version in your `Ubuntu>=20.04` machine? Then this `PR` has some fixes regarding local setup.

> The `README.md` consists of required information.

*P.S.: I've added my webpage link to the table of `al-folio` users.*

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,6 @@ If you like the theme, give it a star!
 
 [![Preview](https://raw.githubusercontent.com/alshedivat/al-folio/master/assets/img/al-folio-preview.png)](https://alshedivat.github.io/al-folio/)
 
-
 ## User community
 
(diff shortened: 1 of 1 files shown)


pourmand1376:
I think Ubuntu problems can be solved via `Dockerfile` instructions easily since the base of the docker image is also Debian. 

https://github.com/alshedivat/al-folio/blob/a9685570e5889f1f57b3fac5d5ff8f79795b9bcb/Dockerfile#L3-L15

After that we can do:

https://github.com/alshedivat/al-folio/blob/a9685570e5889f1f57b3fac5d5ff8f79795b9bcb/Dockerfile#L24

That should be working fine.

PR #1085, opened 2023-01-07
Add news page

### LINKED ISSUE #993: WEBrick index page shows up when navigating to `<website-url>/news/`
I just set up the theme and while navigating among the pages to see which to include and which not, I figured out that there isn't any landing page dedicated to the news. This means that when one manually navigates to `<website-url>/news/`, without any actual posted news appended in the url, it will load the default WEBrick index page, whereas I would have expected a simple page containing all the news like it already works for the blog page.

**To Reproduce**
1. Locally [setup and run the theme](https://github.com/alshedivat/al-folio#local-setup-using-docker-recommended-on-windows)
2. Go to [localhost:8080/al-folio/news/](http://localhost:8080/al-folio/news/)
4. See the WEBrick index page

**Expected behavior**
A simple page containing all the news like it already works for the blog page.

**Screenshots**
Screenshot from 2022-11-18 13-33-03

 - OS: Ubuntu 20.04.3 LTS
 - Browser (and its version): Google Chrome 106.0.5249.103 (Official Build) (64-bit)
 - Jekyll version: using the Docker container
 - Ruby version using the Docker container

rohandebsarkar:
fixes #993 

Also added links to news and publications from respective section headers on the home page. Changed the colors of the header links to make it consistent with the previous look.


_layouts/news.html
```diff
@@ -0,0 +1,35 @@
...
+            {%- endif %} 
+          </td>
+        </tr>
+      {%- endfor %} 
+      </table>
+    </div>
+  {%- else -%} 
+    <p>No news so far...</p>
+  {%- endif %} 
+  </div>
```
alshedivat, line 34: looks like this code is mostly a copy-paste from `_includes/news.html`. instead of adding this extra layout, let's just add `{% include news.html %}` directly in `_pages/news.md`.

_pages/news.md
```diff
@@ -0,0 +1,5 @@
+---
+layout: news
+title: news
+permalink: /news/
+---
```
alshedivat, line 5:
```suggestion
---

{% include news.html %}
```

alshedivat (changes requested):
thanks for adding the news page, looks great! before merging, let's reconcile mostly identical code in `_layouts/news.html` and `_includes/news.html`.

_layouts/news.html
rohandebsarkar: Yes, it is copy-paste except for the h2 header `news`. If that is not removed there will be two `news` headers.

alshedivat: let's move the header to `_layouts/about.html` so that `_includes/news.html` can be included in `_pages/news.md` without creating the double header.

alshedivat (approved):

PR #1094, opened 2023-01-09
Installation issues for Ubuntu>=20.04

rodosingh:
Instructed some changes in `README.md` that account for installation issues with `Ruby`, `rubygems`, and `Bundler` of specific version in `Ubuntu>=20.04` machine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,6 +217,51 @@ $ bundle install
 $ bundle exec jekyll serve
 ```
 
+*Note: If you are in Ubuntu machine (>=20.04) and facing issues with `bundle install` and getting error message as follows:*
+> sass-embedded-1.57.1-x86_64-linux-gnu requires rubygems version >= 3.3.22, which is incompatible with the current version, 3.1.2
+then follow the steps below:
+
+```bash
+$ sudo apt-get install ruby-full
+# Create a folder by name ~/.rbenv and clone the rbenv repository into it.
+$ git clone https://github.com/rbenv/rbenv.git ~/.rbenv
+$ echo 'eval "$(~/.rbenv/bin/rbenv init - bash)"' >> ~/.bashrc # visit: https://github.com/rbenv/rbenv for more details
+
+# list the latest stable versions:
+$ rbenv install -l
+
+# list all local versions:
+$ rbenv install -L
+
+# install a Ruby version (e.g. 3.2.0 the latest stable version as of 2021-09-01):
+$ rbenv install 3.2.0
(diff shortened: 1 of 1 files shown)


pourmand1376:
Base Image for Dockerfile is `Debian`. We can easily use something like this for ubuntu 

https://github.com/alshedivat/al-folio/blob/aca95e563bc6968fe8897d72013bfdd93a1d315f/Dockerfile#L13-L24

Would be have problems this way?

I see your method is somewhat complicated. I can't understand why we have to use such complex scenario.

rodosingh:
> Base Image for Dockerfile is `Debian`. We can easily use something like this for ubuntu
> 
> https://github.com/alshedivat/al-folio/blob/aca95e563bc6968fe8897d72013bfdd93a1d315f/Dockerfile#L13-L24
> 
> Would be have problems this way?
> 
> I see your method is somewhat complicated. I can't understand why we have to use such complex scenario.

Hi @pourmand1376 thanks for pointing this out. Genuinely, I haven't tried out installation via `Dockerfile`. Since, the method mentioned in `Local setup` section doesn't seem to work, I thought of adding this PR which may help if someone is going for pure local setup.
- Plus, due to `Ruby>=3.2.x` onset, some of the `gems` (e.g., `nokogiri`) aren't installable via `bundler`, for which we are requiring `rbenv`.
- So instead of surfing through `rbenv` documentation, I appended what all `rbenv` commands one need to run to get the job done.

*Note: Since this PR is just for info, not any enhancement or solving bug, feel free to close this PR. And if any question arises regarding **local steup for Ubuntu** then as an answer we can put this PR link*

pourmand1376:
Okay, I think it is not a good idea to have installation script just for Ubuntu. 

But we should change the main standard way to reflect your changes also.

rodosingh:
> Okay, I think it is not a good idea to have installation script just for Ubuntu.
> 
> But we should change the main standard way to reflect your changes also.

pourmand1376:
What is your idea? @alshedivat @rohandebsarkar

rodosingh:
> Okay, I think it is not a good idea to have installation script just for Ubuntu.
> 
> But we should change the main standard way to reflect your changes also.

Thanks @pourmand1376 for your suggestions.

george-gca:
Maybe all this could be included in a wiki, like [this one](https://github.com/tqdm/tqdm/wiki), or in the [FAQ](https://github.com/alshedivat/al-folio#faq)?

rohandebsarkar:
This addresses a very particular problem (and this guide should probably only be followed if the user encounters the problem). I think this should be addressed in an issue and a link could be added in the FAQ section.

george-gca:
I didn't have this problem when running locally (on Ubuntu 22.04), but at the time the latest ruby version available was `3.1.2` and installing evertything was quite smooth.

PR #1167, opened 2023-02-09
Update README.md to encourage people to use `Use this template` 

KS1019:
This PR changes README.md to encourage people to use `Use this template` rather than `fork`.

### Cons
Possible con would be that it is harder to pull the new changes from this repo to the users' repos. It does not seem like impossible to do so but it will require additonal actions on the users end.
- https://github.com/orgs/community/discussions/23528
- https://stackoverflow.com/questions/56577184/github-pull-changes-from-a-template-repository

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,7 +209,7 @@ $ docker-compose -f docker-local.yml up
 
 #### Local Setup (Standard)
 
-Assuming you have [Ruby](https://www.ruby-lang.org/en/downloads/) and [Bundler](https://bundler.io/) installed on your system (*hint: for ease of managing ruby gems, consider using [rbenv](https://github.com/rbenv/rbenv)*), first [fork](https://guides.github.com/activities/forking/) the theme from `github.com:alshedivat/al-folio` to `github.com:<your-username>/<your-repo-name>` and do the following:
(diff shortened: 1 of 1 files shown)


george-gca:
It would be useful to add this as a solution to keep your repo up-to-date with the template:
https://github.com/AndreasAugustin/actions-template-sync

alshedivat (approved):
thank you for the readme update!

I meant to do it after switching the repo to template, but never go around to testing end-to-end that deployment is the same after clicking on "use this template." have you tried using this repo as a template for your personal webpage?

KS1019:
@alshedivat 

> thank you for the readme update!
> 
> I meant to do it after switching the repo to template, but never go around to testing end-to-end that deployment is the same after clicking on "use this template." have you tried using this repo as a template for your personal webpage?

I have been playing with this theme as a public theme and the only additional step I needed to do was to give GitHub Actions permission to write and read like [this](https://stackoverflow.com/questions/72851548/permission-denied-to-github-actionsbot)

You can check how it turned out [here](https://github.com/ks1019/test)

KS1019:
@alshedivat I added a new line about giving read and write permissions to GitHub Actions at 3ba39f9

PR #1168, opened 2023-02-09
Added related posts section

george-gca:
Created a related posts section. These are calculated based on how many of the same tags the posts have.
Added option to enable this in the post by adding `related_posts: true` in the header section of the post.

Source: https://heymichellemac.com/related-posts-jekyll

image

Signed-off-by: George Araújo <george.gcac@gmail.com>


rohandebsarkar (commented):
Wouldn't it be better if it was turned on by default.

george-gca:
Where should I put the explanation about how to disable it? Create a new blog post? Put somewhere on the README? Or just leave it disabled on some posts and the user will figure it out?

_includes/related_posts.html
```diff
@@ -0,0 +1,46 @@
+{% assign maxRelated = 5 %}
+{% assign minCommonTags =  1 %}
```
alshedivat, line 2: could you convert these constants (specifically: 5, 1) to parameters specified in `_config.yml`?

_layouts/post.html
```diff
@@ -43,6 +43,10 @@ <h1 class="post-title">{{ page.title }}</h1>
     {{ content }}
   </article>
 
+  {%- if page.related_posts == null or page.related_posts -%}
```
alshedivat, line 46: also add a check for the global flag (`site.related_posts.enable`)

alshedivat (commented):
this looks great, thank you for contributing this feature!!

I have a few small requests, but apart from that should be ready to merge.

alshedivat:
@george-gca, the issue with adding `related_posts` variable to `_config.yml` made me look into this, and it looks like jekyll already has support for related posts (see https://jekyllrb.com/docs/variables/#site-variables). `site.related_posts` is an internal variable that gets populated with a list of related posts automatically.

if you call jekyll build command with `--lsi` flag, it will use latent semantic indexing to create a list of up to 10 related posts. I would kind of prefer that over explicitly counting common tags (not all people would use tags) and we can make github action that builds the webpage use `--lsi` flag. what do you think?

george-gca:
@alshedivat completely agree. I thought before they did not have this implemented, since I never saw it being used elsewhere, and also found a lot of posts about implementing this manually. Will give it a try and update the commits.

george-gca:
According to the [docs](https://jekyllrb.com/docs/variables/#site-variables):

> Also note GitHub Pages does not support the `lsi` option when generating sites.

I think that might be outdated, because I could make it work. Submitting pr now.

alshedivat:
> Also note GitHub Pages does not support the lsi option when generating sites.

yes, since we use a custom github action for building the website, github pages limitations don't really apply.

Gemfile
```diff
@@ -1,23 +1,24 @@
 source 'https://rubygems.org'
 group :jekyll_plugins do
+    gem 'classifier-reborn'
```
alshedivat, line 3: I don't thinks this is necessary?

Gemfile
```diff
@@ -1,23 +1,24 @@
...
     gem 'jekyll-twitter-plugin'
     gem 'jemoji'
     gem 'mini_racer'
     gem 'unicode_utils'
     gem 'webrick'
 end
 group :other_plugins do
-    gem 'httparty'
     gem 'feedjira'
+    gem 'httparty'
```
alshedivat, line 23: what is this used for?

alshedivat (approved):

george-gca:
Classifier reborn is needed for the lsi, it is stated in the docs.

Regarding the other change, I have no idea. I just sorted them
alphabetically.

alshedivat:
> Classifier reborn is needed for the lsi, it is stated in the docs. Regarding the other change, I have no idea. I just sorted them alphabetically.

got it, thanks again for this contribution!

PR #1194, opened 2023-02-17
Update 404.html wording for accessibility

p2635:
Fixing in accordance to https://webaccess.berkeley.edu/ask-pecan/click-here

Also relates to https://github.com/alshedivat/al-folio/discussions/1188


p2635:
Or maybe I should call it 'go back to the home page manually'? I'm not sure what the best practice is.

alshedivat (approved):

PR #1226, opened 2023-03-03
Added information about using github action

george-gca:
By using this Github action it is possible to keep the derived repository up-to-date with its template.

I am currently using this to keep my website up to date with the template repository. An example of the PR this action creates can be seen [here](https://github.com/george-gca/george-gca.github.io/pull/13).


alshedivat (approved):
nice! thanks for this contribution 🙌

PR #1229, opened 2023-03-04
Added github 🏆 section in repositories

george-gca:
Basically, created a new section with your [user trophies](https://github.com/ryo-ma/github-profile-trophy#about-rank) under `repositories`, that looks like this:

image

It only displays the github user above when there is more than one.

image

Since there is room for [a lot of customization](https://github.com/ryo-ma/github-profile-trophy#optional-request-parameters) in the trophies, I decided to set some defaults. I set to only display trophies that are above than C level and used some margins that just felt right.

Also, I was shocked to see how many trophies @alshedivat has 🏆.

### DIFF
--- a/README.md
+++ b/README.md
@@ -488,7 +488,8 @@ Easily create beautiful grids within your blog posts and project pages:
 ### Other features
 
 #### GitHub repositories and user stats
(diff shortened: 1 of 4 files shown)


alshedivat (approved):
the new section looks great, thanks for this contribution!

before seeing your PR, I had no idea that github had some kind of trophies lol

PR #1234, opened 2023-03-05
Replace bin/deploy with GitHub Action

scottleechua:
### tl;dr
**This PR updates `deploy.yml` to use the GitHub Action [Deploy to GitHub Pages](https://github.com/marketplace/actions/deploy-to-github-pages).**

### Current setup
Currently, `deploy.yml` uses the deploy script `bin/deploy` to build and deploy the site. 

This script has worked well so far, but:
- (from experience!) is harder to read, and thus, customize;
- is another non-core script for maintainers to upkeep.

### Proposed setup
This Action replicates the default behavior of the deploy script: it clears the `gh-pages` branch, then pushes the contents of `_site` to it.

#### Changes:
- Replace `bin/deploy` with [Deploy to GitHub Pages](https://github.com/marketplace/actions/deploy-to-github-pages) Action.
- Create build step (`bundle exec jekyll build`) in `deploy.yml`.
- Enable deploy workflow to be manually triggered.
- _(Minor)_ Bump Checkout action from v2 to v3.
- _(Minor)_ Bump ruby version.

#### Benefits:
- Both automatic **and** manual deployment no longer depend on `bin/deploy`.
- Deploy logic is more explicit and easier to follow.
- Customization of deploy behavior (e.g., change branch site is served from) is now a matter of editing YAML, not shell scripts.

Thank you for this wonderful theme! I hope this PR helps even more folks use it. 😄 🚀

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -9,6 +9,7 @@ on:
     branches:
       - master
       - main
+  workflow_dispatch:
 
 jobs:
   deploy:
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -11,24 +11,26 @@ on:
       - main
   workflow_dispatch:
 
+permissions:
+  contents: write
+
 jobs:
   deploy:
     runs-on: ubuntu-latest
     steps:
-    - name: Checkout code
+    - name: Checkout 🛎️
       uses: actions/checkout@v3
     - name: Setup Ruby
       uses: ruby/setup-ruby@v1
       with:
         ruby-version: '3.2.1'
         bundler-cache: true
-    - name: Install deps
+    - name: Install and Build 🔧
       run: |
         npm install -g mermaid.cli
-    - name: Build site
-      run: |
         bundle exec jekyll build
-    - name: Deploy website
+    - name: Deploy 🚀
       uses: JamesIves/github-pages-deploy-action@v4
       with:
         folder: _site
+
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -30,6 +30,7 @@ jobs:
         npm install -g mermaid.cli
         bundle exec jekyll build
     - name: Deploy 🚀
+      if: github.event_name != 'pull_request'
       uses: JamesIves/github-pages-deploy-action@v4
       with:
         folder: _site
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -25,19 +25,10 @@ jobs:
     - name: Install deps
       run: |
         npm install -g mermaid.cli
-    - name: Setup deploy options
-      id: setup
+    - name: Build site
       run: |
-        git config --global user.name "GitHub Action"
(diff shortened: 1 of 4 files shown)


rohandebsarkar:
This is a better approach to my knowledge and could solve a lot of problems faced by new users regarding the deploy script.

alshedivat:
@scottleechua, thanks for this contribution!

I don't think deployment works as expected after this change, unfortunately. I tried fixing, but it still fails saying `remote: Permission to alshedivat/al-folio.git denied to github-actions[bot]`. do you know if there's a way to fix this?

scottleechua:
@alshedivat hello, it's a great template, happy to contribute!

I think this is not a code issue: just now I created a fresh al-folio fork, updated just the `deploy.yml`, and it created [this website](https://sentencespacetest.github.io/) without issues. So perhaps there is a permissions issue somewhere.

**Do you have branch protection on the `gh-pages` branch that prevents force-pushing?**

Looking at the error message in the [failed deploy workflows](https://github.com/alshedivat/al-folio/actions/runs/4457638232/jobs/7828825363), the workflow fails because it is trying to force-push into `gh-pages` and cannot:

```
Reset branch 'github-pages-deploy-action/jw1nfepa3'
/usr/bin/chmod -R +rw github-pages-deploy-action-temp-deployment-folder
/usr/bin/git worktree remove github-pages-deploy-action-temp-deployment-folder --force
Error: The deploy step encountered an error: The process '/usr/bin/git' failed with exit code 128 ❌
```

If that's not the issue, let me know. I'll keep thinking on my end as well. 😄 👍

rohandebsarkar:
I think the script is expected to fail in a pull request. It should only be successful in pushing to `gh-pages` when it has been invoked from a push in to `master` (or `main`) branch.

Previously we performed the PR check and deploy through the same script. But with the new script a condition has to be applied to not deploy when it is run for pull requests.

.github/workflows/deploy.yml
```diff
@@ -9,34 +9,28 @@ on:
...
-        elif [[ ${GITHUB_REF} = refs/heads/* ]]; then # branch, e.g. master, source etc
-          echo "SRC_BRANCH=${GITHUB_REF#refs/heads/}" >> $GITHUB_OUTPUT
-        fi
-        echo "DEPLOY_BRANCH=gh-pages" >> $GITHUB_OUTPUT
-    - name: Deploy website 
-      run:  yes | bash bin/deploy --verbose ${{ steps.setup.outputs.NO_PUSH }}
-                    --src ${{ steps.setup.outputs.SRC_BRANCH }} 
-                    --deploy ${{ steps.setup.outputs.DEPLOY_BRANCH }} 
+        bundle exec jekyll build
+    - name: Deploy 🚀
```
rohandebsarkar, line 32:
```suggestion
    - name: Deploy 🚀
      if: github.event_name != 'pull_request'
```

rohandebsarkar:
I have made the change to not deploy when invoked from a pull request. But I am not yet sure will it work when pushed into main branch.
But as @scottleechua, said it works on a fresh repository, I think this will work.

alshedivat:
ok, sounds good. we can always revert if something doesn't work.

PR #1253, opened 2023-03-11
Add mp4 and embeded video to figure.html

Kolompos:
Needed to show off some videos too so I made this. The CSS behind this needs to be made but since I am not fluent in web-dev I did not make it. Hope someone can help me with that...


george-gca:
I believe it would be better to split this, keep figure only for images and create a video for the videos, doesn't it?

alshedivat:
I agree, best to split into separate components for figures and videos.

george-gca:
Can you do it @Kolompos?

Issue #1262, opened 2023-03-15
Add a `Copy` button to code blocks


george-gca:
**Is your feature request related to a problem? Please describe.**
It would be really useful to have a Copy button appear in code blocks (preferably only when on hover).

**Describe the solution you'd like**
When hovering a code block, it should display a Copy button, and when clicking copy the code block's contents.

**Describe alternatives you've considered**

**Additional context**
The style of the copy button displayed [in this blog post](https://www.neoslash.net/blog/jekyllcopy) is interesting, although IDK if this is accessible. Found [this](https://fontawesome.com/icons/clipboard?s=light&f=classic) and [this](https://fontawesome.com/icons/clipboard-check?s=light&f=classic) icons that could be used for displaying. Also [this post](https://www.aleksandrhovhannisyan.com/blog/how-to-add-a-copy-to-clipboard-button-to-your-jekyll-blog/) has a really easy to implement solution, but the author did not made available the complete css it used.
Started to implement this using the code from this last post, but failed to position the button correctly (I struggled a lot with css customization), but the javascript works.

tostenzel:
I would love that feature, too!

alshedivat:
implemented in #1267

PR #1267, opened 2023-03-16
Implemented copy code button feature

george-gca:
Implemented support for copy code button in code blocks (#1262), also updated blog post about code to reflect it.

Here's how it looks like in light mode:

image

image

And in dark mode:

image

image

It does only appears when displaying code blocks using the markdown syntax, like:
````markdown
```c++
code code code
```
````

### DIFF
--- a/_layouts/post.html
+++ b/_layouts/post.html
@@ -57,3 +57,5 @@ layout: default
   {%- endif -%}
 
 </div>
+
+<script defer src="{{ '/assets/js/copy_code.js' | relative_url }}" type="text/javascript"></script>
--- a/_layouts/post.html
+++ b/_layouts/post.html
@@ -57,5 +57,3 @@ layout: default
   {%- endif -%}
 
 </div>
-
-<script defer src="{{ '/assets/js/copy_code.js' | relative_url }}" type="text/javascript"></script>
--- a/_posts/2015-07-15-code.md
+++ b/_posts/2015-07-15-code.md
@@ -39,3 +39,24 @@ int main(int argc, char const \*argv[])
 }
 
 {% endhighlight %}
+
+{% include copy_code.html %}
+```c++
+int main(int argc, char const \*argv[])
+{
+    string myString;
+
+    cout << "input a string: ";
+    getline(cin, myString);
+    int length = myString.length();
+
+    char charArray = new char * [length];
+
+    charArray = myString;
+    for(int i = 0; i < length; ++i){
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by george-gca** [comment]
Just now realized that my vscode formatted the css file differently. If you want to, I can undo the formatting.

File: _layouts/post.html [hunk 1]
```diff
@@ -57,3 +57,5 @@ <h1 class="post-title">{{ page.title }}</h1>
   {%- endif -%}
 
 </div>
+
+<script defer src="{{ '/assets/js/copy_code.js' | relative_url }}" type="text/javascript"></script>
```
alshedivat, line 61: this line should go to `_includes/scripts/misc.html`, which will then be included in the default layout: https://github.com/alshedivat/al-folio/blob/master/_layouts/default.html#L30

_posts/2015-07-15-code.md
```diff
@@ -39,3 +69,5 @@ int main(int argc, char const \*argv[])
 }
 
 {% endhighlight %}
+
+Note that when using the liquid tag, the `copy code` button is not displayed.
```
alshedivat, line 73: do you know why this is the case? in principle, is it possible to make `copy_code.js` work with both types of code blocks?

_sass/_base.scss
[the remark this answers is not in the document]
george-gca: Actually I use the VSCode default formatter, I just changed to format before saving the file. But adding a linter/formatter would be a nice idea to keep the code tidy. The VSCode formatter even change the current formatting of the README file every time I save, so I always undo some of them before commiting.

_posts/2015-07-15-code.md
george-gca: I am investigating this now, but apparently this is a [known issue](https://github.com/jekyll/jekyll/issues/6448) in Jekyll.

Under the hood though, I found out that this is the difference (when using the c++ example):
- Using markdown code block
```html
<pre class="highlight"><code>
```
- Using jekyll highlight
```html
<pre><code class="language-c--" data-lang="c++">
```

Maybe just removing the filter with `highlight` class works? Will try it now.

george-gca:
Made requested changes.
Now I realized that there is an extra space after code when using line numbers (don't know why this happens, but it always did).

image

alshedivat (approved):
lgtm and thanks again for the contribution! the extra space at the end of numbered code block was due to a spurious margin.

PR #1268, opened 2023-03-17
Avoiding specific warning on Google Chrome

george-gca:
Basically avoiding a warning that appears when opening the site on Google Chrome


alshedivat (approved):
lgtm, thanks for the fix!

PR #1282, opened 2023-03-21
Update Altmetric and Dimensions Badges

### LINKED ISSUE #1281: Harmonize Dimensions and Altmetric Badges

Dimensions and Altmetric badges do not align properly and are of visibly different style.

Use round Dimensions and Altmetric badges.  

This badge combination is the most consistent in its appearance. An alternative could be aligning the rectangle badges, but they color and resolution of the text remains different.

razvanazamfirei:
Enhancement to Altmetric and Dimension Badges
Fixes #1281


alshedivat:
@razvanazamfirei, thank you for contributing this!

as @george-gca mentioned in #1281, these badges look nice, but too big and colorful, which makes them not quite fit the style of al-folio. if there was a way to make them smaller and less colorful (ideally, single color that matches the theme color), that could've worked. from my understanding though, altmetric and dimensions provide very limited customization options for their badges, unfortunately. (maybe worth reaching out to them and asking for more flexible customization?)

I'll keep this PR open so more people can check it out and comment if they like or do not like to have alternative badges. if there is enough interest, we can add this version of the badges under a flag in the config.

Issue #1303, opened 2023-03-29
Copy code button keep fixed position when code width is larger than window


george-gca:
**Acknowledge the following**
- [X] I carefully read and followed the [Getting Started](https://github.com/alshedivat/al-folio#getting-started) guide.
- [X] I read through [FAQ](https://github.com/alshedivat/al-folio#faq) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [X] The issue I am raising is a potential bug in al-folio and not just a usage question. <br> [For usage questions, please post in the [Discussions](https://github.com/alshedivat/al-folio/discussions) instead of raising an issue.]

**Describe the bug**
When the code being displayed is larger in width than the window containing it, the button position doesn't scroll with it.

**To Reproduce**
1. Go to https://alshedivat.github.io/al-folio/publications/
2. Click on the BIB button on the first publication
3. Scroll right the bibtex code

**Expected behavior**
The button should keep its position based on the view, not in the content.

**Screenshots**
![image](https://user-images.githubusercontent.com/31376482/228686633-26c22a04-b9dd-428c-8871-6b051fd5639a.png)

![image](https://user-images.githubusercontent.com/31376482/228686660-112a92b7-a5db-4ca5-b1e6-41b3d0ad3ffc.png)

**System (please complete the following information):**
 - OS: Windows 11
 - Browser (and its version): Google Chrome 111.0.5563.112
 - Jekyll version: current version in example site
- Ruby version: current version in example site

PR #1348, opened 2023-04-23
Fix copy code button position

YifanJiang233:
Fix issue #1303 by adding a code wrapper outside the code block. 
Ref: https://stackoverflow.com/questions/60771472/position-a-button-such-that-it-wont-scroll-with-contents


george-gca:

PR #1367, opened 2023-04-25
Audio and Video support

george-gca:
Since #1253 has been hanging there for a while, I decided to part from it and add audio and video support.


alshedivat (approved):
looks awesome! thanks for adding support for audio and video media 🚀

Hkllopp:
So cool, thx !

Issue #425, opened 2021-10-14
Allow adding featured images to blog posts


atonkamanda:
How can we get pictures such as distll
![eAjEHq7bDRpX](https://user-images.githubusercontent.com/29407638/137230850-f40557a4-52e0-44d9-b2bd-43e6887d0ff8.png)
 
Near our blogpost in the blog section
![gDdGtdVVutbl](https://user-images.githubusercontent.com/29407638/137230952-554c117d-32ad-46d5-92ec-44446b8e6496.png)

atonkamanda:
I can't get which file I am supposed to modify and what am I supposed to do in order to do this

DoHaiSon:
It's a great improvement if we can add thumb or side img to blog post.
Clearly to reader and greatful if this thumb is loaded when we post link to FB, Tw.

lorak41:
Will this be added any time soon? 
Regards

aabs7:
You can just add a couple of lines of code in blogs [index.html](https://github.com/alshedivat/al-folio/blob/master/blog/index.html) like @alshedivat said.
REMOVE these lines in the [index.html](https://github.com/alshedivat/al-folio/blob/master/blog/index.html)
```
  <h3>
        {% if post.redirect == blank %}
          <a class="post-title" href="{{ post.url | prepend: site.baseurl }}">{{ post.title }}</a>
        {% else %}
        <a class="post-title" href="{% if post.redirect contains '://' %}{{ post.redirect }}{% else %}{{ post.redirect | relative_url }}{% endif %}">{{ post.title }}</a>
        {% endif %}
      </h3>
      <p>{{ post.description }}</p>
      <p class="post-meta"> {{read_time}} min read &nbsp; &middot; &nbsp;
        {{ post.date | date: '%B %-d, %Y' }}
      </p>
      <p class="post-tags">
        <a href="{{ year | prepend: '/blog/' | prepend: site.baseurl}}">
          <i class="fas fa-calendar fa-sm"></i> {{ year }} </a>

          {% if tags != "" %}
          &nbsp; &middot; &nbsp;
            {% for tag in post.tags %}
            <a href="{{ tag | prepend: '/blog/tag/' | prepend: site.baseurl}}">
              <i class="fas fa-hashtag fa-sm"></i> {{ tag }}</a> &nbsp;
              {% endfor %}
          {% endif %}

          {% if categories != "" %}
          &nbsp; &middot; &nbsp;
            {% for category in post.categories %}
[...]

PR #1371, opened 2023-04-26
Added thumbnail support for blog posts

george-gca:
Implemented #425 based on given code.

image

### DIFF
--- a/blog/index.html
+++ b/blog/index.html
@@ -53,7 +53,10 @@ pagination:
     {% assign categories = post.categories | join: "" %}
 
     <li>
-      <h3>
+      {%- if post.thumbnail -%}
+        <div class="row">
+          <div class="col-sm-9">
+<h3>
         {% if post.redirect == blank %}
           <a class="post-title" href="{{ post.url | prepend: site.baseurl }}">{{ post.title }}</a>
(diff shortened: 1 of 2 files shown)


blog/index.html
```diff
@@ -95,6 +98,55 @@ <h3>
               {% endfor %}
           {% endif %}
     </p>
+    </div>
+  <div class="col-sm-3">
+    <img class="card-img" src="{{post.thumbnail | relative_url}}" style="object-fit: cover; height: 90%" alt="image">
+  </div>
```
alshedivat, line 104: please put only this part should be inside `{% if post.thumbnail %} {% endif %}` so that the rest of the code is shared and not duplicated

alshedivat (changes requested):
thanks for adding this feature!

this PR introduces a lot of code duplication, which is a bad practice. let's avoid code duplication and refactor such that only the thumbnail-specific code is introduced (see suggestion below).

blog/index.html
george-gca: Agreed and done.

george-gca: Actually it was needed in two parts, to open and close the divs.

PR #1373, opened 2023-04-26
Fixed hard-coded bib files in templates

Lodour:
The bibliography filename `papers.bib` is hard coded in templates, and the equivalent setting in `_config.yml` is not used anywhere. So changing the bib filename in `_config.yml` won't have any effects.

This PR replaces hard-coded filenames with the value from `_config.yml`.


george-gca:

PR #1411, opened 2023-05-11
Fixed typo in blog post

danickblouin:


george-gca:
Thanks for noticing and fixing it

PR #1418, opened 2023-05-12
Fix typo ; Update _config.yml

furkanakkurt1335:


george-gca:
Thanks for noticing and fixing it.

furkanakkurt1335:
No worries, you got one too 🙂

PR #1424, opened 2023-05-15
Fix table color

george-gca:
Fix table color when dark mode is default.


george-gca:
> However, it seems to have affected the styling of the table within markdown pages. 

What you mean by that @lancewilhelm? If you check the [blog post about the tables](https://alshedivat.github.io/al-folio/blog/2023/tables/) in the demo it is working correctly. Also, I tried to add a table to another page, like `repositories`, and it displayed correctly.

lancewilhelm:
Ok I will check it out. Not sure how my tables were styled correctly before but not after. Thanks.

Issue #1249, opened 2023-03-09
The brute force filtering in Bibtex keywords will cause problem if the title also contain the keyword


timxzz:
**Acknowledge the following**
- [x] I carefully read and followed the [Getting Started](https://github.com/alshedivat/al-folio#getting-started) guide.
- [x] I read through [FAQ](https://github.com/alshedivat/al-folio#faq) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [x] The issue I am raising is a potential bug in al-folio and not just a usage question. <br> [For usage questions, please post in the [Discussions](https://github.com/alshedivat/al-folio/discussions) instead of raising an issue.]

**Describe the bug**

I have a BibTeX entry that contain the word `Autoencoders` in the `title`, which causes the resulting bibtex rendered on the website missing the entire `title` field. By digging into it a bit more, I found out that it's the substring `code` that causes the problem. Since `code` is a keyword in `filtered_bibtex_keywords`, the problem should be easily fixed in the regex of `hideCustomBibtex.rb`.

**To Reproduce**
1. Go to `papers.bib`
2. Add a word contain the substring `code`, e.g., `encode`,  into the title attribute.
3. In command line run `bundle exec jekyll serve --lsi`
4. Go to the website publication page and click on the corresponding `BIB` and see the title is missing.

**Expected behavior**
Keywords should only be filtered out if it's in the attribute of the BibTeX, rather then in the content of the attribute.

**System (please complete the following information):**
 - OS: [macOS Monterey v12.1]
 - Browser [Safari v15.2]
- Ruby version [3.1.3]

PR #1429, opened 2023-05-15
Only filter keywords in attributes

paberr:
I ran into the same issue as #1249 because one of my titles included "En*code*rs".

This PR fixes the problem that the keywords match any line that includes those and only considers the beginning of the line instead (i.e., the attribute in a properly formatted BibTeX file).

If this is not the expected behavior, I'm happy to adapt this minimalistic PR. 😄 
Any feedback is welcome.

### DIFF
--- a/_plugins/hideCustomBibtex.rb
+++ b/_plugins/hideCustomBibtex.rb
@@ -4,7 +4,7 @@
 	  keywords = @context.registers[:site].config['filtered_bibtex_keywords']
 
 	  keywords.each do |keyword|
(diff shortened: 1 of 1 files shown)


george-gca:
Tried to create this problem both trying to insert the word `Encode` in the beginning or the middle of a papers' title. Is this still an issue in the main repo @paberr?

alshedivat:
thanks for contributing this fix! looks like #1629 further improves upon it.
@paberr, I'm going to close this PR and merge #1629 instead, acknowledging your contribution.

PR #1432, opened 2023-05-18
Fixes "General Information" section of CV in dark mode.

jkschin:
## Problem

When the dark mode is default, or if you set the page to dark mode and then refresh, the content in "General Information" can't be seen.

Screen Shot 2023-05-18 at 10 46 23 AM

## Solution

This PR fixes the problem.

### DIFF
--- a/_includes/cv/map.html
+++ b/_includes/cv/map.html
@@ -1,8 +1,8 @@
-<table class="table table-sm table-borderless table-responsive">
-    {% for content in entry.contents %}
-      <tr>
-        <td class="p-1 pr-2 font-weight-bold"><b>{{ content.name }}</b></td>
-        <td class="p-1 pl-2 font-weight-light text">{{ content.value }}</td>
-      </tr>
-    {% endfor %}
+<table class="table table-sm table-borderless table-responsive" style="background-color: transparent; border: none; color: var(--global-text-color);">
+  {% for content in entry.contents %}
+    <tr>
+      <td class="p-1 pr-2 font-weight-bold"><b>{{ content.name }}</b></td>
+      <td class="p-1 pl-2 font-weight-light text">{{ content.value }}</td>
+    </tr>
+  {% endfor %}
 </table>
--- a/_includes/cv/map.html
+++ b/_includes/cv/map.html
@@ -1,4 +1,4 @@
-<table class="table table-sm table-borderless table-responsive" style="background-color: transparent; border: none; color: var(--global-text-color);">
+<table class="table table-sm table-borderless table-responsive table-cv-map">
   {% for content in entry.contents %}
     <tr>
       <td class="p-1 pr-2 font-weight-bold"><b>{{ content.name }}</b></td>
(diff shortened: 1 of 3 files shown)


assets/js/theme.js
```diff
@@ -49,44 +49,41 @@ let setHighlight = (theme) => {
...
     }
   }
 
-  setTheme(theme);
-}
-
+  // Wait for the DOM to finish loading before setting theme
+  document.addEventListener("DOMContentLoaded", function () {
+    setTheme(theme);
+  });
```
alshedivat, line 86: if I understand correctly, this is the main change. everything else is small refactors?

alshedivat: actually, this change seems to introduce flickering effect: when you switch to dark mode and then start clicking on other pages, it first loads the new page in the light mode and then switches it to dark mode, creating a weird transition.

alshedivat (changes requested):
please fix the flickering effect, and I'll be happy to merge.

assets/js/theme.js
jkschin: Yup. The other refactors were proposed by Prettier. Good catch on the flickering effect. Let me see how I can fix it.

jkschin:
@alshedivat I've posted a nicer fix. I did two things

1. I set the background of the table in `map.html` to transparent.
2. I set the color to follow the `global-text-color`.

This makes it look better now, without the flickering effect too.

_includes/cv/map.html
```diff
@@ -1,8 +1,8 @@
-<table class="table table-sm table-borderless table-responsive">
-    {% for content in entry.contents %}
-      <tr>
-        <td class="p-1 pr-2 font-weight-bold"><b>{{ content.name }}</b></td>
-        <td class="p-1 pl-2 font-weight-light text">{{ content.value }}</td>
-      </tr>
-    {% endfor %}
-</table>
\ No newline at end of file
+<table class="table table-sm table-borderless table-responsive" style="background-color: transparent; border: none; color: var(--global-text-color);">
```
alshedivat, line 1: do these styles have to be in-line or you can move them to `_sass/base.scss`?

alshedivat (approved):
LGTM, thanks again for the fix!

_includes/cv/map.html
jkschin: I think the classes like `table`, `table-sm`, etc. might be used in more than one place. I wasn't sure if this effect is something we want to apply globally. Tables could also appear in posts or other places.

Intuitively, it should be what we want since people who use the exact tables in dark mode would probably see nothing? But I can't be sure. I'll defer to your judgment on this.

jkschin: or actually we could create a `table-cv-map` class and then assign the styles there in `_sass/base.scss`?

alshedivat: yeah, let's create a separate class and assign it these styles.

jkschin:
@alshedivat this still can't be merged as a maintainer needs to approve.

alshedivat (approved):
my apologies for the slow reply, really appreciate you contribution! going to merge now.

PR #1449, opened 2023-05-26
Clarify property type for `google_site_verification`

lucasresck:
Domain property verification is the first (and more complete) option in [Google Search Console](https://search.google.com/search-console/welcome) to add a website property and it also calls the verification code "`google-site-verification`".

It took me a few days before realizing that `google_site_verification` in `_config.yml`
https://github.com/alshedivat/al-folio/blob/6e01a61a4ac1f39a8cef41d8bc249c54ef5187c5/_config.yml#L118
refers to the **HTML tag for URL prefix** property verification in Google Search Console:
https://github.com/alshedivat/al-folio/blob/6e01a61a4ac1f39a8cef41d8bc249c54ef5187c5/_includes/metadata.html#L4

This pull request expands `google_site_verification` comment to clarify Google Search Console property type.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -115,7 +115,8 @@ contact_note: >
 google_analytics:  # your Google Analytics measurement ID (format: G-XXXXXXXXXX)
 panelbear_analytics:  # panelbear analytics site ID (format: XXXXXXXXX)
 
-google_site_verification:  # your google-site-verification ID (Google Search Console)
+google_site_verification:  # your google-site-verification ID (HTML tag for URL
+                           # prefix property in Google Search Console)
 bing_site_verification:  # out your bing-site-verification ID (Bing Webmaster)
 
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


alshedivat:
@lucasresck, thanks for clarifying this. I thought `google-site-verification` would be just a string, not an HTML tag. could you clarify this by giving an example of how correctly specified `google_site_verification` field in `_config.yml` should look like?

lucasresck:
Hi, @alshedivat, thanks for the answer.

I added a new commit to further clarify how to get the code from the HTML tag.

`google_site_verification` is, indeed, just a string. When adding a new URL prefix property using HTML tag verification (the verification type used by `al-folio`), Google will ask to add the following HTML tag to the HTML page:
```
<meta name="google-site-verification" content="123456abcdef" />
```
An `al-folio` user does not need to update the HTML page source code. Instead, the user will only copy the `content` string and place it in `_config.yml`'s `google_site_verification`:
```
google_site_verification: 123456abcdef
```

alshedivat:
thinking more, I find the expanded description a bit unnecessary tbh. to keep things simple, I'll close this PR for now. if other users get confused about `google_site_verification` field, I'll consider revisiting this PR in the future.

PR #1451, opened 2023-05-26
Make latest post redirect if the blog post redirects

### LINKED ISSUE #1450: Latest blog post doesn't handle redirects
- [X] I carefully read and followed the [Getting Started](https://github.com/alshedivat/al-folio#getting-started) guide.
- [X] I read through [FAQ](https://github.com/alshedivat/al-folio#faq) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [X] The issue I am raising is a potential bug in al-folio and not just a usage question. <br> [For usage questions, please post in the [Discussions](https://github.com/alshedivat/al-folio/discussions) instead of raising an issue.]

**Describe the bug**

On the homepage, a latest blog post that redirects (internal or external) will not send the user to the `redirect` link when clicked and will instead link to the contents of the post, which will be pretty much empty. Then, the user will be sent back to the homepage in a few seconds.

**To Reproduce**

See screenshot below, which was taken after the following steps:

1. Check out the `master` branch of this repository
2. Delete `_posts/2022-12-10-giscus-comments` to `_posts/2023-05-12-custom-blockquotes.md` (7 files). This makes `2022-02-01-redirect.md` one of the latest posts.
3. Deploy the site locally

image

After clicking on the internal redirect page, you'll be sent to the blog post instead of the PDF it should redirect to:

image

Similar behavior for the external post. Also, the external post is missing the external post symbol.

**Expected behavior**

A latest post that redirects should redirect the user to the redirect URL (internal or external). Also, a latest post that redirects externally should have the external post symbol.

image

image

Lastly, the normal post that doesn't redirect also works normally.

image

 - OS: macOS
 - Browser: Safari 16.3
 - Jekyll version: 4.3.2
- Ruby version: 3.2.2

**Additional context**

I fixed the problem in #1451, as shown in the above screenshots. The immediately above "solution" screenshots were taken after checking out `fix-latest-posts` and steps 2 and 3 of the To Reproduce steps.

jesse-wei:
Fixes #1450, and the behavior of the site with this "new" code is shown in the Expected behavior screenshots there.

I copy-pasted the redirect logic from `blog/index.html` into `_includes/latest_posts.html`, with minor name changes.

I also cleaned up a line in `blog/index.html`.

### DIFF
--- a/_includes/latest_posts.html
+++ b/_includes/latest_posts.html
@@ -16,15 +16,13 @@
                   <td>
                     {% if item.redirect == blank %}
                       <a class="news-title" href="{{ item.url | relative_url }}">{{ item.title }}</a>
+                    {% elsif item.redirect contains '://' %}
+                      <a class="news-title" href="{{ item.redirect }}" target="_blank">{{ item.title }}</a>
+                      <svg width="2rem" height="2rem" viewBox="0 0 40 40" xmlns="http://www.w3.org/2000/svg">
+                        <path d="M17 13.5v6H5v-12h6m3-3h6v6m0-6-9 9" class="icon_svg-stroke" stroke="#999" stroke-width="1.5" fill="none" fill-rule="evenodd" stroke-linecap="round" stroke-linejoin="round"></path>
+                      </svg>
                     {% else %}
-                      {% if item.redirect contains '://' %}
-                        <a class="news-title" href="{{ item.redirect }}" target="_blank">{{ item.title }}</a>
-                        <svg width="2rem" height="2rem" viewBox="0 0 40 40" xmlns="http://www.w3.org/2000/svg">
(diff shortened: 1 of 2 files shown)


_includes/latest_posts.html
```diff
@@ -14,7 +14,18 @@
                 <tr>
                   <th scope="row">{{ item.date | date: "%b %-d, %Y" }}</th>
                   <td>
-                    <a class="news-title" href="{{ item.url | relative_url }}">{{ item.title }}</a>
+                    {% if item.redirect == blank %}
+                      <a class="news-title" href="{{ item.url | relative_url }}">{{ item.title }}</a>
+                    {% else %}
+                      {% if item.redirect contains '://' %}
```
alshedivat, line 20: nit: could you switch to using `elsif` pattern here instead of nested if-then-else? here's an example: https://github.com/alshedivat/al-folio/blob/master/_includes/head.html#L27

alshedivat (approved):
LGTM, thanks for this fix! just a small code style nit, happy to merge once fixed.

_includes/latest_posts.html
jesse-wei: @alshedivat Done

Issue #1395, opened 2023-05-03
Force updating cached style files


Lodour:
**Is your feature request related to a problem? Please describe.**
After updating the style file `main.css`, the browser may not update its cached older version of this style file.

**Describe the solution you'd like**
Adding a hash of the file content as the query string to the style file, like `main.css?xxxmd5hashxxx`.

**Describe alternatives you've considered**
It is possible to version the file with build time by `main.css?{{ site.time }}`, but that would force an update even when the style file was not modified.

**Additional context**
The modification would apply to the head template:
https://github.com/alshedivat/al-folio/blob/c1112d7e71c12572bb92ec082055ba28faa9e1b6/_includes/head.html#L30

EDIT: After some investigation, I find it tricky to implement this. Because the Jekyll filter might render before the sass preprocessor. Maybe using the `{{ site.time }}` would be better, otherwise it is more like a separate plugin beyond this repo.

george-gca:
This can be solved client side with a simple Shift + F5 on Google Chrome, to force update ignoring cache. Now, implementing this on the template seems kind of difficult.

Lodour:
Thanks for the response. I agree that implementing this is hard. I think we can close this issue if you feel the `{{ site.time }}` feature is unnecessary at this point.

But for the Shift + F5 solution, the main issue is that the visitors won't do that when they are not aware of the updated style files.

george-gca:
I agree that the visitors are not aware of the updated styles, but will the changes in the css files be that often? I mean, the changes might not happen that much to make this a burden.

Also, out of curiosity, how usually this situation is handled in other sites? Idk.

Lodour:
Yea the change won't be that often. I found this problem when I was frequently testing styles in deployment.

I think other sites usually version the CSS files with a number or its hash, as far as I can tell from https://stackoverflow.com/questions/1922910/force-browser-to-clear-cache

george-gca:
Found [this post](https://ultimatecourses.com/blog/cache-busting-jekyll-github-pages) that implements exactly what you said about a timestamp. But I believe a more elegant approach would be to calculate a hash from the file, since between updates to your site the css and js files might not get changed.

I found [this plugin](https://github.com/n13org/jekyll-kw-sri) that could be used for this, but could not make it work. Also tried getting the base code from [here](https://github.com/vcsjones/vcsjones.dev/blob/main/_plugins/sri.rb), but to no avail.

SanabilQureshi:
[...]

PR #1478, opened 2023-06-10
Update head.html to use cache-busting for CSS styling

SanabilQureshi:
Updated to address issue https://github.com/alshedivat/al-folio/issues/1395.

Implemented a cache-busting feature which refreshes the CSS url using a uniquely generated timestamp made when a new build occurs. This forces browsers to download/reference the updated styling.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -27,7 +27,9 @@
     {% elsif site.icon != blank %}
     <link rel="shortcut icon" href="{{ site.icon | prepend: '/assets/img/' | relative_url}}"/>
     {% endif %}
-    <link rel="stylesheet" href="{{ '/assets/css/main.css' | relative_url }}">
+
+    {% assign cacheBust = site.time | date:'?v=%s' %}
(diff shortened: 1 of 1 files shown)


Lodour:
Thanks for following up on this issue.

As discussed in the original issue, simply adding a time-based cache-busting would force reloading style files on every new build, even those without updating the style files. However, we were not able to make a hash-based cache-busting work.

I think the time-based approach is okay, as long as there is a switch in the theme config file to disable it. It might be better to have this feature disabled by default, given the unnecessary reloading for the more frequent content-only site updates.

alshedivat:
@SanabilQureshi, thanks for adding cache-busting option. as @Lodour mentions, this is a developer feature, so it should be disabled by default. what do you think about adding a `dev_mode` flag to `_config.yml` (set to false by default) and enabling cache-busting only if `dev_mode` is true?

george-gca:
I believe using [this plugin](https://github.com/yegor256/jekyll-git-hash) is the best option.

PR #1482, opened 2023-06-12
add functionality to display categories, like tags, in blog home page

furkanakkurt1335:
Added the functionality to display categories on the front page of the blog, just like tags.


george-gca:

PR #1487, opened 2023-06-13
Create unittest.py

yaellevy:
This PR was sent as part of OSDC taught by: @szabgab
I wanted to share my test contributions with you for the project.  
By testing the function (unit testing) you can prevent the program from doing unexpected things. The idea is that you test both good and bad input
 tests are all run automatically. This means that while you code, you can run the unit tests and verify that your code still works.

### DIFF
--- a/unittest.py
+++ b/unittest.py
@@ -1,0 +1,15 @@
+import unittest
+
+class TestStringMethods(unittest.TestCase):
+
+    def hello(self):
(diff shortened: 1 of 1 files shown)


szabgab:
@yaellevy I don't understand what is this PR. This is in Python in a project that is not in python and the code does not test anything. Is there some misunderstanding?

Issue #1440, opened 2023-05-22
Add an option to pin certain blog posts


alshedivat:
it would be nice to allow pinning certain blog posts at the top of the feed.

first of all, this would be useful for highlighting [the distill blog post layout](https://alshedivat.github.io/al-folio/blog/2021/distill/), which got pushed to the 3rd page of the blog posts feed on the demo page — this is one of the favorite features of many users, so it's worth highlighting it at the top. I'm sure pinned posts would be also of interest to some al-folio users that write a lot of posts and want to highlight their favorites.

PR #1498, opened 2023-06-22
Added support for featured blog posts

george-gca:
Implementing #1440. Still not sure if the year should be shown there or not. Also, I limited to be displayed at least 2 at most 3 elements on the row of featured blog posts, since when having only 1 featured post the card would occupy the whole row (and it looks weird). What do you think @alshedivat? Also, idk how to force the cards to have the same height. I think it would look nicer, but my lack of web dev skills made a difference here.

Some current screenshots:

image

image


george-gca:
One other idea is to limit the number of characters in the post description, by using something like [text-overflow: clip ellipsis](https://developer.mozilla.org/pt-BR/docs/Web/CSS/text-overflow) css property.

alshedivat:
this looks awesome! thanks for implementing. I think it's ok to allow post description to be 2-3 lines or less. +1 for clipping very long descriptions with ellipsis though.

blog/index.html
```diff
@@ -51,6 +51,49 @@ <h2>{{ site.blog_description }}</h2>
...
+                    </div>
+                    <h3 class="card-title text-lowercase">{{ post.title }}</h3>
+                    <p class="card-text">{{ post.description }}</p>
+
+                    {% if post.external_source == blank %}
+                        {% assign read_time = post.content | number_of_words | divided_by: 180 | plus: 1 %}
+                      {% else %}
+                        {% assign read_time = post.feed_content | strip_html | number_of_words | divided_by: 180 | plus: 1 %}
+                      {% endif %}
+                      {% assign year = post.date | date: "%Y" %}
```
alshedivat, line 78: nit: the indent is a bit off
```suggestion
                    {% if post.external_source == blank %}
                      {% assign read_time = post.content | number_of_words | divided_by: 180 | plus: 1 %}
                    {% else %}
                      {% assign read_time = post.feed_content | strip_html | number_of_words | divided_by: 180 | plus: 1 %}
                    {% endif %}
                    {% assign year = post.date | date: "%Y" %}
```

alshedivat (approved):
LGTM, thanks again for implementing! 🚀

RehmatSChawla:
For some reason, the post titles become lowercase when pinned, even when they are not so originally. Can this be fixed?

george-gca:
@RehmatSChawla since this PR is already closed, this should be asked in the Q&A section under discussion. To do this simply remove `text-lowercase` class here:

https://github.com/alshedivat/al-folio/blob/b30b3f4ec0c3223366c06183b49bdb8f0a95664c/_pages/blog.md?plain=1#L76

Issue #417, opened 2021-10-04
Support for jupyter notebooks in the posts/pages (using jekyll-jupyter-notebook).


alshedivat:

pietrolesci:
Hi @alshedivat,

Thanks a lot for this amazing project. 
I am in the process of migrating my website from hugo to al-folio. Since I have some blogposts written in jupyter notebooks, this functionality would really be helpful. Any update on this feature?

Thanks a lot in advance for your attention.

Best,
Pietro

liuh886:
Me too...This could be a very useful enhancement!

a good external solution I have seen is [this](https://michaelwornow.net/jupyter-notebook-to-markdown/), Hope it is useful.

george-gca:
@alshedivat I managed to implement this and I am willing to do a PR, but I have two main issues:

![image](https://user-images.githubusercontent.com/31376482/224865664-243ec4e3-e974-405f-8542-e94f970c4948.png)

1 - I used the [jekyll-jupyter-notebook](https://github.com/red-data-tools/jekyll-jupyter-notebook) you recommended here, but I currently can't test it in a docker environment. I installed everything in my machine (jupyter notebook) and then tested it only locally. What is the best way for me to test it with docker before submitting a PR, that doesn't involve installing docker locally?
2 - when the post can't find the converted jupyter notebook (the plugin calls the `jupyter` command to convert the notebook to an html), it displays inside the 404.html page, which ends up redirecting and loading the main page inside again. How can I check if the file exists before attempting to display it? Both the ipynb and converted html file are not displayed in the `site.static_files` variable.

![image](https://user-images.githubusercontent.com/31376482/224866004-0ee126eb-ff47-4703-b83a-9e44abdcc0b7.png)

![image](https://user-images.githubusercontent.com/31376482/224866044-a8bf1de7-0ca9-4079-bbc6-6a8cdf7f9b7e.png)

gil2rok:
As I wrote in issue #1405, this issue may be able to be fixed by using Quarto instead of Distill. Quarto seems to natively support posts and pages written in Jupyter notebooks. Simultaneously, Quarto is inspired by the Distill format. 

Let me know what you all think!

alshedivat:
@george-gca, thanks for taking a stab at implementing jupyter notebook rendering. do you have a branch / PR with the changes, would love to take a look to better understand the issues in the docker env and page not found.

[...]

PR #1511, opened 2023-06-25
Added support for jupyter notebooks #417

george-gca:
Implemented #417. Just don't know if everything that I need to run the code was added correctly in the `Dockerfile`. What is the best way for me to check for this?


george-gca:
Currently failing because it couldn't find `jupyter` command in docker image. How can we add jupyter to it? I believe it doesn't use the `Dockerfile` contained in the repo to create the image used.

george-gca:
@pourmand1376 can you add `jupyter` dependency to the docker image? I believe `apt install python3-jupyter-core` is enough.

pourmand1376:
> @pourmand1376 can you add `jupyter` dependency to the docker image? I believe `apt install python3-jupyter-core` is enough.

Have you tested this solution on your local docker?

george-gca:
I re-run the build check and it failed again.

The `Dockerfile` was updated, but not the docker image in the docker hub. If you check the image layers for the [latest tag](https://hub.docker.com/r/amirpourmand/al-folio/tags), you'll see that the `apt-get` lines are still the same.

pourmand1376:
Deploy will fail because you haven't added `jupyter` to github action. I do not know how to do this. 

Dockerhub image will update after we merge this request. In order to test it, use:
```
docker compose -f docker-local.yml up
```
and you will see that it builds an image and deploy the website locally.

george-gca:
Just realized that the GitHub action doesn't use your docker image. I always presumed it did.

Looking into [.github/workflows/deploy.yml](https://github.com/alshedivat/al-folio/blob/master/.github/workflows/deploy.yml) I saw this line `runs-on: ubuntu-latest`. After doing some digging, found that this image comes from [here](https://github.com/actions/runner-images), and it already has [these tools](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md) installed. Added [step to install jupyter there](https://github.com/alshedivat/al-folio/pull/1511/commits/7e50922f2cb7bbfba0bc22b28e2fc5d580633595) and apparently everything runned smoothly.

george-gca:
Just managed to test in a docker environment and it worked like a charm. Merging it into main.

lonzoweb:
I'm trying to delpoy in netlify but getting similar jupyter errors, added jekyll-jupter to gemfile and plugins but still getting: 

Missing dependency: jupyter

any ideas?

george-gca:
Do you have permission to install things outside ruby? If so, you are missing jupyter installation. Install it via `pip3 install jupyter`

Issue #1520, opened 2023-06-29
mini_racer problem


Abhijrathod:
Resolving dependencies...
Could not find compatible versions

Because mini_racer >= 0.8.0 depends on libv8-node ~> 18.16.0.0
  and Gemfile depends on mini_racer ~> 0.8.0,
  libv8-node ~> 18.16.0.0 is required.
So, because Gemfile depends on libv8-node ~> 20.2,
  version solving has failed.

bundle exec jekyll serve
Could not find gem 'mini_racer (~> 0.8.0)' in locally installed gems.
Run `bundle install` to install missing gems.

george-gca:
Can you provide further info? What is your setup, system version, etc?

Abhijrathod:
it is jekyll problem in installing bundle install problem ruby

pourmand1376:
Have you tested the docker image? It is just `docker compose up`.

Installing these dependencies manually is really painful.

Issue #1491, opened 2023-06-15
Fresh Installation Error


majauhar:
**Acknowledge the following**
- [x] I carefully read and followed the [Getting Started](https://github.com/alshedivat/al-folio#getting-started) guide.
- [x] I read through [FAQ](https://github.com/alshedivat/al-folio#faq) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [x] The issue I am raising is a potential bug in al-folio and not just a usage question. <br> [For usage questions, please post in the [Discussions](https://github.com/alshedivat/al-folio/discussions) instead of raising an issue.]

**Describe the bug**
I got the following error message when I tried `bundle exec jekyll serve --lsi` locally as well as on GitHub pages build
```
Incremental build: disabled. Enable with --incremental
      Generating...
"Fetching external posts from medium.com:"
                    ------------------------------------------------
      Jekyll 4.3.2   Please append `--trace` to the `serve` command
                     for any additional information or backtrace.
                    ------------------------------------------------
/usr/local/lib/ruby/gems/3.1.0/gems/feedjira-3.2.2/lib/feedjira.rb:63:in `parse': No valid parser for XML. (Feedjira::NoParserAvailable)
```
**System (please complete the following information):**
 - OS: [MacOS 12.3.1 (Intel)]
 - Jekyll version [4.3.2]
- Ruby version [3.1.2]

**Additional context**

george-gca:
I might be wrong here, but maybe `feedjira` have a lib dependecy that is not currently satisfied in the system (maybe `libxml2` idk).

majauhar:
I am trying to find what dependency might be unfulfilled. I have installed the dependencies listed in the `feedjira` website though. What's surprising for me is how it doesn't even run successfully inside GitHub since I would assume it should only throw errors, if any, when trying to run it locally.

majauhar:
Update: I disabled `imagemagick` and now Pages deployment works fine. However, even on fresh installation, the issue persists on my local machine.

majauhar:
That would be the obvious work around. However, since I did not modify the source code (except the repository name), it should have built without any issue. By the way, I am working via docker now and its working fine.

george-gca:
@majauhar agreed. But without having the same system I can't help that much. Also, apparently this is not a common error, I couldn't find any recent issues on this matter.

eguar11011:
I am on linux with manjaro and i have the same error

[...]

Issue #1469, opened 2023-06-07
Docker manifest for Mac Silicon 


cranmer:
**Is your feature request related to a problem? Please describe.**

On an apple silicon Mac I have installed docker-compose, but there is no docker manifest for this architecture. 

> $docker-compose up
> [+] Running 0/1
> ⠏ jekyll Pulling                                                                                                                                            0.9s 
> no matching manifest for linux/arm64/v8 in the manifest list entries

**Describe the solution you'd like**
A docker image that will work on apple silicon

**Describe alternatives you've considered**
I tried the standard install, but it is also having some issues related to mini-racer.

cranmer:
Follow up. Building locally with `docker-compose -f docker-local.yml up` worked fine. I'm not sure if there is a way to contribute that image to docker hub so others can pull it.

alshedivat:
thanks for reporting this. @pourmand1376, is there a way to push a new docker image built on apple silicon to the [dockerhub repository](https://hub.docker.com/r/amirpourmand/al-folio) that you are maintaining?

pourmand1376:
Hi, 

I do not have an M1 MacBook. I will soon borrow one from my friends to see why it doesn't work. The whole purpose of docker was to make our website cross-platform. I can not understand why It shouldn't work. 

We may have to build a docker image per CPU configuration. (One for intel, One for AMD, and One for M1). I have to see if there is an automatic way of creating these images via Github Actions. 

Still, One can push docker images to Dockerhub manually. However, manually pushing docker images is tedious and not a stable solution.

theophilegervet:
Hi @pourmand1376, I can't find the version compatible with `arm64/v8` at https://hub.docker.com/r/amirpourmand/al-folio. 
```
docker pull amirpourmand/al-folio
```
still gives
```
Using default tag: latest
latest: Pulling from amirpourmand/al-folio
no matching manifest for linux/arm64/v8 in the manifest list entries
```
Am I missing something?

pvbaudin:
Same issue, honestly I would consider this fixed if the README were modified to specify that docker-local.yml should be used to build for Apple Silicon based systems

Right now the docker-local instructions are under a dropdown that says
> Note: this approach is only necessary if you would like to build an older or very custom version of al-folio.

M1 and M2 are now common enough platforms that an explicit instruction linked int he TOC would be a good idea

pourmand1376:
Hi @theophilegervet 
[...]

PR #1523, opened 2023-06-30
Update README to make Docker Recommended way

pourmand1376:
Docker is well-tested for a long time. I think it is better to make it the recommended method for all OS types. 

This way, it is a lot easier to install docker (than to install jekyll and ruby and what not) and also we wouldn't have to deal with issues like these:
- https://github.com/alshedivat/al-folio/issues/1520
- https://github.com/alshedivat/al-folio/issues/1491


george-gca:
Is there a reason this PR is still marked as draft @pourmand1376? If not, let's accept it already.

george-gca (approved):

pourmand1376:
I am waiting until someone confirms that the docker image is successfully working on Mac Silicon. 

- #1469

Issue #1524, opened 2023-06-30
Generate Pagespeed Insights from GitHub Action


george-gca:
**Is your feature request related to a problem? Please describe.**
Currently the Pagespeed image is "hard coded" into the README file, meaning that the image must be manually updated.

**Describe the solution you'd like**
It would be better if it was generated from a [GitHub Action](https://github.com/ankurparihar/readme-pagespeed-insights), this way if a commit made the page way slower this could be rapidly visualized.

pourmand1376:
I agree. This is a really good feature.

george-gca:
Tried to use this [GitHub Action](https://github.com/ankurparihar/readme-pagespeed-insights/blob/master/.github/workflows/pagespeed.yml), but the url of the herokuapp is not working anymore. Probably it'll be needed an action that depends on the output of [PSI Action](https://github.com/stefafafan/psi-action) and draws the svg as done in [README PageSpeed Insights](https://github.com/ankurparihar/readme-pagespeed-insights).

george-gca:
Finally found one such action and was able to implement it in my repo. There are only a few issues:

1. it must be done by @alshedivat since a [repository secret](https://github.com/MyActionWay/lighthouse-badger-workflows#lighthouse-badger-easyyml) is needed, but the rest of the steps I can create a PR with it. You can check this working [in my repo](https://github.com/george-gca/multi-language-al-folio?tab=readme-ov-file#lighthouse-pagespeed-insights)
2. I couldn't make it generate automatically after a successful page build. Probably I have to set it to look for page builds in the `gh-pages` branch, but I will check that later. 
3. This is not such a big problem. I disabled the PWA (Progressive Web Apps) part, but it still generates that in the png, but at least is not showing a terrible value there (even though doing a PWA is not the point in this template).

PR #1530, opened 2023-07-02
Revert "Created GitHub action to update site pagespeed svg"

george-gca:
Reverts alshedivat/al-folio#1529


alshedivat:
I was going though PRs and noticed that we reverted #1529 but never added it back. what was the issue with it?

george-gca:
I created #1529 to test the GitHub action because I could not find a way to test it in isolation. But then I realized that the heroku app it uses is no longer online. I detailed a possible solution [here](https://github.com/alshedivat/al-folio/issues/1524#issuecomment-1616903293).

PR #1538, opened 2023-07-06
Added support for jupyter dark mode

george-gca:


awsaf49:
Hi @george-gca, thank you for the update. It appears that changing the website from light to dark mode does not affect the theme of the Jupyter notebook. Does this mean we need to manually set the dark mode in the notebook using `jt -t <theme-name>`?

george-gca:
Are you sure @awsaf49? Try opening it up on an anonymous tab, and check if it works. Sometimes it is the browser cache that gets in the way. Opening the page and pressing ctrl + F5 forces a reload cleaning the cache, that fixes the issue.

awsaf49:
After clearing the cache, it works on this repo but doesn't work on my forked one. I also tried anonymous tab but had no luck. I simply added your committed changes to my repo.

page: https://awsaf49.github.io/blog/2022/gcvit/
repo: https://github.com/awsaf49/awsaf49.github.io

It looks something like this,
image

george-gca:
Here it is displaying as expected

image

awsaf49:
Weird it is working now. Thanks btw =)

george-gca:
I just tried it again on an anonymous tab, and apparently the dark mode sometimes doesn't work on first load. I will open an issue for that. Thanks for noticing that.

Issue #1513, opened 2023-06-26
Json-Resume integration for the CV part


arunoruto:
**Is your feature request related to a problem? Please describe.**
We can all agree that writing the same things at multiple places can be tedious. So we do it once and try to reuse it everywhere else. That is what [jsonresume](https://jsonresume.org/) aims for resumes. Write all information once in a json file, put it on github gist and use it everywhere else!

**Describe the solution you'd like**
Provide a jsonresume link in the config to be loaded and populate everything in the CV section according to the schema. For that, we can use the [jekyll-get-json](https://github.com/brockfanning/jekyll-get-json) plugin like:
```yaml
jekyll_get_json:
  - data: resume
    json: "https://gist.githubusercontent.com/<username>/<id>/raw"
```
And use the data with `{{ site.data.resume }}`.
To keep compatibility with previous setups, one can prioritize the jsonresume data. If not found, populate it using the yaml file.

**Describe alternatives you've considered**
The obvious alternative is to copy all the data into the yaml file <yikes>.

**Additional context**
I am currently trying out to port the styling for jsonresume. The file will be a lot longer/messier, but maybe someone has an idea how to make it more sleek.

**EDIT**
I tried to implement it on my own, and you can see the result [here](https://academic-5lq.pages.dev/cv). The relevant changes have been made in [include](https://github.com/arunoruto/academic/tree/master/_includes/resume) with a new folder resume (time table and map have been modified), and the [cv layout](https://github.com/arunoruto/academic/blob/master/_layouts/cv.html) has been split in two.

I know, it isn't beautiful, but it works for now. Maybe someone has a more beautiful way of how to manage it?

george-gca:
Never heard of it, but it looks interesting. Definitely worth taking a look.

arunoruto:
I sat down and made for each category its appropriate html file (for those I need and deemed necessary). You can find them in the [resume](https://github.com/arunoruto/academic/tree/master/_includes/resume) folder in my fork of al-folio. I also modified the [cv.html](https://github.com/arunoruto/academic/blob/master/_layouts/cv.html) layout to also take into consideration the jsonresume data and prioritize it. Some sections can be skipped, like the meta files and publications, since thats already available in a seperate bib file in al-folio. In my opinion it is a nicer solution, but maybe the bib file can be generated using data from jsonresume, but jsonresume lacks some informations in that regard for now...

[...]

PR #1339, opened 2023-04-19
CV Restyled

jmrplens:
I have redesigned the CV to have some options that I have in my pdf CV.

I'm not too inspired to include more example data, if anyone wants to add data, go ahead.

With my data the CV looks like this: [https://jmrplens.github.io/cv/](https://jmrplens.github.io/cv/ ).
My data: [VIEW FILE](https://github.com/jmrplens/jmrplens.github.io/blob/ce8b5b67dbc5a37dafe9083681b92a6746640b13/_data/cv.yml)

Some screenshots:
Screenshot of Google Chrome (19-4-23, 23-09-51)Screenshot of Google Chrome (19-4-23, 23-10-22)
Screenshot of Google Chrome (19-4-23, 23-10-37)Screenshot of Google Chrome (19-4-23, 23-10-54)

Mobile:

Screenshot of Google Chrome (19-4-23, 23-21-36)Screenshot of Google Chrome (19-4-23, 23-21-48)
Screenshot of Google Chrome (19-4-23, 23-22-00)Screenshot of Google Chrome (19-4-23, 23-22-14)

PR #1547, opened 2023-07-09
Add jsonresume support

arunoruto:
Just as discussed in #1513 , this pull request adds support for [jsonresume](https://jsonresume.org/). Create the resume once, either as a file in the repository or at [github gist](https://gist.github.com/) called `resume.json`. Put the file in the `_config.yaml`  file and that's it! Other platforms like [resumake](https://latexresu.me/) use the same schema.

I also incorperated the changes introduced by #1339 to the best of my abilites. The style could be further improved.

**Please merge this pull request after #1339 is merged, due to dependencies from it**

If someone has a bettet approach on how to solve the problem that each section needs its own template, please let me know. But for now it works fine and is still backwards compatible with the `cv.yaml` file.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,6 +15,7 @@ group :jekyll_plugins do
     gem 'jekyll-sitemap'
     gem 'jekyll-toc'
     gem 'jekyll-twitter-plugin'
+    gem 'jekyll-get-json'
     gem 'jemoji'
     gem 'mini_racer'
     gem 'unicode_utils'
--- a/_config.yml
+++ b/_config.yml
@@ -230,6 +230,7 @@ plugins:
   - jekyll-sitemap
   - jekyll-toc
   - jekyll-twitter-plugin
+  - jekyll-get-json
   - jemoji
 
 # Sitemap settings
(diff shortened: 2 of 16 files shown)


george-gca:
Nice work! Thanks for this feature.

george-gca:
@arunoruto just now I realized that not all info available in [cv.yml](https://github.com/alshedivat/al-folio/blob/master/_data/cv.yml) was replicated to [resume.json](https://github.com/alshedivat/al-folio/blob/master/assets/json/resume.json). Could you add the missing pieces and also exclude the `cv.yml` file? Or should we keep it as another option? What do you think @pourmand1376?

t4rf9:
It seems the old _data/cv.yml is not working, is it able to have both json and yaml as alternative options? I want the yaml cv back

george-gca:
Afaik the old `_data/cv.yml` is still working. [Here](https://github.com/alshedivat/al-folio/blob/c24ed74855a86d4e67008095ddfe59d08aa8c084/_layouts/cv.html#L5) you can see that we kept it as a fallback if it doesn't find the jsonresume defined.

PR #1561, opened 2023-07-14
Change analytics script from Panelbear to Cronitor

yahaha-233:
Panelbear is now Cronitor RUM, the old script does not work.


george-gca:
Thanks for noticing this.

Issue #1329, opened 2023-04-12
Improve Mobile Performance


varuniyer:
**Is your feature request related to a problem? Please describe.**
It would be nice if the performance score from Pagespeed Insights was higher for mobile devices. Even though the desktop performance score is 93, the mobile performance score is only 54.

**Describe the solution you'd like**
Looking at the Pagespeed advice, it seems the stylesheets are blocking render time on mobile. There may be a way to reduce the performance drop this has on mobile.

**Describe alternatives you've considered**
Alternative academic websites like the Hugo templates have better performance scores on mobile.

varuniyer:
I found I was able to improve the performance number by using https://purifycss.online/ to remove unused CSS from large stylesheets for things like material design bootstrap. I also moved all these assets previously fetched from jsdelivr to my repo. This brings the mobile performance up, but it's not consistently hitting 90+ on pagespeed.web.dev. I found the JS associated with some of these stylesheets like mdb.min.js also could be shrunken down since most functions there aren't being used. However, there are no tools like purifycss that I can use to remove these unused functions. Any tips on how one might go about removing the unused functions?

varuniyer:
I'd like to update this saying I've managed to boost my mobile Pagespeed Insights score significantly by using the UnCSS CLI to generate smaller stylesheets that I store locally. I also improved the score by commenting out the script tag for mdb.min.js without experiencing any loss of usability. However, this may not be feasible for all al-folio users.

george-gca:
Maybe this solution could be incorporated. What do you think @alshedivat @rohandebsarkar @pourmand1376?

> I found I was able to improve the performance number by using https://purifycss.online/ to remove unused CSS from large stylesheets for things like material design bootstrap. I also moved all these assets previously fetched from jsdelivr to my repo. This brings the mobile performance up, but it's not consistently hitting 90+ on pagespeed.web.dev. I found the JS associated with some of these stylesheets like mdb.min.js also could be shrunken down since most functions there aren't being used. However, there are no tools like purifycss that I can use to remove these unused functions. Any tips on how one might go about removing the unused functions?

About `mdb`, once I removed and got some weird layouts, but don't exactly remember where.

varuniyer:
[This](https://github.com/episource/jekyll-uncss) may help incorporate performance improvements from removing unused CSS. As for removing mdb, I only removed mdb.min.js so the necessary CSS is still there.

rohandebsarkar:
[...]

PR #1562, opened 2023-07-14
Added support to PurgeCSS

george-gca:
Tackled #1329 with [PurgeCSS](https://purgecss.com/).

Being talking with @varuniyer about using [jekyll-uncss](https://github.com/episource/jekyll-uncss) to reduce css file sizes by ditching unused classes. This approach have 3 main problems:

1 - have some limitations as pointed [here](https://github.com/alshedivat/al-folio/issues/1329#issuecomment-1546517327)
2 - last update to [jekyll-uncss](https://github.com/episource/jekyll-uncss) was about 3 years ago, so it might have a few issues
3 - [uncss](https://github.com/uncss/uncss) haven't seem a new release in a while, currently [lacking maintenance](https://github.com/uncss/uncss/issues/459), and using some deprecated libraries as seem here:

```
npm install -g uncss
npm WARN deprecated request-promise-native@1.0.9: request-promise-native has been deprecated because it extends the now deprecated request package, see https://github.com/request/request/issues/3142
npm WARN deprecated har-validator@5.1.5: this library is no longer supported
npm WARN deprecated w3c-hr-time@1.0.2: Use your platform's native performance.now() and performance.timeOrigin.
npm WARN deprecated uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
npm WARN deprecated request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142
```

I thought about giving PurgeCSS a go, since it has been more [actively maintaned](https://github.com/FullHuman/purgecss), but [jekyll-purgecss](https://github.com/mhanberg/jekyll-purgecss) haven't.

For this, I needed to change to use some local libraries instead of getting them via CDN. The good news is that it is quite effective in reducing css file sizes. Comparing dir sizes with `du -hs _site/assets/css/`:

| current | minify | PurgeCSS | PurgeCSS + minify |
| ------- | ------ | -------- | ----------------- |
| 1,1M    | 988K   | 456K     | 420K              |


pourmand1376:
It would be a lot better if we could see the online version of the site after the pull request without building it locally. Is it possible for Jekyll? @rohandebsarkar

george-gca:
For this specific case I can use these commits in [my demo site](https://george-gca.github.io/multi-language-al-folio/) and we can check if everything is ok there. What do you think?

Another solution is to create an install script to install everything needed to run locally outside of docker, so you could test it. But I agree, see an online version after each PR (or at least select the PRs to which generate a temporary demo from) would be better.

george-gca:
Just tried out the changes in my personal site and run a few times.
Before:

| | Desktop | Mobile |
| --- | --- | --- |
| Before | Screenshot from 2023-07-30 19-30-34 | Screenshot from 2023-07-30 19-30-40 |
| After | Screenshot from 2023-07-30 19-39-05 | Screenshot from 2023-07-30 19-39-10 |

george-gca:
Tried to add `bootstrap-table` to use local files, but for some reason it messes with it and I can't avoid it, even after adding it to ignore files in purgecss. Best keep it loading from web.

Before purge:

image

After purge:

image

alshedivat (approved):
just got a chance to read through thes changes. thanks for adding PurgeCSS and switching away from CDNs. looks good to me, I'm ok with merging it.

the only concern I have is that updating versions of the libraries will be more involved (previously, we could update fontawesome and other libraries by changing a couple of lines in `_config.yml`), but I guess better speed and more flexibility are worth this inconvenience.

PR #1629, opened 2023-08-03
Fix regex to limit keyword scope to fields

gzwongkk:
This fix in https://github.com/alshedivat/al-folio/pull/1429 is not perfect.
`input = input.gsub(/^.*#{keyword}.*$\n/, '')` will fail but return true for cases like looking for "code" from "coder"

The following is the real fix
`input = input.gsub(/^.*\b#{keyword}\b *= *\{.*$\n/, '')`

The `\b` in the regex expression is a word boundary. This means it will not match substrings in other words.
The ` *= *\{` will ensure that the keyword is a field name because bib should follows `<field name> = {<value>}`.


alshedivat:
thanks for contributing an improved fix!

Issue #1724, opened 2023-09-13
Add WeChat to socials


johanneshoerner:
**Is your feature request related to a problem? Please describe.**
There is no clean possibility to add a WeChat account to socials. 

**Describe the solution you'd like**
A button in socials that shows a WeChat QR code when clicked or hovered.

**Describe alternatives you've considered**
Directly link to WeChat account when clicking the button. But as far as I know that is not possible at the moment, because WeChat doesn't allow for that. 

**Additional context**
As WeChat is almost exclusively mobile based, I think that showing the QR code is the best solution. Otherwise it would also be possible to just show the WeChat ID.

johanneshoerner:
The QR code is now displayed in the center of the screen when the button is clicked. I think a hover image would look nicer, but this was the simplest way for me to implement it.

PR #1730, opened 2023-09-14
add WeChat to socials

johanneshoerner:
When the WeChat icon in socials is clicked, a picture of your WeChat QR code is shown in the center of the screen. The actual picture of your QR code has to be uploaded.  Related to #1724


alshedivat:
thanks for adding this feature! I've slightly adjusted your code and put html, css, and js in appropriate places.

Issue #1609, opened 2023-07-26
Mermaid Diagrams in Dark Mode


DavidKrassnig:
I noticed that mermaid diagrams don't render correctly in dark mode (or, at least, not  really legibly). The dark colours should be switched to bright colours to contrast against the dark background..

**Current look:**
![mermaidiagram_dark_alfolio](https://github.com/alshedivat/al-folio/assets/96623786/f3831b34-8d4e-40b0-95cf-b0d028940e6c)

**Very same mermaid diagram rendered by, e.g., jekyll-theme-chirpy, for comparison:**
![mermaidiagram_dark_chirpy](https://github.com/alshedivat/al-folio/assets/96623786/3792b9a6-697d-4f9a-a8e1-1b9c4e8dc455)

I assume that the style sheets need to be modified accordingly. Unfortunately, not my field of expertise at all.

george-gca:
Just verified and [mermaid support dark themes](https://mermaid.js.org/config/theming.html#available-themes). It is a question of changing this when changing to dark mode, probably the same way it is done with other elements in [assets/js/theme.js](https://github.com/alshedivat/al-folio/blob/master/assets/js/theme.js). If no one tackles this, maybe I can take a look, but not until next week.

george-gca:
I might be wrong about this, but as I understood this can't be done so easily in our case.

We use [jekyll-diagrams](https://github.com/zhustec/jekyll-diagrams) plugin which takes care of rendering the diagram as a svg. It generates the svg "offline" and embed it directly into the html. As a result, it doesn't make any include to the `mermaid.js`.

If you check the template you sent, it does include `mermaid.min.js`, so it can dinamically render the svg and change its properties. So it is possible in that template to change mermaid theme and tell it to re-render the svg.

What you guys think @alshedivat @pourmand1376? The best solution I could think here would be to drop this plugin and use mermaid as an include so we could dinamically change this.

pourmand1376:
I agree too. It is better to have an online version so that we can dynamically generate mermaid diagrams on-the-flow. 

How much is the size of this `mermaid.mis.js`?

Also, we should load it only if we have a mermaid diagram on that specific page. This could be tagged by some property like `has_mermaid` in page properties.

george-gca:
Apparently [mermaid.min.js](https://cdn.jsdelivr.net/npm/mermaid@10.3.0/dist/mermaid.min.js) is 2.78 MB.

[...]

PR #1786, opened 2023-10-02
Update Base Docker Image

pourmand1376:


alshedivat (approved):

george-gca:
Question, why change this for the 23.10? The `latest` tag is automatically set for the latest LTS release, which currently is 22.04. The next LTS release will be 24.04 I believe.

pourmand1376:
You're right. I thought this would help to solve the windows docker issue. But it didn't.

george-gca:
> You're right. I thought this would help to solve the windows docker issue. But it didn't.

Have you added the mermaid dependency? But nevermind, today I will try to work on #1609 that maybe will also fix this, since most errors with Windows are related to the post with mermaid diagrams.

alshedivat:
should we close this one?

pourmand1376:
Yeah.

PR #1792, opened 2023-10-04
Update giscus setup via github action

alshedivat:
giscus settings in `_config.yml` had al-folio repo specified by default. many users kept these defaults in their own repos, which resulted in getting comments from blog posts of different users posted to al-folio discussions. this is undesirable, since users lose control over the discussions in their blogs.

this PR solves the issue:
- first, we set `giscus.repo` to blank in `_config.yml`. if the field is kept blank, when the website is built locally, the user will see a warning saying that giscus comments are misconfigured.
- second, we add a step to the `deploy` workflow that writes repository name to `giscus.repo` in `_config.yml`. that way, even if `giscus.repo` field is left black or set to an incorrect repo, it gets correctly populated at deployment time.

other small changes in this PR are small stylistic adjustments.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -26,6 +26,13 @@ jobs:
       with:
         ruby-version: '3.2.2'
         bundler-cache: true
+    - name: Update _config.yml ⚙️
+      uses: fjogeleit/yaml-update-action@v0.13.1
+      with:
+        commitChange: false
+        valueFile: '_config.yml'
+        propertyPath: 'giscus.repo'
+        value: ${{ github.repository }}
     - name: Install and Build 🔧
       run: |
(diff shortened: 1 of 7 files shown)


andrewmzhang:
This decision gave me some grief. I was trying to keep my al-folio repo private and have giscus comments in a separate public repo and I was not expecting the repo I put in under `giscus.repo` to be overwritten by GHA. I feel like this behavior should be noted somewhere, or only use the current repo if the user leaves `giscus.repo` blank. 

Let me know if I put this comment in the wrong place.

george-gca:
@andrewmzhang I agree with you about only updating if the variable is empty. Would you mind sending a PR for this?

andrewmzhang:
@george-gca I'll take a look at this tonight/tomorrow and get a PR in

Issue #164, opened 2021-01-02
Footer with references disappearing in distill blog posts when removing author and afiiliation bar


LoryPack:
**Acknowledge the following**
- [x] I carefully read and followed the [Getting Started](https://github.com/alshedivat/al-folio#getting-started) guide.
- [x] I read through [FAQ](https://github.com/alshedivat/al-folio#faq) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [x] The issue I am raising is a potential bug in al-folio and not just a usage question. <br> [For usage questions, please use [gitter chat](https://gitter.im/alshedivat/al-folio) instead of raising an issue.]

**Describe the bug**
Hi, thanks so much for providing such a beautiful theme. I am having some troubles with the distill post format. I want to write a blog post with that (rather than with the "post" layout) in order to exploit the bibliographic capabilities. However, I would like to remove the authors bar on the top, which appears immediately below the title (see the example https://alshedivat.github.io/al-folio/blog/2018/distill/). I am able to remove that in two ways: 

1.  by removing the line
https://github.com/alshedivat/al-folio/blob/4f9cdb131e01fa2febb075df518faaa41465b633/_layouts/distill.html#L66

2. by removing
https://github.com/alshedivat/al-folio/blob/4f9cdb131e01fa2febb075df518faaa41465b633/_layouts/distill.html#L17-L33

Both of them however have the side effect of removing also the footer were references and footnotes are supposed to appear. 

**To Reproduce**
Update the `distill.html` file as described above. I have done it in my fork [here](https://github.com/LoryPack/lorypack.github.io/blob/bd681d67fd4ec27c7152552162067231e5938802/_layouts/distill_no_author_1.html) and [here](https://github.com/LoryPack/lorypack.github.io/blob/bd681d67fd4ec27c7152552162067231e5938802/_layouts/distill_no_author_2.html) for the two above options, which yield the webpages which you can see [here](http://www.lorenzopacchiardi.me/blog/2018/distill/) and [here](http://www.lorenzopacchiardi.me/blog/2018/distill_2/).

**Expected behavior**
I expected the author and affiliation bar to disappear, but the footer with references to remain there.

**System (please complete the following information):**
 - OS: Ubuntu 20.04
 - Browser: Firefox 84.0, Chrome  81.0.4044.122 
 - Jekyll version: 3.9.0
[...]

PR #1804, opened 2023-10-07
fix reference disappearing for distill blog in absence of author

YifanJiang233:
This a quick fix of issue #164. The by-line will be removed if the author of the post is not given. Please find _posts/2023-10-07-distill-no-author.md for a minimal example.


alshedivat (approved):

alshedivat:
thank you for the fix! 🙏

I appreciate it that you added `_posts/2023-10-07-distill-no-author.md` to demonstrate that the fix works. I don't think we want to keep this additional distill example though, so I'm going to remove it and go ahead and merge.

PR #1821, opened 2023-10-12
Update resume.json

pablodiegoo:
populating resume.json to give easier start to people that dont understand very well how to setup the fields


george-gca (approved):

george-gca:
Thanks for updating this.

Issue #1499, opened 2023-06-22
Fix weird alignment between date and title in latest posts


george-gca:
**Acknowledge the following**
- [x] I carefully read and followed the [Getting Started](https://github.com/alshedivat/al-folio#getting-started) guide.
- [x] I read through [FAQ](https://github.com/alshedivat/al-folio#faq) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [x] The issue I am raising is a potential bug in al-folio and not just a usage question. <br> [For usage questions, please post in the [Discussions](https://github.com/alshedivat/al-folio/discussions) instead of raising an issue.]

**Describe the bug**
Fix the weird alignment between date and post title in `latest posts` section in about page. It should have the same aligment as the `news` section.

**To Reproduce**
https://alshedivat.github.io/al-folio/

**Expected behavior**
Same alignment as the `news` section.

**Screenshots**
![image](https://github.com/alshedivat/al-folio/assets/31376482/6b9c3ed2-8d3a-43b8-800f-01c76904745d)

VanniPrashar:
Can I fix this please

PR #1832, opened 2023-10-14
fix inconsistent alignment for latest posts

YifanJiang233:
This is a quick fix of issue #1499. The widths of the date and the title are by default dynamically determined by their lengths. I manually set the width of date as 20% of the table's width, so the alignment can be consistent.


alshedivat:
thanks for the fix!

PR #1869, opened 2023-10-23
check for active children in submenus

Tianmaru:
Apparently, submenus were not highlighted, even if the current page was a child of that submenu.
This patch checks, if the title of the current page matches with the title of a submenu child. If so, it sets the submenu and the corresponding child as active, highlighting it in the header.

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -67,9 +67,15 @@
               {%- for p in sorted_pages -%}
               {%- if p.nav and p.autogen == nil -%}
               {%- if p.dropdown %}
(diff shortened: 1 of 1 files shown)


george-gca:
What is the use case exactly? Just so I can test your solution.

Tianmaru:
Testing should be possible in any use case, where navigation bar items have drop-down sub-menus :)

alshedivat:
thanks for this contribution. tested visually and seems to work as expected. merging.

PR #1892, opened 2023-11-01
bring back `repo_id` and `category_id` in `giscus` section

alerezaaa:
After deploying my website (Cloudflare Pages), my Giscus did not work, I realized these two paramesters are removed. After adding them again, my Giscus worked correctly.
These new values are values from previous commits of original repository

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -138,9 +138,9 @@ related_blog_posts:
 # the information below.
 giscus:
   repo:                          # <your-github-user-name>/<your-github-repo-name>
-  repo_id: MDEwOlJlcG9zaXRvcnk2MDAyNDM2NQ==
+  repo_id:                       # leave empty or specify your repo_id (see https://giscus.app/) 
   category: Comments             # name of the category under which discussions will be created
-  category_id: DIC_kwDOA5PmLc4CTBt6
+  category_id:                   # leave empty or specify your category_id (see https://giscus.app/) 
   mapping: title                 # identify discussions by post title
   strict: 1                      # use strict identification mode
   reactions_enabled: 1           # enable (1) or disable (0) emoji reactions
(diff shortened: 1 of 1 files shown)


alshedivat:
`repo_id` and `category_id ` that you added back are specific to al-folio repository and should not be used any other repos. instead, you should follow instructions on https://giscus.app/ to generate config that is specific to your repo. previously, people would keep using al-folio giscus configuration, and all their comments would end up getting posted to discussions in this repo instead of their own repos. that's why I removed configuration specific to this repo.

giscus seems to work fine without these two attributes on [the demo website](https://alshedivat.github.io/al-folio/blog/2022/giscus-comments/). I'm not sure why it doesn't work for you without `repo_id` and `category_id `, and I'm surprised it starts working once you add back fields, which are incorrect for your repo.

alerezaaa:
> `repo_id` and `category_id ` that you added back are specific to al-folio repository and should not be used any other repos. instead, you should follow instructions on https://giscus.app/ to generate config that is specific to your repo. previously, people would keep using al-folio giscus configuration, and all their comments would end up getting posted to discussions in this repo instead of their own repos. that's why I removed configuration specific to this repo.
> 
> giscus seems to work fine without these two attributes on [the demo website](https://alshedivat.github.io/al-folio/blog/2022/giscus-comments/). I'm not sure why it doesn't work for you without `repo_id` and `category_id`, and I'm surprised it starts working once you add back fields, which are incorrect for your repo.

I use Cloudflare Pages and it does not work for me without these attributes. btw, Build process is the same (Github Action) and CF-Pages only reads built files from proper branch 🤔

So, maybe its better to add empty feilds for `repo_id` and `category_id`, then delete these attributes during build time (IF they are empty), like what you did for changing `repo` attribute during build process.

**I did not add `al-folio` values for my website**, I used proper `repo_id` and `category_id ` for my website using giscus guides

george-gca (approved):

PR #1909, opened 2023-11-09
added ability to switch off pagination

AbstractGeek:
Just added a few lines of code so that the blog works with pagination off. It displays all posts - a zoomed out image is shown below.

Cheers!

image

### DIFF
--- a/blog/index.html
+++ b/blog/index.html
@@ -95,7 +95,14 @@ pagination:
   {% endif %}
 
   <ul class="post-list">
-    {% for post in paginator.posts %}
+
+    {%- if page.pagination.enabled -%}
+      {%- assign postlist = paginator.posts -%}
+    {%- else -%}
+      {%- assign postlist = site.posts -%}
+    {%- endif -%}
+
+    {% for post in postlist %}
 
     {% if post.external_source == blank %}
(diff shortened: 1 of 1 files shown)


blog/index.html
ghltshubh, line 1: @AbstractGeek Is this feature documented in the documentation?

george-gca (approved):

blog/index.html
george-gca: What do you mean? You think this information should be added somewhere? If so, where?

ghltshubh: I mean will this be added to the README which kind of serves as the documentation.

george-gca: I think it is a good idea. @AbstractGeek can you add this info to the README?

AbstractGeek: Hmm, I am not sure what needs to be added? It is just switching on or off pagination. Not really a big feature that needs documentation 😅

alshedivat: the blog section in `_config.yml` had `pagination` field, which makes this somewhat self-explanatory. I think it's ok to not add anything to readme unless we see people start getting confused / ask questions.

alshedivat (approved):
thanks for adding this feature, LGTM!

PR #1992, opened 2023-12-13
Changed mermaid support in template

### LINKED ISSUE #1609: Mermaid Diagrams in Dark Mode

george-gca:
Fix #1609


george-gca:
I decided to implement a mix of @reinaldoacdc solution with the one in [jekyll-theme-chirpy](https://github.com/cotes2020/jekyll-theme-chirpy). While the first allow zooming in the diagram, the other solution allow creating a diagram by using plain markdown. For example, check their [post](https://chirpy.cotes.page/posts/text-and-typography/) and its [code](https://github.com/cotes2020/jekyll-theme-chirpy/blob/master/_posts/2019-08-08-text-and-typography.md).

The diagram is working fine with dark theme support, but there is only one problem: if zooming in diagram is enabled, it only works before you change the theme. If you switch between light/dark mode, the zoom stops working and idk why.

@reinaldoacdc you seem to understand this more than me, do you know how can I fix the d3js zoom after changing the theme?

alshedivat:
looked into this for a bit: if you run the following code block in the console, the zoom starts working again:
```javascript
var svgs = d3.selectAll(".mermaid svg");
svgs.each(function () {
  var svg = d3.select(this);
  svg.html("<g>" + svg.html() + "</g>");
  var inner = svg.select("g");
  var zoom = d3.zoom().on("zoom", function (event) {
    inner.attr("transform", event.transform);
   });
   svg.call(zoom);
   console.log(svg);
});
```
since the SVG are being regenerated on theme toggle, the on zoom even handler should be reattached to the newly generated SVG, correct? is it being attached somewhere in `setMermaidTheme`?

george-gca:
> since the SVGs are being regenerated on theme toggle, the on zoom event handler should be reattached to the newly generated SVGs, correct? is zoom handler being attached somewhere in `setMermaidTheme`?

Correct. It supposedly is, but it is not working. See this part in `theme.js`:

```javascript
const observable = document.querySelector(".mermaid svg");
if (observable !== null) {
  var observer = new MutationObserver(addMermaidZoom);
  const observerOptions = {
    // attributes: true,
    childList: true,
    // subtree: true,
  };
  observer.observe(observable, observerOptions);
}
```

As I understood from the codes I based mine on, we need to re-render the svg and only add the zoom when it is done rendering, so that is why we need this `MutationObserver`. When it finishes updating the `childList` of the `observable` elements observed by the `observer`, it calls `addMermaidZoom` function, which is exactly the code that you pasted.

While debugging you can see that it correctly adds the `<g>` element that will contain the whole svg inside, but not the zoom function.

alshedivat:
@george-gca, what do you think of just merging the current version and creating an issue to track the disappearance of zoom function when toggling light/dark themes? I feel like the current version is already a significant improvement.

george-gca:
I agree. And in most cases the user will not be changing the theme while viewing a diagram. I will merge this.

Issue #1795, opened 2023-10-05
Hide "blog" from the navigation bar and adjust navigation order


xiong-zx:
I would like to hide "bolg" in the navigation bar. #1621 suggests setting
 
> blog_nav_title: blog

in `_config.yml`

However, if I want to keep the blog in the navigation bar yet adjust its order, how to do this?

Related issues:
#1559 #1189 #1196

george-gca:
To adjust its order, currently you have to manually change it in the [_includes/header.html](https://github.com/alshedivat/al-folio/blob/a21bd492bc9950f86f160baea1a2e31d3cc6227d/_includes/header.html#L54). As you can see, I opened #1559 as a reminder to implement an easier way.

PR #1996, opened 2023-12-14
Now blog can be reorganized in header

### LINKED ISSUE #1559: Enable change the blog position in header
Related questions: #1189 #1196

There should be an easy way to change blog position in header, including adding it to drop-down menu.

george-gca:
Fixes #1559 #1795


alshedivat (approved):
LGTM, thanks for the fix!

PR #2000, opened 2023-12-17
Fix navigation on the header

GihanAyesh:
When on page other than about; the title is shown as a clickable object. But when clicked nothing happens. Fixed this issue by redirecting to the main page.

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -5,7 +5,7 @@
       <nav id="navbar" class="navbar navbar-light navbar-expand-sm {% if site.navbar_fixed %}fixed-top{% else %}sticky-top{% endif %}">
         <div class="container">
           {% if page.permalink != '/' -%}
-          <a class="navbar-brand title font-weight-lighter" href="{{ site.baseurl }}/">
+          <a class="navbar-brand title font-weight-lighter" href="{{ site.url }}/">
             {%- if site.title == "blank" -%}
               {%- if site.first_name -%}
                 <span class="font-weight-bold">{{- site.first_name -}}&nbsp;</span>


george-gca:
Are you sure this is happening? Everything seems fine on the [template demo site](https://alshedivat.github.io/al-folio/).

GihanAyesh:
Yes this happened to me and two more friends. 

Reproduce steps: Deploy the page without docker (Didn't check with docker deployments don't think that's an issue)

george-gca:
I don't know if this is actually the problem. In the current solution, it generates `href="/al-folio/"`, which in turn creates the link `127.0.0.1:4000/al-folio/` that can be seen in the bottom left here:

image

With your solution, it generates the whole url, which would be `href="http://localhost:4000/"`, creating the link `localhost:4000` as can be seen in the bottom left:

image

Which leads to the wrong url and consequently to the `Page not found`.

_includes/header.html
```diff
@@ -5,7 +5,7 @@
       <nav id="navbar" class="navbar navbar-light navbar-expand-sm {% if site.navbar_fixed %}fixed-top{% else %}sticky-top{% endif %}">
         <div class="container">
           {% if page.permalink != '/' -%}
-          <a class="navbar-brand title font-weight-lighter" href="{{ site.baseurl }}/">
+          <a class="navbar-brand title font-weight-lighter" href="{{ site.url }}/">
```
alshedivat, line 8:
```suggestion
          <a class="navbar-brand title font-weight-lighter" href="{{ site.baseurl | prepend: site.url }}/">
```

alshedivat (changes requested):
please prepend `url` instead of substituting `baseurl` with it and test if it works. happy to merge if that solves your problem.

alshedivat:
closing this since it will require a rebase and the change is just a one-liner. please rebase and reopen if still an issue.

Issue #1882, opened 2023-10-27
Dates in CSV too near and too low with respect to left text


raffaem:
The dates in the CV are too near the text on the left (i.e., the margin is too low), and moreover they are too low with respect to the text on the left:

![image](https://github.com/alshedivat/al-folio/assets/54762742/b080d5ff-9fcb-42aa-861f-541a1fa1db63)

vedanthanekar45:
How can I work on it?

george-gca:
You will basically need to play with the css of the page.

vedanthanekar45:
Oh, okay. This is actually my first time doing open source, so... Thank you

PR #2003, opened 2023-12-18
Fixed: Dates in CSV too near and too low with respect to left text #1882

icarus:
Added _'white-space: normal;'_ to the style of the span that contains _{{ content.year }}_ on __includes/cv/time_table.html._ 

This let's the text break without being too near to the text on the right, making it able to handle any year range without looking too cluttered. 

Screenshot 2023-12-18 at 14 51 39

### DIFF
--- a/_data/cv.yml
+++ b/_data/cv.yml
@@ -72,18 +72,18 @@
   type: time_table
   contents:
     - year: 1921
-      items: 
-        - Nobel Prize in Physics 
+      items:
+        - Nobel Prize in Physics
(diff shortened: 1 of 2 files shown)


george-gca:
This also gave inconsistent layout for the template example. Look:

image

icarus:
Would you like me to find a way to keep regular year ranges (i.e. 1907 - 1910) in the same line, and breaking the line once it has months or similar (i.e. 1905.31 - 1930.13)?

PR #2006, opened 2023-12-21
Added anchor to category sections

fabawi:
Added anchor to categories so one could navigate to the category section from the URL:

e.g.,

https://alshedivat.github.io/al-folio/projects#fun


george-gca (commented):
Based on this, I believe we should also have links for them.

alshedivat:
@fabawi, I agree with @george-gca. could you add a `href` attribute with a link to the category page? I'll go ahead and merge after that.

fabawi:
@george-gca @alshedivat I added the header to the anchor link. Doing it the other way around (link inside header) is cleaner, but changes the look of the header, so I went with the former. Now when you click on the category's header it navigates to the page anchor

george-gca:
@fabawi can you update this PR so we can merge it?

fabawi:
@george-gca updated

PR #2014, opened 2023-12-24
Responsive image fix

### LINKED ISSUE #1967: Webp images are not correctly being used and can result in long load times
**Acknowledge the following**
- [x] The issue I am raising is a potential bug in al-folio and not just a usage question. <br> [For usage questions, please post in the [Discussions](https://github.com/alshedivat/al-folio/discussions) instead of raising an issue.]

**Describe the bug**
I have built a website using this template (which is wonderful by the way!) but the responsive images are not working the way I think they are intended to. Take a look at this page: [https://bencbartlett.com/publications/](https://bencbartlett.com/publications/). I have webp images for each publication thumbnail ([example](https://bencbartlett.com/assets/img/publication_preview/in_situ_diagram-150.webp)) but they are not correctly getting loaded by the browser and instead the full-size png images are loading. Try loading the linked publication page, open the inspection tools (cmd+opt+I) and go to the network tab and sort by waterfall plot, then do a hard refresh (cmd+shift+r) and you can see that none of the small webp images are loading. The viewports have a width of about 100px so the `-150.webp` images should load first.

With `site.imagemagick` enabled, in `figure.html` there is this snippet which I suspect is where the problem is but I have pretty limited web development experience so I'm not sure.

```html
...
  <picture>
    {% if site.imagemagick.enabled %}
    {% for i in site.imagemagick.widths -%}
      <source
        class="responsive-img-srcset"
        media="(max-width: {{ i }}px)"     
        srcset="{{ img_path | relative_url }}-{{ i }}.webp"
      />
    {% endfor -%}
    {% endif %}

    <!-- Fallback to the original file -->
    ....
  </picture>

  {%- if include.caption -%}<figcaption class="caption">{{ include.caption }}</figcaption>{%- endif %}

</figure>
```

**To Reproduce**
1. Go to [https://bencbartlett.com/publications/](https://bencbartlett.com/publications/)
2. Open the inspection tools (cmd+opt+I) and go to the network tab and sort by waterfall plot, then do a hard refresh (cmd+shift+r)
3. Observe that [https://bencbartlett.com/assets/img/publication_preview/insitu_backprop-480.webp](https://bencbartlett.com/assets/img/publication_preview/insitu_backprop-480.webp) exists and should be a valid source tag.
4. Look in the inspector and see that the small webp does not load first, and instead the full png which is several MB loads: [https://bencbartlett.com/assets/img/publication_preview/insitu_backprop.png](https://bencbartlett.com/assets/img/publication_preview/insitu_backprop.png) 
5. This drastically increases the loading time of the page and network footprint.

**Expected behavior**
When you do the above, one of the small webp images should load instead of the png.

**Screenshots**

image

 - OS: MacOS 14.0
 - Browser (and its version) Chrome Version 119.0.6045.159 (Official Build) (arm64)
 - Jekyll version [e.g. 3.8.7]
- Ruby version [e.g. 2.6.5]

**Additional context**

nWestie:
Fixes #1967 

Update the way responsive images are used to only use one srcset, and add corresponding `sizes` attribute to the relevant layouts and templates. 
I did not go through and add `sizes` to all the example posts/projects. When `sizes` is not specified by the user, the `figure.html` template defaults to 95% the width of the viewport which should work fine for most cases; users can optimize further if they wish by feeding `sizes` into the template.

Additionally:
 - Enabled support for .gif to .webp compression
 - fix error in jekyll-imagemagick config where all images were resized to be 800px or less. (for example img-1400.webp was actually only 800px wide, etc.)
 - added note about making sure imagemagick is installed before enabling it in responsive images section of `_config.yml`

### DIFF
--- a/_includes/figure.html
+++ b/_includes/figure.html
@@ -3,6 +3,8 @@
 <figure>
   <picture>
     <!-- Auto scaling with imagemagick -->
+    <!-- See https://www.debugbear.com/blog/responsive-images#w-descriptors-and-the-sizes-attribute and 
+      https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images for info on defining 'sizes' for responsive images-->
     {% if site.imagemagick.enabled %}
       <source class = "responsive-img-srcset" srcset="
       {% for i in site.imagemagick.widths -%}
@@ -14,7 +16,7 @@
     <img
       src="{% if include.cache_bust %}{{ include.path | relative_url | bust_file_cache }}{% else %}{{ include.path | relative_url }}{% endif %}"
       {% if include.class %}class="{{ include.class }}"{% endif %}
-      {% if include.width %}width="{{ include.width }}"{% else %}width="auto"{% endif %}
+      {% if include.width %}width="{{ include.width }}"{% else %}width="100%"{% endif %}
       {% if include.height %}height="{{ include.height }}"{% else %}height="auto"{% endif %}
       {% if include.min-width %}min-width="{{ include.min-width }}"{% endif %}
       {% if include.min-height %}min-height="{{ include.min-height }}"{% endif %}
--- a/_includes/figure.html
+++ b/_includes/figure.html
@@ -1,19 +1,16 @@
-{%- assign img_path = include.path | remove: ".jpg" | remove: ".jpeg" | remove: ".png" | remove: ".tiff" -%}
+{%- assign img_path = include.path | remove: ".jpg" | remove: ".jpeg" | remove: ".png" | remove: ".tiff" | remove: ".gif" -%}
 
 <figure>
-
   <picture>
+    <!-- Auto scaling with imagemagick -->
     {% if site.imagemagick.enabled %}
-    {% for i in site.imagemagick.widths -%}
-      <source
-        class="responsive-img-srcset"
-        media="(max-width: {{ i }}px)"
-        srcset="{{ img_path | relative_url }}-{{ i }}.webp"
-      />
-    {% endfor -%}
+      <source class = "responsive-img-srcset" srcset="
+      {% for i in site.imagemagick.widths -%}
+      {{ img_path | relative_url }}-{{ i }}.webp {{i}}w,
+      {% endfor -%}" 
+      {% if include.sizes %} sizes="{{include.sizes}}" {% else %} sizes="95vw" {% endif %}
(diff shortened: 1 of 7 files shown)


_layouts/about.html
```diff
@@ -26,6 +26,7 @@ <h1 class="post-title">
               {% include figure.html
               path=profile_image_path
               class=profile_image_class
+              sizes = "(min-width: 800px) 231, (min-width: 576px) 30vw, 95vw"
```
alshedivat, line 29: should this be `230px`?

_layouts/profiles.html
```diff
@@ -22,6 +22,7 @@
               {% include figure.html
               path=profile_image_path
               class=profile_image_class
+              sizes = "(min-width: 800px) 231, (min-width: 576px) 30vw, 95vw"
```
alshedivat, line 25: same here: should this be `231px`?

alshedivat (approved):
thanks for the fix! LGTM overall, just a couple of quick questions.

_includes/figure.html
```diff
@@ -1,23 +1,22 @@
...
+      {% if include.sizes %} sizes="{{include.sizes}}" {% else %} sizes="95vw" {% endif %}
+      type="image/webp"/>
     {% endif %}
-
-    <!-- Fallback to the original file -->
     <img
       src="{% if include.cache_bust %}{{ include.path | relative_url | bust_file_cache }}{% else %}{{ include.path | relative_url }}{% endif %}"
       {% if include.class %}class="{{ include.class }}"{% endif %}
-      {% if include.width %}width="{{ include.width }}"{% else %}width="auto"{% endif %}
+      {% if include.width %}width="{{ include.width }}"{% else %}width="100%"{% endif %}
```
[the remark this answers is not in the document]
nWestie: Leaving width as auto led to weird scaling behavior, mainly visible on the example projects page, where the .img-fluid sets max-width of 100%, not width of 100%. Changing the default to 100% makes images predictably take up the full width available to them. Without it, the width of the image behaves oddly and is not predictable as it will be different depending which size image is loaded. See examples of odd widths below.

image
image

_layouts/about.html
nWestie: yes, it should be 231px thanks. But I'm remembering that the 800px max width is easily variable in the config file... so I should probably use liquid to base it on that instead, I'll go do that.

PR #2022, opened 2023-12-28
fix live-reloading in docker

### LINKED ISSUE #2021: Live-Reloading does not work when using Docker
**Acknowledge the following**
- [x] The issue I am raising is a potential bug in al-folio and not just a usage question. <br> [For usage questions, please post in the [Discussions](https://github.com/alshedivat/al-folio/discussions) instead of raising an issue.]

**Describe the bug**
The webpage hosted by docker would not be automatically reloaded, because the port used for live-reloading (port 35729) is not forwarded.

**To Reproduce**
1. Run a localhost server using `docker compose up`
2. Open the webpage `http://localhost:8080` in a browser.
3. Edit a file (e.g. `_pages/about.md`) and save it.
4. The browser does not auto-reload.

**Expected behavior**
The webpage should auto-reload once a file change is saved.

**Note**
I have founded the solution & I am submitting a PR to fix this.

wayne-tsai:
Fix #2021 : The webpage hosted by docker would not be automatically reloaded because the port used for live-reloading (port 35729) is not forwarded.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,5 +6,6 @@ services:
     build: .
     ports:
       - 8080:8080
+      - 35729:35729
     volumes:
       - .:/srv/jekyll


pourmand1376:
Brilliant. Thanks for the fix.

PR #2028, opened 2023-12-30
Remove imagemagick strip

nWestie:
I had added -strip arg to the imagemagick call as it is generally desirable to strip metadata from images before publishing them, however it seems to occasionally cause issues with jekyll-imagemagick auto-regeneration, regenerating images even when they have not changed.


alshedivat:
thanks! merging

PR #2048, opened 2024-01-06
format code with prettier.io

alshedivat:
summary:
- adds prettier formatter configuration
- formats the entire repo using prettier, ignoring minified files (`*.min.css`) and  heavy generated html
- changes extensions of all `.html` files to `.liquid`, which is more correct and necessary for prettier to work correctly
- replaces "%-" and "-%" with just "%" — manual liquid formatting using minus signs is superfluous since we are compressing and minifying the code anyway
- adds CI action for running prettier check on PR and pushes to master


alshedivat:
@george-gca, just a heads up that I'm going to merge this PR to improve code formatting throughout the repo.

Issue #2066, opened 2024-01-12
Custom blockquotes stopped working


george-gca:
**Acknowledge the following**

- [x] I carefully read and followed the [Getting Started](../../README.md#getting-started) guide.
- [x] I read through [FAQ](../../FAQ.md) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [x] The issue I am raising is a potential bug in al-folio and not just a usage question. <br> [For usage questions, please post in the [Discussions](https://github.com/alshedivat/al-folio/discussions) instead of raising an issue.]

**Describe the bug**
Custom blockquotes stopped working in the template.

**To Reproduce**
Simply navigate to [related blog post](https://alshedivat.github.io/al-folio/blog/2023/custom-blockquotes/)

PR #2071, opened 2024-01-13
Fix custom blockquotes

YifanJiang233:
Fixed issue #2066. Removed `>` in front of `{: .block-tip }` which was introduced by prettier in the example post. Maybe add /_post/*.md into .prettierignore to prevent similar breaks in the future.


alshedivat:
thanks for the fix. I added prettier ignore statements to avoid incorrect formatting.

PR #2077, opened 2024-01-15
Fix broken links

george-gca:
Added GitHub action to check for broken links in repo before commit, ignoring files with liquid tags since these were not yet processed and also README since some sites created from this template doesn't have valid urls anymore. Fixed some broken links in repo.


alshedivat (approved):
LGTM, thanks for adding the workflow!

PR #2078, opened 2024-01-15
Local posts with the first md file writen by own

Thesoul20:


Thesoul20:
Although we can receive the master branch, but the git push flow is complex now for the usage of a new local branch names as local_used

PR #2082, opened 2024-01-15
Changed conditions of running GitHub actions

george-gca:
The GitHub actions were running everytime a new commit was made to the repo, but that wasn't needed. For example, we don't need to try to create a new docker image if the libraries didn't change, same to build the whole site after a change in the README.md.

### DIFF
--- a/.github/workflows/deploy-docker-tag.yml
+++ b/.github/workflows/deploy-docker-tag.yml
@@ -4,6 +4,8 @@ on:
   push:
     tags:
       - "v*"
+    paths:
+      - "Dockerfile"
 
 jobs:
   build:
--- a/.github/workflows/deploy-docker-tag.yml
+++ b/.github/workflows/deploy-docker-tag.yml
@@ -5,7 +5,12 @@ on:
     tags:
       - "v*"
     paths:
+      - "bin/entry_point.sh"
       - "Dockerfile"
+      - "Gemfile"
+      - "Gemfile.lock"
+      - "package.json"
+      - "package-lock.json"
 
 jobs:
   build:
--- a/.github/workflows/deploy-docker-tag.yml
+++ b/.github/workflows/deploy-docker-tag.yml
@@ -18,28 +18,28 @@ jobs:
 
     steps:
       - name: Checkout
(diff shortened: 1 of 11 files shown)


pourmand1376 (approved):

pourmand1376:
Awesome!

.github/workflows/deploy-docker-tag.yml
```diff
@@ -4,6 +4,8 @@ on:
   push:
     tags:
       - "v*"
+    paths:
```
pourmand1376, line 7: What if dependencies in the Gemfile change?

pourmand1376: Also, I think `package.json` and `package-lock.json` and `bin/entry_point.sh` should be included. I have to see if any other file exist for this CI/CD.

george-gca: Nice, that is exactly why I added you as a reviewer since you are our docker master haha

george-gca: If you find any other additions (or don't) let me know so I can merge these on master

.github/workflows/docker-slim.yml
```diff
@@ -14,30 +14,30 @@ on:
...
 
       - name: update docker-compose
         shell: bash
         run: |
-          sed -i "s|\.:|${{ github.workspace }}:|g" ${{ github.workspace }}/docker-compose.yml 
+          sed -i "s|\.:|${{ github.workspace }}:|g" ${{ github.workspace }}/docker-compose.yml
           cat ${{ github.workspace }}/docker-compose.yml
 
-      - uses: kitabisa/docker-slim-action@v1.0.3
+      - uses: kitabisa/docker-slim-action@v1
```
pourmand1376, line 40: Why this version is being downgraded? 

I have discussed this with the developer and version 1 would definitely cause problems.

george-gca: Actually I did this because I assumed that if we only setup the major version, it would automatically use the latest minor version, but I couldn't find anything that backs my assumption. Reverting this change.

PR #2085, opened 2024-01-16
Prettier check now generates a diff comment on PR

george-gca:
Splitted prettier GitHub action into two:
- `on push`, only runs on direct pushes, if test fails generates an artifact (that lasts for 3 days) with an html version of the changes needed to pass prettier test
- `on PR`, only runs on PRs, if test fails comments on the PR with the HTML content of the diff

I couldn't actually test the `on PR` version since it needs to be on a PR in the master branch, so this will only be triggered after this PR is accepted, and for the next PR that fails prettier test.

PS: currently the artifact is a zip file with the html inside. It is not currently possible to generate it other way, we have to wait for [this issue](https://github.com/actions/upload-artifact/issues/14) to be closed.

### DIFF
--- a/.github/workflows/prettier-on-push.yml
+++ b/.github/workflows/prettier-on-push.yml
@@ -31,7 +31,7 @@ jobs:
           diff2html -i file -s side -F diff.html -- diff.txt
       - name: Upload html diff
         if: ${{ failure() && steps.prettier.conclusion == 'failure' }}
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: HTML Diff
           path: diff.html
--- a/.github/workflows/prettier-on-push.yml
+++ b/.github/workflows/prettier-on-push.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout 🛎️
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
       - name: Setup Node.js ⚙️
         uses: actions/setup-node@v4
       - name: Install Prettier 💾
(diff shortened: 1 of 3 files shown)


.github/workflows/prettier-on-push.yml
```diff
@@ -0,0 +1,38 @@
...
+          git diff -- . ':(exclude)package-lock.json' ':(exclude)package.json' > diff.txt
+          npm install -g diff2html-cli
+          diff2html -i file -s side -F diff.html -- diff.txt
+      - name: Upload html diff
+        if: ${{ failure() && steps.prettier.conclusion == 'failure' }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: HTML Diff
+          path: diff.html
+          retention-days: 3
```
alshedivat, line 38: I'm not sure if we need this artifact generation on pushes. most of the contributions are done through PRs anyway and prettier check runs on PRs. I can also protect the master branch from direct pushes, making PRs the only way to submit changes. what do you think?

alshedivat (approved):
LGTM, this looks great. please test the PR diff logic once this one is merged.

pourmand1376 (approved):
Great! I always wanted to add this thing to our source code. But for some reason, I always put it off! 

Thanks for this great PR.

.github/workflows/prettier-on-push.yml
george-gca: It can be a good idea to protect the master branch, but independently I think uploading this artifact is useful.

For example, sometimes you want to fix small things and then end up doing this from the web interface instead of from local setup. For these cases having a report of what are the changes that you need to do instead of having to "guess" or run from local setup is useful.

Also a lot of our users are not from coding background and might not have installed prettier in their setup, so having a way to check for errors is nice. For this matter it would be useful to also have prettier installed as a local dependency in the template, and somehow setup to be run before submitting the PRs, but idk how to do that.

george-gca:
> Great! I always wanted to add this thing to our source code. But for some reason, I always put it off!
> 
> Thanks for this great PR.

Kudos for @alshedivat for doing this, I just splitted it and added the artifact o/

PR #2110, opened 2024-01-22
Improved issues templates

george-gca:
Since a lot of issues opened are just questions that should be opened as so, and sometimes the issue template is just not followed, I decided to make opening new issues a little more "enforced", if not clear.

I based the new templates on the ones from [pytorch lightning](https://github.com/Lightning-AI/pytorch-lightning/issues/new/choose), which I think are pretty good. The visuals will only appear after the merge is processed, but you can have an idea of how it will look like by checking their templates and how they look.


.github/ISSUE_TEMPLATE/1_bug_report.yml
```diff
@@ -0,0 +1,98 @@
...
+    id: os
+    attributes:
+      label: What operating system are you using?
+      description: select all OSs where you have experienced this issue
+      multiple: true
+      options:
+        - Linux
+        - Mac
+        - Windows
+        - Not applicable
```
alshedivat, line 85: what does it mean that OS "not applicable"?

.github/ISSUE_TEMPLATE/1_bug_report.yml
```diff
@@ -0,0 +1,98 @@
...
+
+  - type: dropdown
+    id: environment
+    attributes:
+      label: Where are you seeing the problem on?
+      description: select all environments where you have experienced this issue
+      multiple: true
+      options:
+        - "Running locally with Docker"
+        - "Running locally without Docker"
```
alshedivat, line 70: if running locally w/o docker, it'd be good to ask about jekyll and ruby versions

alshedivat (approved):
LGTM, thanks for adding a much better template! left just a couple quick suggestions.

pourmand1376 (approved):

.github/ISSUE_TEMPLATE/1_bug_report.yml
george-gca: I put it there for the case when the error is when running on web, like after deploying. What do you think?

george-gca: Changed text to explicit when it is not applicable.

PR #2113, opened 2024-01-22
Fixed prettier actions

george-gca:
After some thorough testing, finally made prettier on PR work. Since an html file can't be used as comment on GitHub, changed to simply point to the action that failed and to the file containing the html diff.


alshedivat:
do you have an example PR that demonstrates the diff?

george-gca:
Yes, [this one](https://github.com/george-gca/multi-language-al-folio/pull/14) in my repo.

PR #2117, opened 2024-01-23
Minor Improvements:

yazdanimehdi:
- Add ordering to the cv items
- Centering the footer text
- Adding links to publications
- Capitalizing the first letter in profile
- Adding courses to CV

### DIFF
--- a/_layouts/about.liquid
+++ b/_layouts/about.liquid
@@ -1,6 +1,7 @@
 ---
 layout: default
 ---
+<!-- about.html -->
 <div class="post">
   <header class="post-header">
     <h1 class="post-title">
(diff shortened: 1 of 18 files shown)


_layouts/default.liquid
```diff
@@ -14,7 +14,7 @@
     {% include header.liquid %}
 
     <!-- Content -->
-    <div class="container mt-5" role="main">
```
george-gca, line 17: This role here is used for [accessibility purposes](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/main_role). I believe it should be kept.

_layouts/about.liquid
george-gca, line 1: I don't know about this change, since all the titles in the header are also lowercased. It is kind of the site aesthetics. What do you think @alshedivat?

george-gca (changes requested):
In a lot of files were included some comments at the top, like `<!-- about.html -->`. Is there a reason for this?

yazdanimehdi:
> In a lot of files were included some comments at the top, like `<!-- about.html -->`. Is there a reason for this?

No, I can remove them.

_layouts/default.liquid
yazdanimehdi: ok. I will revert the changes in the default.liquid.

PR #2119, opened 2024-01-23
Improvements

yazdanimehdi:
- Adding ordering to the cv items
- Adding clickable links to publications
- Capitalize the first letter of titles
- Centering the footer text
- Adding Courses to the CV
- Fixing latest post template
- Removing the unnecessary comments
- Reverting the accessibility


alshedivat:
thanks for this contribution! one issue with this PR is that it bundles together a few unrelated changes. some of your changes we'd be happy to accept (e.g., bug fixes), but other stylistic changes that alternate the current design (such as capitalizing titles) won't be accepted. some changes are kind of borderline, and I'm not sure whether we want them or not without looking into it closer.

my recommendation is to split this PR into multiple, and address one particular thing in each PR. this would make it much easier to review and discuss proposed changes.

PR #2129, opened 2024-01-27
Added Echarts support

george-gca:
Since I added support for chart.js and while doing so found some other wonderful charting libraries, I thought, why not give support to some more? So adding support to [Apache Echarts](https://echarts.apache.org/en/index.html).

image

image


alshedivat (approved):
looks awesome, thanks for adding this feature!

PR #2130, opened 2024-01-27
GeoJSON support

george-gca:
Added support for GeoJSON maps.

image


alshedivat (approved):
this looks super cool, thanks for adding this feature!

PR #2132, opened 2024-01-27
Advanced image viewers

george-gca:
Image slider with [Swiper](https://swiperjs.com/).

image

Image comparison slider with [img-comparison-slider](https://img-comparison-slider.sneas.io/).

image


george-gca:
@alshedivat just now realized that after my changes to the GitHub actions the build process is not happening after the check for broken links here. Any idea on how to make it work for PRs again?

alshedivat:
> @alshedivat just now realized that after my changes to the GitHub actions the build process is not happening after the check for broken links here. Any idea on how to make it work for PRs again?

hmm, not sure. is the workflow name correctly specified in the `workflow_run` section?

george-gca:
Yes it is. Check the [broken links](https://github.com/alshedivat/al-folio/blob/master/.github/workflows/broken-links.yml) workflow and the [deploy](https://github.com/alshedivat/al-folio/blob/master/.github/workflows/deploy.yml) one. Also they run correctly after the PR is accepted into master.

george-gca:
I am thinking now in creating a new workflow to work just for PRs to ensure a successful build. But I'll do this in another PR.

PR #2165, opened 2024-02-06
center footer text by default

eolesinski:
This PR changes the footer text on the website (for both the sticky and fixed footer) to be centered by default. This change is based on the discussion post #257.


george-gca (approved):

george-gca:
Thanks for the change.

Issue #2087, opened 2024-01-16
Implement `axe` accessibility suggestions


george-gca:
**Is your feature request related to a problem? Please describe.**

I implemented support for [axe accessibility tests](https://github.com/dequelabs/axe-core-npm/tree/develop/packages/cli) in #2080. I decided by default to let this be run only manually, since fixing some of the issues is not straightforward. I ran it once on the [main page](https://github.com/alshedivat/al-folio/actions/runs/7545927376/job/20542599488) (about page) of the template and also in the [blog main page](https://github.com/alshedivat/al-folio/actions/runs/7545956596/job/20542689121), and here are the issues.

For the about page:

```
  Violation of "landmark-banner-is-top-level" with 1 occurrences!
    Ensures the banner landmark is at top level. Correct invalid elements at:
     - .post-header
    For details, see: https://dequeuniversity.com/rules/axe/4.8/landmark-banner-is-top-level

1 Accessibility issues detected.
Testing complete of 1 pages
```

For the blog main page:

```
  Violation of "color-contrast" with 36 occurrences!
    Ensures the contrast between foreground and background colors meets WCAG 2 AA minimum contrast ratio thresholds. Correct invalid elements at:
     - .post-meta > a[href$="distill/"]
     - a[href="/blog/2021"]
     - .post-meta > a[href$="code/"]
     - a[href="/blog/2015"]
     - li:nth-child(1) > .post-meta
     - li:nth-child(1) > .post-tags
     - li:nth-child(1) > .post-tags > a[href="/blog/2023"]
     - li:nth-child(1) > .post-tags > a[href$="formatting"]
     - a[href$="diagrams"]
     - li:nth-child(1) > .post-tags > a[href$="sample-posts"]
     - li:nth-child(2) > .post-meta
     - li:nth-child(2) > .post-tags
     - li:nth-child(2) > .post-tags > a[href="/blog/2023"]
     - li:nth-child(2) > .post-tags > a[href$="formatting"]
     - a[href$="bib"]
     - li:nth-child(2) > .post-tags > a[href$="sample-posts"]
     - li:nth-child(3) > .post-meta
     - li:nth-child(3) > .post-tags
     - li:nth-child(3) > .post-tags > a[href="/blog/2023"]
     - li:nth-child(3) > .post-tags > a[href$="formatting"]
     - a[href$="jupyter"]
     - li:nth-child(3) > .post-tags > a[href$="sample-posts"]
[...]

PR #2169, opened 2024-02-07
implemented axe accessibility changes for about + blog

brielle5810:
aria-labels in relevant links
"landmark-banner-is-top-level" violation in about changed
change of bullet points from paragraphs to list items
discernible text/color contrast violations fixed
For issue #2087

### DIFF
--- a/_layouts/about.liquid
+++ b/_layouts/about.liquid
@@ -1,8 +1,8 @@
 ---
 layout: default
 ---
-<div class="post">
-  <header class="post-header">
+<body>
+  <div class="post">
     <h1 class="post-title">
       {% if site.title == 'blank' %}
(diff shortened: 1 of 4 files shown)


george-gca:
@brielle5810 thanks for the hard work. I would love for you to give me some explanation about these changes, so that I can understand how these work. For example, I saw that you added a `body` tag to [_layouts/about.liquid](https://github.com/alshedivat/al-folio/pull/2169/files#diff-9a9434af40c4503e3ca4fbd0e87e85a1518b3d46f8e4b1ff44c6b43741aa5ca6), but this page uses `layout: default` internally, which already has a `body` tag, so how does this work?

I saw that the font is a little darker, I figured that some color changes was needed. The only thing is that in the blog page dark version, the `Newer` word is barely readable.

image

PR #2172, opened 2024-02-08
Update README.md

deepanshudagdi:
Contributor and Maintener


george-gca:
Please add yours to the end of the list, so we could keep a more `append` approach to the academics sites.

alshedivat:
tried opening the link, but it says 404 site note found

PR #2183, opened 2024-02-11
Implement Default Lazy Loading for Images with Configurable Option

ZL-Asica:
This PR introduces an enhancement to the site's image-loading strategy by implementing lazy loading as the default behavior for all images. The motivation behind this change is to improve the site's load times.

**Changes:**
1. **Default Lazy Loading:** By `default`, all images will now have the `loading="lazy"` attribute.

2. **Configurable Option in `_config.yml`:** Added a new configuration option under `lazy_loading_images.enable`.

3. **Override Capability:** For `specific images` where the user doesn't want lazy loading, they can `override` this by directly setting the `loading attribute` in the `image tag`.


alshedivat (approved):

alshedivat:
thanks for this contribution! the change looks good to me, will merge.

PR #2188, opened 2024-02-12
remove the space in names of "site"

furkanakkurt1335:
I removed the spaces between the first, middle and last names in the header.

It's just my preference, I don't know if it would be accepted but I have done this for my own page using this repository as a template.


george-gca:
@alshedivat is there a reason those characters are used?

alshedivat:
> @alshedivat is there a reason those characters are used?

no idea tbh. if it looks the same or similar, I'm in favor of removing.

alshedivat (approved):

Issue #2185, opened 2024-02-11
Line break before comma in author list.


goldblum:
### Have you checked that your issue isn't already filed?

- [X] I read through [FAQ](https://github.com/alshedivat/al-folio/blob/master/FAQ.md) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [X] Yes, I have checked that this issue isn't already filed.

### Bug description

In author lists, the lines often break right before a comma so the next line starts with a comma.

### How to reproduce the bug

This issue happens when you leave max_author_limit blank in _config.yml. See https://goldblum.github.io/publications/.  I attached a screenshot.
<img width="991" alt="Screenshot 2024-02-11 at 09 53 38" src="https://github.com/alshedivat/al-folio/assets/442416/e2453d4d-20b0-4e20-8397-f401175d20be">

### What operating system are you using?

Not applicable (e.g. you're using GitHub Pages or other hosting)

### Where are you seeing the problem on?

Deployed site

george-gca:
Related discussion #2153.

goldblum:
> Related discussion #2153.

Is the solution to the line breaks to remove the space before the commas?

george-gca:
Probably yes, but it is kind of tricky to find the right places to remove the spaces. If you manage to do so, please submit a PR.

goldblum:
Should it be on line 104 of bib.liquie:
`            {% assign more_authors_show = more_authors_show | append: ', ' %}
`
to
            {%- assign more_authors_show = more_authors_show | append: ', ' %}

goldblum:
That didn't work

george-gca:
Like I said, it is kinda tricky. Probably needs the `-` in a few more places.

PR #2197, opened 2024-02-14
fix author spacing

eolesinski:
This PR fixes issue https://github.com/alshedivat/al-folio/issues/2185 and addresses discussion https://github.com/alshedivat/al-folio/discussions/2153.

- removes leading whitespace before commas in author list
- fixes author spacing

### DIFF
--- a/_layouts/bib.liquid
+++ b/_layouts/bib.liquid
@@ -50,60 +50,60 @@
         {% assign author_array_limit = site.max_author_limit %}
       {% endif %}
 
-      {% for author in entry.author_array limit: author_array_limit %}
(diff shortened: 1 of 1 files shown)


george-gca:
I always figured we had to add - to a lot of places, never figured it would be too many. Thanks for the fix.

george-gca (approved):

eolesinski:
yeah - it definitely was a bit tricky to find all of them

egmompo:
Lines 89 & 90 in bib.liquid should also be modified in order to read like lines 85 & 86. Sorry, I don't know how to commit a PR.

PR #2233, opened 2024-02-27
Add functionality to open external links in jupyter notebooks in new tab

mscherrmann:
See https://github.com/alshedivat/al-folio/pull/2230


george-gca (approved):

george-gca:
You know you could just send new commits to that branch and the code in the PR would be updated, right?

LGTM, thanks for the PR.

george-gca:
Now just missing fixing the `prettier` formatter complaints, and it is good to go.

mscherrmann:
Hi, thank you for your advice, I will keep that in mind in the future.

PR #2237, opened 2024-02-29
Update INSTALL.md

SasikaA073:
Added "Recommended Approach" as a subheading which was not under Installing , but was in the contents section.


george-gca (approved):
Thanks for noticing and fixing this.

PR #2243, opened 2024-03-02
Correct config variable for enabling latest_posts on about page

slapcat:
I noticed disabling latest_posts in `_config.yml` didn't work because the variable in the liquid template was seemingly incorrect. This should fix that.

### DIFF
--- a/_layouts/about.liquid
+++ b/_layouts/about.liquid
@@ -49,7 +49,7 @@ layout: default
     {% endif %}
 
     <!-- Latest posts -->
-    {% if page.latest_posts %}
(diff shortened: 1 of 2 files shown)


george-gca (changes requested):
Actually if you look at the [about page](https://github.com/alshedivat/al-folio/blob/master/_pages/about.md) you can see that we have a `latest_posts: true` in its frontmatter. But I agree that that is unnecessary. Could you also remove it from there?

slapcat:
Good call, I hadn't noticed it there as well. I removed it.

george-gca (approved):

PR #2245, opened 2024-03-03
Add option for local HTML files

slapcat:
Add option to link HTML local files from publications page.


george-gca (changes requested):
Instead of pointing to a fully fledged html page with lots of images, could you point to a dummy html? This way we would only need a new simple html file instead of a lot of files to use as an example.

george-gca (changes requested):
Could you just change both usages of `Relativity :` to `Relativity:`?

george-gca (approved):

george-gca:
Thanks for your contribution 🎉

slapcat:
Happy to do so, thanks for the quick responses and review!

PR #2261, opened 2024-03-08
Added "follow system" option to theme toggle

SamAdamDay:
The theme toggle button now has a third option, which follows the user's system preferences.

- In the code there's now a distinction between the theme setting (which can be "dark", "light" or "system") and the computed theme.
- The theme setting is now stored as the "theme-setting" local storage variable. Since this is different from the old variable ("theme"), this will effectively reset all current user themes to the default "system". Maybe this is not what you want.
- The "system" theme icon is currently a half circle symbol.
- The toggle button now displays the current theme setting, rather than the next theme setting (as far as I know this is consistent with other sites which have three theme options).
- `theme.js` is now loaded regardless of `site.enable_darkmode`. This is because other scripts which were always loaded relied on being able to determine the theme. `theme.js` no longer initialises the theme itself though; this only happens when `site.enable_darkmode`.
- When the theme setting is "system", the theme will change immediately whenever the user changes their system theme.

#2261


george-gca:
Thanks for your solution. My only issue here is about the icon, but not really sure how to fix this. Usually the current icon is used as `contrast`, but the `moon-over-sun` icon is pro only. What do you think @alshedivat?

george-gca:
Just give a thought about this, and one possible solution to keep the icons consistent between them is to use another iconset for this specific part. What do you guys think about using [tabler icons](https://tabler.io/icons)? More specifically, [moon](https://tabler.io/icons/icon/moon), [sun](https://tabler.io/icons/icon/sun), and [sun-moon](https://tabler.io/icons/icon/sun-moon). It looks like this in the template:

image

image

image

And it also supports usage via scss, so after purgecss does its part, only the used icons will be kept on the final css file.

george-gca:
Yes, it is not the prettiest one, but at least keeps the consistency in the icons. I will apply the changes and wait for @alshedivat feedback.

george-gca (approved):

alshedivat (approved):
LGTM, thanks for implementing this!

sahiljhawar:
How to toggle between only light and dark mode?

I dont want this: 
image

george-gca:
Can you ask this in Q&A? This PR was already merged, also there it is easier for more people to find it.

PR #2286, opened 2024-03-18
add flickr_url to social integration

nuance1979:
- add flickr_url to social integration

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -84,6 +84,7 @@ lattes_id: # your ID on Lattes (Brazilian Lattes CV)
 osf_id: # your OSF ID
 research_gate_profile: # your profile on ResearchGate
 scopus_id: # your profile on Scopus
+flickr_url: # your flickr URL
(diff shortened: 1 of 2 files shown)


_config.yml
```diff
@@ -84,6 +84,7 @@ lattes_id: # your ID on Lattes (Brazilian Lattes CV)
 osf_id: # your OSF ID
 research_gate_profile: # your profile on ResearchGate
 scopus_id: # your profile on Scopus
+flickr_url: # your flickr URL
```
george-gca, line 87: Instead of setting the entire flickr url here, it would be better if you set just the user, and then in `_includes/social.liquid` construct the url from the user, as it is done with most of the other socials.

george-gca (changes requested):

_config.yml
nuance1979: Sure.

george-gca (approved):

PR #2297, opened 2024-03-26
Update cv.liquid

AneeshBose:
Added Table of Contents which was there in earlier versions, but missing in current version

### DIFF
--- a/_layouts/cv.liquid
+++ b/_layouts/cv.liquid
@@ -21,6 +21,17 @@ layout: default
       {% endif %}
     </header>
 
+    <h4>Table of contents</h4>
+    <ul class="timeline">
+      {% for entry in site.data.cv %}
+        <li>
+          <a href="#{{ entry.title }}"
+            ><span class="badge-toc">{{ entry.title }}</span></a
+          >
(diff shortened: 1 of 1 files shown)


alshedivat:
I'm not sure if table of contents was removed intentionally or accidentally. @george-gca, do you know?

george-gca:
I don't remember now, but I believe it was removed since it currently already displays a table of contents on the side, as can be seen on the [demo site](https://alshedivat.github.io/al-folio/cv/).

AneeshBose:
> I don't remember now, but I believe it was removed since it currently already displays a table of contents on the side, as can be seen on the [demo site](https://alshedivat.github.io/al-folio/cv/).

The toc by the side does not look visually appealing and the text seems to bleed out for me, not sure if the others are facing the same. Which is why I had to revert to adding the toc in the beginning. If there is a picture I can help with, please let me know!

george-gca:
> the text seems to bleed out for m

What do you mean? For me it looks like this, and it seems fine:
image

alshedivat:
closing since toc on side works as expected

PR #2298, opened 2024-03-26
Update work.liquid to show the location of work experience under cv page

christopherjang1:
adding "location" element when location object is added to work experience. Location icon and the entered value will appear on the CV page under the work experience section. This feature already exists for education section, but not for work. The change is made to make the experiece more consistent.

image


george-gca (approved):

george-gca:
Thanks for your contribution.

PR #2306, opened 2024-03-30
Remove bootstrap-table from archive tables

sim642:
The blog archive page tables have `table-borderless`, but bootstrap-table JS adds `table-bordered` by default and this forces an `!important` border.
Since these tables are also supposed to be borderless, this PR excludes them from bootstrap-table, just like the news and CV table are.

### DIFF
--- a/_layouts/archive-category.liquid
+++ b/_layouts/archive-category.liquid
@@ -7,7 +7,7 @@ layout: default
     <p class="post-description">an archive of posts in this category</p>
   </header>
 
-  <article>
+  <article class="archive">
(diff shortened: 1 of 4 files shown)


george-gca (approved):

george-gca:
As soon as the prettier formatting is fixed I will accept this PR. Check the [artifacts](https://github.com/alshedivat/al-folio/actions/runs/8493298707) for more information on what to fix if you don't want to run prettier locally.

george-gca (approved):

PR #2312, opened 2024-04-02
Offline mode (closes #1181)

### LINKED ISSUE #1181: Run al-folio without external ressources/cdns | Privacy / Offline Version

In its default configuration, al-folio uses lots of externally hosted resources. 
For example, it uses JavaScript libraries like `bootstrap`, `imagesloaded`, `jquery`, `masonry-layout`, `mdbootstrap`, `medium-zoom` that are hosted on `cdn.jsdelivr.net`.
Also, Google Font `fonts.googleapis.com` is used per default.
They can be found in  [header.html](https://github.com/alshedivat/al-folio/blob/master/_includes/head.html).

Some features, like `altmetric_badge` and `dimensions_badge`, can be disabled simply in `_config.yml`. By their nature, these features cannot work on a privacy oriented website since they must request their respective API services.

I understand the benefits of using CDNs, but I would love to have the possibility to run the website with respect to privacy.
The best option IMHO would be a configuration option like `privacy_mode`. It could also be called `offline_mode` since, since it provides the advantage to run the site while being offline.

When activated, this option should automatically download all required libraries (e.g., those that are defined in `_config.yml`) and put them in an appropriate assets folder.
`header.html` should then use these files instead.

I looked into two different directions:

1. I looked into existing jekyll modules if anything for this purpose exists. I could not really find anything, but I have to say, that I'm very new to jekyll. So I might overlooked something.

2. I could simply download the files manually, place them in my `assets/js` folder and edit `header.html`. I'm not sure if this is the expected way to do so. Future updates of al-folio are likely to cause trouble.

george-gca:
Created a plugin to tackle #1181. Currently have an issue with tikzjax since it imports some wasm file from its javascript. The rest should work as expected.


alshedivat:
@george-gca, looks like this PR adds multiple plugins / modifies a few. could you elaborate on all added modifications?

george-gca:
It actually only adds one plug-in. The changes made to the others was just about indentation, to keep a standard. Some of them used 4 spaces, others 2.

george-gca:
I also renamed one plugin file to keep it consistent with the others. And moved almost all urls from libraries to `_config.yml` to allow downloading them from the new plugin in `_plugins/download-3rd-party.rb`

alshedivat (approved):

PR #2324, opened 2024-04-07
Feature: added "award" for publications

CheariX:
This PR adds an "award" button to publications.
It takes the `award` value from the bibtex entry and displays(incl. Markdown rendering) the text in a box similarly to abstract and bibtex.

User can set the entry `award_name` to configure the value. See example config with `award_name: Nobel Prize`.

The color of the award box can be configured in `_base.scss`.

Note, there is a similar PR #2175, it I saw to issues with:

1. There was no progress
2. The award button just prints the text directly in the button, similarly to `award_name`. Long award names could clutter the webpage.
3. IMHO, it brokes the current al-folio design, since butons do have a fixed size/text. However, variable prize names are also possible with this PR.
***
Pictures:

Default. Text are hidden:

grafik

Additional info is shown when the button is clicked. Markdown supported.

grafik

Only one text box shown at the same time, like it is with "ABS" and "BIB":
grafik

***

Feedback welcome.

You can also check [my website](https://christianmainka.de/publications/awarded), which was the base for this PR.

### DIFF
--- a/_bibliography/papers.bib
+++ b/_bibliography/papers.bib
@@ -83,6 +83,21 @@
   year={1905}
 }
 
+@Article{einstein1905photoelectriceffect,
+  bibtex_show={true},
+  abbr={Ann. Phys.},
+  title="{{\"U}ber einen die Erzeugung und Verwandlung des Lichtes betreffenden heuristischen Gesichtspunkt}",
+  author={Albert Einstein},
+  abstract={This is the abstract text.},
+  journal={Ann. Phys.},
+  volume={322},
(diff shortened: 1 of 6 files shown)


CheariX:
Default. Text are hidden:

grafik

Additional info is shown when the button is clicked. Markdown supported.

grafik

Only one text box shown at the same time, like it is with "ABS" and "BIB":
grafik

george-gca (approved):

george-gca:
Nice solution. Thanks for your contribution.

PR #2332, opened 2024-04-10
[MINOR] Re-enable the image caption on the about page

ywcb00:
Hi,
this PR re-enables the image caption for the profile picture on the about page by passing the caption parameter to the figure include.


george-gca:
Why would the profile picture need caption? The alt text seems enough to me, also we have more info being displayed below the profile picture.

alshedivat:
closing as unplanned / unnecessary feature

PR #2337, opened 2024-04-13
feat: make video embeddings optional.

CheariX:
As discussed in #1181, I suggest to make embedding videos an optional feature.

This behavior aligns well with recently merged PR #2312.

Open questions:

1. I added a youtube link to `papers.bib`. Is this link okay?
2. I set `enable_video_embedding: false` as the default. I argue that privacy settings should be the default. Also, the current implementation of `video.liquid` only works for some very specific video URLs. For example, to embed youtube, specialized links must be used to avoid `X-Frame-Option` issues. This behavior can lead to a broken embedding, which would not look very nice.

Feedback welcome.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -390,6 +390,7 @@ enable_navbar_social: false # enables displaying social links in the navbar on t
 enable_project_categories: true # enables categorization of projects into multiple categories
 enable_medium_zoom: true # enables image zoom feature (as on medium.com)
 enable_progressbar: true # enables a horizontal progress bar linked to the vertical scroll position
+enable_video_embedding: false # enables video embedding for bibtex entries. If false, the button opens a the video link in a new window.
 
 # -----------------------------------------------------------------------------
 # Library versions
--- a/_config.yml
+++ b/_config.yml
@@ -390,7 +390,7 @@ enable_navbar_social: false # enables displaying social links in the navbar on t
 enable_project_categories: true # enables categorization of projects into multiple categories
 enable_medium_zoom: true # enables image zoom feature (as on medium.com)
 enable_progressbar: true # enables a horizontal progress bar linked to the vertical scroll position
-enable_video_embedding: false # enables video embedding for bibtex entries. If false, the button opens a the video link in a new window.
+enable_video_embedding: false # enables video embedding for bibtex entries. If false, the button opens the video link in a new window.
 
 # -----------------------------------------------------------------------------
 # Library versions
--- a/_bibliography/papers.bib
+++ b/_bibliography/papers.bib
@@ -53,6 +53,7 @@
   altmetric={248277},
   dimensions={true},
   google_scholar_id={qyhmnyLat1gC},
+  video={https://www.youtube-nocookie.com/embed/aqz-KE-bpKQ},
   selected={true}
 }
 
--- a/_layouts/bib.liquid
+++ b/_layouts/bib.liquid
@@ -189,8 +189,10 @@
           <a href="{{ entry.supp | prepend: '/assets/pdf/' | relative_url }}" class="btn btn-sm z-depth-0" role="button">Supp</a>
         {% endif %}
       {% endif %}
-      {% if entry.video %}
+      {% if entry.video and site.enable_video_embedding %}
         <a class="abstract btn btn-sm z-depth-0" role="button">Video</a>
+      {% elsif entry.video %}
+        <a href="{{ entry.video }}" class="btn btn-sm z-depth-0" role="button">Video</a>
       {% endif %}
       {% if entry.blog %}
         <a href="{{ entry.blog }}" class="btn btn-sm z-depth-0" role="button">Blog</a>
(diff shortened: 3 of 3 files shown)


_config.yml
```diff
@@ -390,6 +390,7 @@ enable_navbar_social: false # enables displaying social links in the navbar on t
 enable_project_categories: true # enables categorization of projects into multiple categories
 enable_medium_zoom: true # enables image zoom feature (as on medium.com)
 enable_progressbar: true # enables a horizontal progress bar linked to the vertical scroll position
+enable_video_embedding: false # enables video embedding for bibtex entries. If false, the button opens a the video link in a new window.
```
george-gca, line 393: small typo here in the comment

george-gca:
> I added a youtube link to papers.bib. Is this link okay?

Yes.

> I argue that privacy settings should be the default.

You also mean the one that downloads the 3rd party libraries?

> Also, the current implementation of `video.liquid` only works for some very specific video URLs. For example, to embed youtube, specialized links must be used to avoid `X-Frame-Option` issues. This behavior can lead to a broken embedding, which would not look very nice.

This PR solves this issue? If not, I believe this would be a nice addition to our README, maybe in `FAQ.md`?

Also, I believe the current solution (displaying the video below in a small collapsible area) for the bib part is not worth it. It is not even displaying properly, also the video is too small to be useful. I think we should change it to always open in a new window. What do you think?

george-gca (changes requested):

_config.yml
CheariX: Good catch. Thanks.

CheariX:
> > I argue that privacy settings should be the default.
> 
> You also mean the one that downloads the 3rd party libraries?

Yes. Basically, I'd like to change these settings in `_config.yml`:
```
enable_publication_badges:
  altmetric: false
  dimensions: false
  google_scholar: false

third_party_libraries:
  download: true
```
By implementing these settings, the website will not send requests to any third-party domains (a desirable feature in compliance with EU/GDPR laws, as there is no need for a consent banner). However, I have noticed one exception: `https://distill.pub/third-party/katex/katex.min.js`. Unfortunately, I could not find any means to disable this particular connection.

People who are fine with third party requests can then easily enable these features in the config.

Maybe we can open a PR or Issue to address and discuss this issue separately.

> 
> > Also, the current implementation of `video.liquid` only works for some very specific video URLs. For example, to embed youtube, specialized links must be used to avoid `X-Frame-Option` issues. This behavior can lead to a broken embedding, which would not look very nice.
> 
> This PR solves this issue? If not, I believe this would be a nice addition to our README, maybe in `FAQ.md`?

- `video.liquid` works for `https://www.youtube.com/embed/aqz-KE-bpKQ` => **/embed/**
- `video.liquid` does not work for `https://www.youtube.com/watch?v=aqz-KEP-bpKQ` => **/watch?v=**
- Please note that I used `youtube-nocookie.com` instead of `youtube.com` in  `paper.bib`. But the result is the same for both domains: one must use the **embed** link to use youtube in an iframe.

I'm not sure what one should expect here.
I always had youtube links in my `video` tag, but no dedicated embedding links. 
We could do some find/replace magic to allow regular  links, but I did not want to touch `video.liquid` since I'm not gonna use the embedding feature anyway.

> Also, I believe the current solution (displaying the video below in a small collapsible area) for the bib part is not worth it. It is not even displaying properly, also the video is too small to be useful. I think we should change it to always open in a new window. What do you think?

I'd be totally happy with that. If you like, I can add this removal to this PR.

Not sure about youtube, but I think one use case was that some people hosted their own `*.mp4/webm/ogg` files and refered them in the `video` tag

george-gca:
> People who are fine with third party requests can then easily enable these features in the config.

I would vote for the opposite, at least regarding the publication badges. It is useful to demonstrate how this would look like in our demo site, in case the user wants to keep it. Maybe add something citing how/why to disable this regarding privacy issues, probably including some of the information already cited here, in one of our README files, maybe `CUSTOMIZE.md`.

> I'd be totally happy with that. If you like, I can add this removal to this PR.
>
> Not sure about youtube, but I think one use case was that some people hosted their own *.mp4/webm/ogg files and refered them in the video tag

I think this addition would be nice to have. Even though someone might embed their own videos there, the viewport is too small to be useful (I think). If the user wants, it would be more useful to open it in a new tab, or even open it in a way similar to a zoomed image works.

CheariX:
> I think this addition would be nice to have. Even though someone might embed their own videos there, the viewport is too small to be useful (I think). If the user wants, it would be more useful to open it in a new tab, or even open it in a way similar to a zoomed image works.

What kind of addition do you mean?
Do you want any further changes in this PR? 

The embedding feature can be enabled with the `enable_video_embedding: true` option.

george-gca:
>> Also, I believe the current solution (displaying the video below in a small collapsible area) for the bib part is not worth it. It is not even displaying properly, also the video is too small to be useful. I think we should change it to always open in a new window. What do you think?
>
> I'd be totally happy with that. If you like, I can add this removal to this PR.

I was talking about this part.

george-gca (approved):

george-gca:
Thanks for the changes.

PR #2349, opened 2024-04-16
 Update work.liquid to show the location of work experience under cv …

alerezaaa:
…page

same as pull request https://github.com/alshedivat/al-folio/pull/2298 and commit https://github.com/alshedivat/al-folio/commit/363f277aa82df6e03f172dbcf2f9c1fbe84aca26 for **Volunteer** section

### DIFF
--- a/_includes/resume/volunteer.liquid
+++ b/_includes/resume/volunteer.liquid
@@ -12,7 +12,7 @@
           {% else %}
             {% assign date = '' %}
           {% endif %}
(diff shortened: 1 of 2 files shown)


george-gca:
Now just missing fixing the `prettier` formatter complaints, and it is good to go. If you don't have `prettier` installed for your setup, you can check the referred failed action in our repo. In that action will be an artifact with an HTML diff with the needed changes.

PR #2357, opened 2024-04-20
Feat reworked project layout (closing #2246)

### LINKED ISSUE #2246: Harmonizing width and height of cards in /projects/

### Description & Motivation

Have an option to set the cards in /projects/ to be the same height in a given row and their width so that they nicely fill the entire text width.

### Pitch

The /projects/ layout has cards of fixed width of 250px and variable height depending on the content. It would render better if the cards are of same height (the one of the longest element grid-item in the row).
Additionally, the cards are left-ragged. Aesthetically, it would render better if their width and padding are such that the cards fill the entire text width.
Would that be possible? Thanks!

### Alternatives

_No response_

### Additional context

_No response_

CheariX:
The current state of project looks a bit unharmonized.

# Vertical layout before:

Bildschirmfoto 2024-04-20 um 18 09 16

# Vertical layout after:

Bildschirmfoto 2024-04-20 um 18 10 28

# Horizontal layout before:

Bildschirmfoto 2024-04-20 um 18 08 54

# Horizontal layout after:

Bildschirmfoto 2024-04-20 um 18 07 46

***

These improvements are not perfect, but to be honest, I have no deep experiences with bootstrap. Actually, I just read the [docs](https://getbootstrap.com/docs/4.4/components/card/) and tried my best. 
But I think it looks way better than before.

I'm pretty happy with the vertical layout, but the horizontal could be improved.
I tried various things to get a better alignment of the image, without any success.
In theroy, there is e.g. `card-img-top`, but I could not make it work.

I changed the project descriptions to make some of these edge-cases visible.

Feedback and improvements (especially code) are more than welcome

Closes #2246

### DIFF
--- a/_includes/projects.liquid
+++ b/_includes/projects.liquid
@@ -11,7 +11,7 @@
         %}
       {% endif %}
       <div class="card-body">
-        <h2 class="card-title text-lowercase">{{ project.title }}</h2>
+        <h2 class="card-title">{{ project.title }}</h2>
         <p class="card-text">{{ project.description }}</p>
         <div class="row ml-1 mr-1 p-0">
           {% if project.github %}
--- a/_includes/projects.liquid
+++ b/_includes/projects.liquid
@@ -1,7 +1,6 @@
-<div class="grid-sizer"></div>
-<div class="grid-item">
+<div class="col">
   <a href="{% if project.redirect %}{{ project.redirect }}{% else %}{{ project.url | relative_url }}{% endif %}">
-    <div class="card hoverable">
+    <div class="card h-100 hoverable">
       {% if project.img %}
         {%
           include figure.liquid
--- a/_includes/projects.liquid
+++ b/_includes/projects.liquid
@@ -8,6 +8,7 @@
           path=project.img
           sizes = "250px"
           alt="project thumbnail"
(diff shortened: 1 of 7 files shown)


CheariX:
I think I found the solution: it was some non-standard `.card-item` class which kind of *destroyed* the layout.
I removed it and remplaced it with a bootstrap standard class.

# Result

Bildschirmfoto 2024-04-21 um 11 08 24

eolesinski:
Just thinking out loud here - but it might be interesting to add an optional field to include a date + link to code/github (perhaps with an icon) on each of the project cards. This would probably be a different PR though - wondering what others think

george-gca (approved):

george-gca:
> These improvements are not perfect, but to be honest, I have no deep experiences with bootstrap. Actually, I just read the [docs](https://getbootstrap.com/docs/4.4/components/card/) and tried my best.

I can't speak for everybody, but that was exactly my case. I started implementing a few minor tweaks I'd like to see on the template, long story short, I am writing plugins for it. This is exactly how contributing to open source works, sometimes it just needs someone with a little more time, patience, practice, or even the will to try it until it works. Thanks for your contribution.

Issue #1502, opened 2023-06-23
Improve use of jekyll liquid tags


george-gca:
I just realized that I've been using the tags somewhat wrong. Not wrong, but it could be better. As I saw posted [here](https://mathieubuisson.github.io/reducing-whitespace-jekyll-html/), simply using `{%- ... -%}` instead of `{% ... %}` reduces the whitespaces created by jekyll.

This is not that simple though, since I tried a dumb replace all in our code, and it stopped showing things correctly. I'll leave it here as a reminder that this change might be applied in a lot of different places, but maybe not all of them.

alshedivat:
Closing as superseded by v1.

Two things changed underneath this issue:

1. **The templates left this repo.** Every layout, include and Sass partial this issue targets now lives in the plugin gems (`al_folio_core` and friends) — this repo is a thin starter and no longer owns any of them. Re-scoping this would mean opening a separate issue per gem.
2. **Whitespace is now machine-enforced.** Liquid formatting is handled by Prettier with `@shopify/prettier-plugin-liquid` at `printWidth: 150`, gated in CI. Hand-tuning `{%- -%}` markers would fight the formatter rather than complement it.

If there's a specific page where whitespace control visibly affects output, that's worth an issue against the owning gem.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #2378, opened 2024-04-29
fix(style): remove whitespace before comma

CheariX:
This PR fixes a very niche bug.

If there is a co-author that is not *me* (=`<em>` tagged) and does not have a `coauthor_url` (=`<a>` tagged), there will be a leading whitespace before the separating comma. If the author list is split into multiple lines (due to long author lists or a small view), the whitespace can lead to a line beginning with a comma.

As I said, very niche, but interestingly, the other cases already did the same whitespace handling. Just the case not-me/no-url was mising.

Currently, we do not have this case on vanilla al-folio, but you can see it for example on [my website](https://christianmainka.de/publications/) if the view is small (e.g., 430px).

Related to #1502.

### DIFF
--- a/_layouts/bib.liquid
+++ b/_layouts/bib.liquid
@@ -92,8 +92,8 @@
               {{ author.last -}}
             </a>
           {%- else -%}
-            {{ author.first }}
-            {{ author.last }}
+            {{- author.first }}
+            {{ author.last -}}
           {% endif %}
         {%- endif -%}
       {% endfor %}


george-gca:
Thanks for this fix. I believe we might have quite a few of these hidden under particular cases, but like this one we'll only find them when they happen.

ajyey:
@george-gca what issues were you seeing with this when you initially tried? I just did a find/replace for 174 instances across the below files and so far I haven't noticed any issues. But maybe I'm missing something obvious. Example diff below

- `_includes/citation.liquid`
- `_includes/scripts/search.liquid`
- `_layouts/bib.liquid`

CleanShot 2024-06-22 at 01 22 51@2x

george-gca:
It is quite the opposite. I was adding the `-` in those lines, like it was done in this commit.

ajyey:
> It is quite the opposite. I was adding the `-` in those lines, like it was done in this commit.

Ah okay sorry I just realized I meant to add this comment over in #1512 That being said, I think I misunderstood the original find/replace and have it backwards. Thanks for clarifying. Will mess around with it

ajyey:
> It is quite the opposite. I was adding the `-` in those lines, like it was done in this commit.

Ah okay sorry I just realized I meant to add this comment over in #1502 That being said, I think I misunderstood the original find/replace and have it backwards. Thanks for clarifying. Will mess around with it

Issue #2395, opened 2024-05-07
"Could not find gem" when running default `docker-compose.yml` for latest `al-folio` docker image


doyled-it:
### Have you checked that your issue isn't already filed?

- [X] I read through [FAQ](https://github.com/alshedivat/al-folio/blob/master/FAQ.md) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [X] Yes, I have checked that this issue isn't already filed.

### Bug description

The jekyll site does not finish initializing and is unreachable at localhost:8080.

### System
MacBook Air
Apple M3
Sonoma 14.4.1

### Steps
1. Clone the repo
2. `docker compose up` (with [this](https://hub.docker.com/layers/amirpourmand/al-folio/latest/images/sha256-c267a4cf8333f40b1e90d63eed9911c225da0b7368da8a207d13a1c4e6c36ad8?context=explore) docker image)

### Behavior

```
[+] Running 2/0
 ✔ Network doyled-itgithubio_default     Created                                                                                                                             0.0s
 ✔ Container doyled-itgithubio-jekyll-1  Created                                                                                                                             0.0s
Attaching to jekyll-1
jekyll-1  | /var/lib/gems/3.0.0/gems/bundler-2.5.7/lib/bundler/resolver.rb:332:in `raise_not_found!': Could not find gem 'jekyll-tabs' in locally installed gems. (Bundler::GemNotFound)
jekyll-1  | 	from /var/lib/gems/3.0.0/gems/bundler-2.5.7/lib/bundler/resolver.rb:392:in `block in prepare_dependencies'
jekyll-1  | 	from /var/lib/gems/3.0.0/gems/bundler-2.5.7/lib/bundler/resolver.rb:377:in `each'
jekyll-1  | 	from /var/lib/gems/3.0.0/gems/bundler-2.5.7/lib/bundler/resolver.rb:377:in `map'
jekyll-1  | 	from /var/lib/gems/3.0.0/gems/bundler-2.5.7/lib/bundler/resolver.rb:377:in `prepare_dependencies'
jekyll-1  | 	from /var/lib/gems/3.0.0/gems/bundler-2.5.7/lib/bundler/resolver.rb:61:in `setup_solver'
jekyll-1  | 	from /var/lib/gems/3.0.0/gems/bundler-2.5.7/lib/bundler/resolver.rb:28:in `start'
[...]

Issue #2392, opened 2024-05-06
docker compose up fails - could not find 'jekyll-tabs'


albusdemens:
### Have you checked that your issue isn't already filed?

- [X] I read through [FAQ](https://github.com/alshedivat/al-folio/blob/master/FAQ.md) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [X] Yes, I have checked that this issue isn't already filed.

### Bug description

I cannot run 'docker compose up'

### How to reproduce the bug

1. Clone the repo
2. Install all necessary dependencies using WSL2 - Ubuntu 22.04
3. sudo docker compose pull
4. sudo docker compose up

### Error messages and logs

```
Attaching to jekyll-1
jekyll-1  | /var/lib/gems/3.0.0/gems/bundler-2.5.7/lib/bundler/resolver.rb:332:in `raise_not_found!': Could not find gem 'jekyll-tabs' in locally installed gems. (Bundler::GemNotFound)
jekyll-1  |     from /var/lib/gems/3.0.0/gems/bundler-2.5.7/lib/bundler/resolver.rb:392:in `block in prepare_dependencies'
jekyll-1  |     from /var/lib/gems/3.0.0/gems/bundler-2.5.7/lib/bundler/resolver.rb:377:in `each'
jekyll-1  |     from /var/lib/gems/3.0.0/gems/bundler-2.5.7/lib/bundler/resolver.rb:377:in `map'
jekyll-1  |     from /var/lib/gems/3.0.0/gems/bundler-2.5.7/lib/bundler/resolver.rb:377:in `prepare_dependencies'
jekyll-1  |     from /var/lib/gems/3.0.0/gems/bundler-2.5.7/lib/bundler/resolver.rb:61:in `setup_solver'
jekyll-1  |     from /var/lib/gems/3.0.0/gems/bundler-2.5.7/lib/bundler/resolver.rb:28:in `start'
jekyll-1  |     from /var/lib/gems/3.0.0/gems/bundler-2.5.7/lib/bundler/definition.rb:626:in `start_resolution'
jekyll-1  |     from /var/lib/gems/3.0.0/gems/bundler-2.5.7/lib/bundler/definition.rb:311:in `resolve'
jekyll-1  |     from /var/lib/gems/3.0.0/gems/bundler-2.5.7/lib/bundler/definition.rb:579:in `materialize'
[...]

Issue #2384, opened 2024-05-03
Docker Image CI Workflow failing


george-gca:
### Have you checked that your issue isn't already filed?

- [X] I read through [FAQ](https://github.com/alshedivat/al-folio/blob/master/FAQ.md) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [X] Yes, I have checked that this issue isn't already filed.

### Bug description

Docker Image CI Workflow is failing due to mini_racer

### How to reproduce the bug

Check the failed [action](https://github.com/alshedivat/al-folio/actions/runs/8928060333/job/24524664008).

### Error messages and logs

Error message of interest:

```
compiling mini_racer_extension.cc
linking shared-object mini_racer_extension.so
/usr/bin/ld: cannot find
/var/lib/gems/3.2.0/gems/libv8-node-21.7.2.0-x86_64-linux/vendor/v8/x86_64-linux-gnu/libv8/obj/libv8_monolith.a:
No such file or directory
collect2: error: ld returned 1 exit status
make: *** [Makefile:265: mini_racer_extension.so] Error 1

An error occurred while installing mini_racer (0.12.0), and Bundler cannot continue.
In Gemfile:
   mini_racer
```

### What operating system are you using?

Not applicable (e.g. you're using GitHub Pages or other hosting)

### Where are you seeing the problem on?

Deployed site

### More info

It is probably due to `mini_racer` version `0.12.0`. I checked their repo and found [this issue](https://github.com/rubyjs/mini_racer/issues/300). Can you take a look at it @pourmand1376?

george-gca:
Update: just [asked on that bug](https://github.com/rubyjs/mini_racer/issues/300#issuecomment-2093711801) and they [pointed me to the cause](https://github.com/rubyjs/libv8-node/pull/36#issuecomment-2045274661).

pourmand1376:
Hmm. Thanks for pointing out the issue. 

I think we should fix versions as suggested [here](https://stackoverflow.com/questions/9265213/should-i-specify-exact-versions-in-my-gemfile) to avoid such problem. 

This would require us to commit `Gemfile.lock` and we should install dependecies based on that. 

This way we would manually update our `Gemfile.lock` and if and only if, there is no problem with upstream branches, we update. 

And to fix this one, I think we may need to downgrade our Gemfile to not use the latest version of this library.

george-gca:
[...]

PR #2400, opened 2024-05-09
Fix docker image building

Herbrant:
This PR should fix the issues reported in https://github.com/alshedivat/al-folio/issues/2384 , https://github.com/alshedivat/al-folio/issues/2392 and https://github.com/alshedivat/al-folio/issues/2395.
It seems that the ruby docker image is not affected by these problems.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:latest
+FROM ruby:latest
 ENV DEBIAN_FRONTEND noninteractive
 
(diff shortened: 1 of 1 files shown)


george-gca:
Thanks for fixing this. I hoped there was an easy solution.

george-gca (approved):

pourmand1376:
Thanks guys. @Herbrant @george-gca 

I really didn't know how to fix this issue. I've tested lots of solutions but none worked.

PR #2402, opened 2024-05-09
Updated dependencies

george-gca:


pourmand1376:
Good fix! 

```
    google-protobuf (4.26.1-aarch64-linux)
      rake (>= 13)
    google-protobuf (4.26.1-x86_64-linux)
      rake (>= 13)
```

Specifying packages like this is very convenient.

george-gca:
Actually `bundler` did all the job. I just did `bundler install` then `bundler update` on my local setup.

pourmand1376:
It's odd that bundler fixed packages for both x86_64 and arm CPUs.

george-gca:
Not really, since you added the second platform in #2396.

PR #2403, opened 2024-05-09
Fix deploy site github action

Herbrant:
I updated the `Gemfile.lock` to fix the "deploy site" Github Action.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -17,7 +17,7 @@ GEM
     bibtex-ruby (6.1.0)
       latex-decode (~> 0.0)
       racc (~> 1.7)
-    bigdecimal (3.1.8)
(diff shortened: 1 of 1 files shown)


george-gca:
Idk why, but some of the libraries in your commit are behind my last update (except for `bundler` itself). Also, what do you mean? The latest run of this action ended successfully, as you can see in our [actions tab](https://github.com/alshedivat/al-folio/actions/workflows/deploy.yml).

Herbrant:
Sorry I had not seen your commit when I was working on this PR. It seems that with your changes everything works. Thank you.

george-gca:
Actually apparently fixing the Docker image fixed it all, so thank you haha.

PR #2415, opened 2024-05-13
Added support for search

george-gca:
Added support for search within the template as suggested in #581. I decided to go with a client side search based on [Ninja keys](https://github.com/ssleptsov/ninja-keys), but using [deepdub's fork](https://github.com/deepdub-ai/ninja-keys) as basis since it supports fuzzy search.

Had to do a bunch of changes to their code to make it work without using node to install everything. Also changed to use some colors defined in our side and using both pages' titles and descriptions for search. Also had to increase the template max width to better accomodate the new item in navigation bar.

Missing implementations:
- [ ] One thing I'd love to do (but currently have no idea how) would be to change the text next to the search button depending on the platform. For example, if the user is accessing the site on a mac they should use ⌘k instead of ctrl k.
- [x] Test how this looks like (and how it is supposed to work) on devices with smaller screens
- [x] Support for offline mode

Some screenshots:

---

## Dark version

Screenshot from 2024-05-13 16-30-12

Screenshot from 2024-05-13 16-30-26

Screenshot from 2024-05-13 16-30-36

---

## Light version

Screenshot from 2024-05-13 16-30-44

Screenshot from 2024-05-13 16-30-50

Screenshot from 2024-05-13 16-30-55


pourmand1376:
Hi. @george-gca 

Thanks for adding the great feature. Images look great. 

Just a question. Do you know anyway so that I can test this online?

george-gca:
The easiest way that I can think of is for you to clone this branch and run from it. I want to add some kind of build and deploy from PR here, but I don't know if GitHub Pages support this. I saw this being done with netlify, as I commented in #1981, but never really looked into it.

george-gca:
Apparently [Netlify](https://docs.netlify.com/site-deploys/deploy-previews/) and [Vercel](https://vercel.com/docs/deployments/git) support this feature.

george-gca:
I've been trying to make Netlify work for our repo, but to no avail. Currently it fails with

```
Liquid Exception: No such file or directory @ rb_sysopen - /opt/build/repo/./_bibliography/references.bib in /opt/build/repo/_posts/en-us/2023-07-12-post-bibliography.md
```

Even though the string `references.bib` doesn't even exist in the repo.

george-gca:
I was FINALLY able to add the build to Netlify. I believe it would be a great addition to our repo to do builds based on PRs, and that can be done while automatically inserting the link to the build page here. Idk if I have the permission to add this in our repo, but I can teach the one with the right permissions @alshedivat @pourmand1376.

[Here is the link](https://ninja-keys--multi-language-al-folio.netlify.app/) to the test.

pourmand1376:
Thanks @george-gca .

I tested it and I like it.
 
I just have a proposal. If you could close the navbar after opening the search, it would give me a lot better experience as a user. Other than that it is very good. 

Also, I don't have access to add Netlify. Only the owner could do it.

pourmand1376 (approved):

george-gca:
> I just have a proposal. If you could close the navbar after opening the search, it would give me a lot better experience as a user. Other than that it is very good.

What do you mean?

> Also, I don't have access to add Netlify. Only the owner could do it.

I think I'm going to open a new issue for this, then we can track that there.

pourmand1376:
> What do you mean?

I mean on web version there is no problem. But on Android and smaller devices, I like to the navbar to automatically hide after pressing the search button. Otherwise, I first have to manually close the navbar then I can starting searching.

george-gca:
Ahhh ok, I can fix that. I will also make the z index of both the header and the footer be inferior to the one from the search. Just now realized that they are above the fade.

george-gca:
@pourmand1376 do you know how to do hide the navbar properly? I tried to simply remove the `show` class from it, but it have 2 side effects:
1. it hides abruptly and without animation
2. it keeps the icon on the navbar as it were opened

george-gca:
I think I am going to push this PR into our main branch already, and create 2 new issues:
1. improvement - change the text next to the search button depending on the platform. For example, if the user is accessing the site on a mac they should use ⌘k instead of ctrl k
2. bug - fix when on mobile devices

pourmand1376:
> @pourmand1376 do you know how to do hide the navbar properly? I tried to simply remove the `show` class from it, but it have 2 side effects:
> 1. it hides abruptly and without animation
> 2. it keeps the icon on the navbar as it were opened

Nope. I don't know.

sahiljhawar:
is it possible to do a site wide search instead of just he options which are shown in the search menu?

george-gca:
In situations like this it is recommended that you open a Q&A in discussions.

Yes, it is possible, but this search works locally, meaning you need to transfer a file with all the content that is going to be used for search. With only titles and descriptions, this file is small and the search is fast. Meanwhile if you give all the titles, descriptions, and contents, this file will increase in size, which will not only make your site heavier, but the search inside the file will be slower.

sahiljhawar:
@george-gca okay, I’ll keep this in mind. But for now I don’t mind search being a bit slower, since I want to use my site as personal instruction repo for future references. Can you tell me how to put all the contents on the search or link to discussion which talks about this. Thanks

george-gca:
Sure. Can you open a Q&A for this? This way it is easier for more people to find the answer to this.

asboyer:
Is there a way to make the search a little more specific? For example when searching "code", it returns results that contain "c" "o" "d" or "e" as well. There has to be a setting or field to restrict results to more exact matches?

george-gca:
This was intentional, that's why it is based on [deepdub's fork](https://github.com/deepdub-ai/ninja-keys) since it uses fuzzy search. If you want a more traditional search you can probably change the [score function](https://github.com/superhuman/command-score/?tab=readme-ov-file#behaviour) used, or change your code to use the original [Ninja keys](https://github.com/ssleptsov/ninja-keys). Note that I made a bunch of changes to [deepdub's code](https://github.com/deepdub-ai/ninja-keys) to make it work in our repo, so I don't think any of these changes are straightforward.

PR #2420, opened 2024-05-15
Fix offline leaflet

### LINKED ISSUE #2419: Leaflet issues in offline mode

### Bug description

Offline mode results in broken integrity hashes for Leaflet lib includes and graphics are missing. 

### How to reproduce the bug

1. Enable offline mode in _config.yml:
    ```
    third_party_libraries:
      download: true
    ```
2. Build static page
3. Open some page with Leaflet map
-> integrity errors
4. Remove integrity hashes manually
-> missing image (marker) files

### Error messages and logs

```
Failed to find a valid digest in the 'integrity' attribute for resource '.../assets/libs/leaflet.css' with computed SHA-256 integrity 'yuviegZOf3HiZ8eBDv629tOtj/DL9pjhq9m/8qjYX4o='. The resource has been blocked.
teaching/:1 Failed to find a valid digest in the 'integrity' attribute for resource '.../assets/libs/leaflet.js' with computed SHA-256 integrity 'fh+o7yCaRelV5DCgj6OhsmjH6qYlpiwY101vi/Ijeew='. The resource has been blocked.
teaching/:144 Uncaught ReferenceError: L is not defined
    at teaching/:144:2453
    at NodeList.forEach (<anonymous>)
    at document.onreadystatechange (teaching/:144:2297)
```

It looks as if the files have actually been changed, at least in the first line.

After deactivating the check, some further problem with missing images (icons/markers) comes up:
I think just those from e.g. https://cdn.jsdelivr.net/npm/leaflet@1.9.4/dist/images/ need to be downloaded to /assets/libs/. Manually copying fixed this in my case.

Deployed site

### More info

Started request in #1181 and was asked to open new issue.

george-gca:
Fixes #2419 by downloading leaflet images. Also changed the path where the libraries are downloaded to avoid not downloading files with same name.

### DIFF
--- a/_includes/scripts/tikzjax.liquid
+++ b/_includes/scripts/tikzjax.liquid
@@ -1,3 +1,8 @@
 {% if page.tikzjax %}
-  <script defer src="https://tikzjax.com/v1/tikzjax.js"></script>
+  <script
+    defer
+    src="https://tikzjax.com/v1/tikzjax.js"
+    integrity="sha256-f4Ln8BLwo8WdQPYX6W/2l0/YvKVk+mgIVe32uEf6zU4="
+    crossorigin="anonymous"
+    ></script>
 {% endif %}
--- a/_includes/scripts/tikzjax.liquid
+++ b/_includes/scripts/tikzjax.liquid
@@ -4,5 +4,5 @@
     src="https://tikzjax.com/v1/tikzjax.js"
     integrity="sha256-f4Ln8BLwo8WdQPYX6W/2l0/YvKVk+mgIVe32uEf6zU4="
     crossorigin="anonymous"
-    ></script>
+  ></script>
 {% endif %}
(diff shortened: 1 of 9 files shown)


_includes/scripts/tikzjax.liquid
```diff
@@ -1,3 +1,8 @@
 {% if page.tikzjax %}
-  <script defer src="https://tikzjax.com/v1/tikzjax.js"></script>
+  <script
+    defer
+    src="https://tikzjax.com/v1/tikzjax.js"
```
CheariX, line 4: I'm wondering if this would not break offline mode?
The `src` seems to be hardcoded.

CheariX:
I'm wondering how the images can downloaded beforehand?
I assume that only the initial images are downloaded but when you scroll the map, there will be missing images?

_includes/scripts/tikzjax.liquid
george-gca: There's nothing much we can do about tikzjax. Converting it to be 100% offline would involve too much change to its original source code.

george-gca:
Yes, there is so much we can do about offline support.

Issue #2425, opened 2024-05-17
Add Back To Top button


simonwei97:
### Have you checked that your feature request isn't already filed?

- [X] I read through [FAQ](https://github.com/alshedivat/al-folio/blob/master/FAQ.md) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my feature request.
- [X] Yes, I have checked that this feature request isn't already filed.

### Description & Motivation

Add back to top button for post.

Some example:
- https://jeffreytse.github.io/jekyll-theme-yat/

![image](https://github.com/alshedivat/al-folio/assets/119845914/69883a5d-995e-4564-89a8-b55ac623f376)

### Pitch

when user browse the post content to any position except the top of the article, `back_to_top` button will appear in the lower left corner of the screen. If user touch the button, it will go back to the top of article.

### Additional context

I will contribute for this feature as possible as I can.

simonwei97:
@george-gca Thanks for your time to review my commit code. 

My PR has been merge into master #2427. I will close this issue.

george-gca:
@simonwei97 thanks for your contribution. Would you mind taking a look at the mobile version? The button is too close to the footer.

simonwei97:
> @simonwei97 thanks for your contribution. Would you mind taking a look at the mobile version? The button is too close to the footer.

OK. I will work on it. I was not considering to much for mobile device before.

- [ ] mobile version optimization for `back-to-top` feature

simonwei97:
> > @simonwei97 thanks for your contribution. Would you mind taking a look at the mobile version? The button is too close to the footer.
> 
> OK. I will work on it. I was not considering to much for mobile device before.
> 
> * [x]  mobile version optimization for `back-to-top` feature

@george-gca hi, #2433 have some optimzation, will you review @abhilesh codes? Thanks.

PR #2427, opened 2024-05-17
feat: add back-to-top button

simonwei97:
slove #2425 

Demo:

image


_includes/scripts/back_to_top.liquid
```diff
@@ -0,0 +1,8 @@
+<script src="{{ '/assets/js/vanilla-back-to-top.min.js' | relative_url | bust_file_cache }}"></script>
+<script>addBackToTop({
+  diameter: 56,
+  cornerOffset: 50, // px
+  backgroundColor: '#ddd',
+  textColor: '#000',
```
george-gca, line 6: I believe it would be better if we set some of these values to the ones already set in [_theme.scss](https://github.com/alshedivat/al-folio/blob/master/_sass/_themes.scss) (e.g. `--global-text-color`), or create new ones for these there. This way all styling can be done just by settings css properties.

george-gca (changes requested):

_includes/scripts/back_to_top.liquid
simonwei97: Ok, I added new variables for `back-to-top` button, pls review.

simonwei97:
## Light Theme

image

## Dark Theme

image

george-gca (approved):

simonwei97:
I sync latest code and update code formatter. 

Need run workflow again.

PR #2444, opened 2024-05-26
Delete extra space ; Update blog.md

furkanakkurt1335:


george-gca (changes requested):
Can you also do the same for `categories` that is right below?

furkanakkurt1335:
> Can you also do the same for `categories` that is right below?

I've done this. Thank you for the recommendation, good for extensibility.

george-gca (approved):

Issue #1961, opened 2023-11-28
Stick the sidebar of table of content (TOC) in the distill template.


luoqinpei:
Hello! This is really an amazing template but I have one thing that I want to figure out. I have just successfully activated the sidebar on the distill template, however, it sticks to the top of the page and when I scroll down the table is not shown on the left as in the post template. I really want to know how to stick it to be kept shown on the left as I have a very long article, thanks a lot!

george-gca:
It is a little bit tricky to make this work. I wasn't able yet to make it work, maybe someone with more hands-on experience in web dev could help us achieve this.

PR #2445, opened 2024-05-27
add sticky toc

eolesinski:
This PR adds a sticky toc to the Distill template as mentioned in #1961

Changes:
- adds sticky toc
- reduces separator line width (for better fit)

### DIFF
--- a/_sass/_distill.scss
+++ b/_sass/_distill.scss
@@ -19,6 +19,7 @@ d-article d-byline a {
 }
 
 d-article {
+  overflow-x: visible !important;
   border-top-color: var(--global-divider-color) !important;
   p,
   h1,
@@ -74,6 +75,10 @@ d-article {
     }
   }
 
+  hr {
+    width: 46%;
+  }
+
   d-contents {
     align-self: start;
     grid-column: 1 / 4;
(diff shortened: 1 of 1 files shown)


alshedivat:
thanks for adding this feature. the only issue I noticed is that transparent background of the sticky toc doesn't look good when the toc bloc starts overlapping with other objects on the page. also, this won't work on mobile well. could you think of a way to fix these issues?

|Screenshot 2024-05-27 at 1 01 07 PM|Screenshot 2024-05-27 at 1 00 27 PM|
|---|---|

alshedivat (changes requested):
see above comment about the stylistic issues. these need to be fixed before we can merge this PR. thank you.

eolesinski:
@alshedivat great point - thanks for the feedback. I'll take a look.

george-gca:
How did you envision this collapsible toc in this case? Where and how it would appear on mobile? I tend to think that just hiding the toc would be a good use, since the mobile screen is too small to fit both the toc and the contents.

hrhoo:
I had a look at the changes.

How about adding this line.
image

When the window is small, it makes the toc fixed

saguileran:
Is there any possible and easy solution?

george-gca:
@hrhoo can you test if your solution works, and if so, send a PR with this?

PR #2447, opened 2024-05-27
feat: search.liquid over all collections

CheariX:
Thank you @george-gca for the awesome work. on #2415.

This PR generalizes the search on all collections. Currently, only projects are added to the search.
This PR uses all of them, such as news. On my personal website, I use a teaching collection which is then also automatically searched.

### DIFF
--- a/_includes/scripts/search.liquid
+++ b/_includes/scripts/search.liquid
@@ -89,7 +89,7 @@
           {
             {%- assign title = item.title | escape -%}
             id: "{{ collection.label }}-{{ title | slugify }}",
-            title: "{{ title}}",
+            title: "{{ title }}",
             description: "{{ item.description | strip_html | strip_newlines | escape }}",
             section: "{{ collection.label | capitalize }}",
             handler: () => {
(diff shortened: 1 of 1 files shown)


george-gca (changes requested):
Could you add spaces between the variables and the `{{ }}`, so they all look like `{{ var }}` instead of `{{var }}` or `{{ var}}`? Nice addition btw.

CheariX:
Good catch :-)

_includes/scripts/search.liquid
```diff
@@ -84,18 +84,20 @@
...
-          description: "{{ project.description | strip_html | strip_newlines | escape }}",
-          section: "Projects",
-          handler: () => {
-            window.location.href = "{{ project.url | relative_url }}";
+      {% for collection in site.collections %}
+        {%- for item in collection.docs -%}
+          {
+            {%- assign title = item.title | escape -%}
+            id: "{{ collection.label }}-{{ title | slugify }}",
+            title: "{{ title}}",
```
george-gca, line 92: Missed the space here.

george-gca (changes requested):

_includes/scripts/search.liquid
CheariX: Thanks. Fixed.

george-gca (approved):

PR #2455, opened 2024-05-28
Fix: date pill position on CV

abhilesh:
Fixes: #2393 

Changes made in this PR - 
Added `style="width: 75px; transform: translateX(-15px) translateY(-5px);">` to move the date pill `15px` to the left and `5px` to the top

| Before | After |
| :-----: | :----: |
| date_pill_before | date_pill_after |


george-gca (changes requested):
Can you move all those csses from `style` attributes to `_sass/_base.scss`? Not sure if this would cause too much trouble, but this way one can change the whole behavior just by messing with css and not into the liquid files.

CheariX:
I'm a bit curious... Moving the pills "some" pixels left looks like a work around to me. 
Does anyone have an idea what is the root cause of this issue?

abhilesh (commented):
Added a new class `date-column` to the liquid files and moved the styling options to `_sass/_base.scss`. I didn't notice any changes elsewhere.

george-gca (approved):

Issue #2437, opened 2024-05-24
Change search text in top bar if on Mac OS


george-gca:
### Have you checked that your feature request isn't already filed?

- [X] I read through [FAQ](https://github.com/alshedivat/al-folio/blob/master/FAQ.md) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my feature request.
- [X] Yes, I have checked that this feature request isn't already filed.

### Description & Motivation

Since the hotkeys for search in Mac OS are `⌘k` instead of `ctrl k`, this can be a little misleading for the user.

### Pitch

Change the text next to the search button depending on the platform. For example, if the user is accessing the site on a mac they should use `⌘k` instead of `ctrl k`

PR #2461, opened 2024-05-29
Feature: Dynamically sets the search shortcut key based on the user's platform

ajyey:
This addresses issue #2437 by:
- Adding a new script that dynamically sets the search keyboard shortcut by checking what platform the user is currently using
- Loading this script in `default.liquid`

SCR-20240529-cdfe

### DIFF
--- a/_layouts/default.liquid
+++ b/_layouts/default.liquid
@@ -13,7 +13,6 @@
       <meta http-equiv="refresh" content="3; url={{ redirect }}">
     {% endif %}
     {% include head.liquid %}
-    <script src="{{ '/assets/js/shortcut-key.js' | relative_url }}"></script>
   </head>
 
   <!-- Body -->
@@ -74,5 +73,6 @@
     {% include scripts/jekyll_tabs.liquid %}
     {% include scripts/back_to_top.liquid %}
     {% include scripts/search.liquid %}
+    <script src="{{ '/assets/js/shortcut-key.js' | relative_url }}"></script>
   </body>
 </html>
--- a/_layouts/default.liquid
+++ b/_layouts/default.liquid
@@ -13,6 +13,7 @@
       <meta http-equiv="refresh" content="3; url={{ redirect }}">
     {% endif %}
     {% include head.liquid %}
+    <script src="{{ '/assets/js/shortcut-key.js' | relative_url }}"></script>
   </head>
 
   <!-- Body -->
--- a/assets/js/shortcut-key.js
+++ b/assets/js/shortcut-key.js
@@ -1,6 +1,6 @@
 // Check if the user is on a Mac and update the shortcut key for search accordingly
 (function () {
-  var isMac = navigator.platform.toUpperCase().indexOf('MAC') >= 0;
+  let isMac = navigator.platform.toUpperCase().indexOf('MAC') >= 0;
   document.addEventListener('readystatechange', function () {
     if (document.readyState === 'interactive') {
       let shortcutKeyElement = document.querySelector('#search-toggle .nav-link');
--- a/assets/js/shortcut-key.js
+++ b/assets/js/shortcut-key.js
@@ -1,0 +1,12 @@
+// Check if the user is on a Mac and update the shortcut key for search accordingly
+
+var isMac = navigator.platform.toUpperCase().indexOf('MAC') >= 0;
+document.addEventListener('readystatechange', function () {
+  if (document.readyState === 'interactive') {
+    var shortcutKeyElement = document.querySelector('#search-toggle .nav-link');
+    if (shortcutKeyElement && isMac) {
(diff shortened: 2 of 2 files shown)


assets/js/shortcut-key.js
```diff
@@ -0,0 +1,12 @@
+// Check if the user is on a Mac and update the shortcut key for search accordingly
+
+var isMac = navigator.platform.toUpperCase().indexOf('MAC') >= 0;
```
CheariX, line 3: Maybe we should hange this line to not pollute the global `window` scope?
E.g. by moving the whole code into a function that is automatically called?

assets/js/shortcut-key.js
```diff
@@ -0,0 +1,12 @@
...
+document.addEventListener('readystatechange', function () {
+  if (document.readyState === 'interactive') {
+    var shortcutKeyElement = document.querySelector('#search-toggle .nav-link');
```
CheariX, line 6:
```suggestion
    let shortcutKeyElement = document.querySelector('#search-toggle .nav-link');
```

assets/js/shortcut-key.js
ajyey: Updated. Let me know if you had something different in mind @CheariX

assets/js/shortcut-key.js
ajyey: Fixed 👍

ajyey:
@CheariX thinking about it a bit more, would it make sense to also use the uncode for `ctrl` to keep it consistent across different platforms or are we ok with the way we have it here?

assets/js/shortcut-key.js
```diff
@@ -0,0 +1,13 @@
+// Check if the user is on a Mac and update the shortcut key for search accordingly
+(function () {
+  var isMac = navigator.platform.toUpperCase().indexOf('MAC') >= 0;
```
CheariX, line 3:
```suggestion
  let isMac = navigator.platform.toUpperCase().indexOf('MAC') >= 0;
```
I think, this could also be let instead of var.

CheariX:
> @CheariX thinking about it a bit more, would it make sense to also use the unicode for `ctrl` to keep it consistent across different platforms or are we ok with the way we have it here?

Hm. Good question! Is there a unicode for `ctrl`? I think on Mac, you sometimes use `⌃` instead of the string `ctrl`.
But on Windows/Linux, `ctrl` is very common. 
I don't have a preference on that.

ajyey:
> > @CheariX thinking about it a bit more, would it make sense to also use the unicode for `ctrl` to keep it consistent across different platforms or are we ok with the way we have it here?
> 
> 
> 
> Hm. Good question! Is there a unicode for `ctrl`? I think on Mac, you sometimes use `⌃` instead of the string `ctrl`.
> 
> But on Windows/Linux, `ctrl` is very common. 
> 
> I don't have a preference on that.

Understood. I'll leave it as is then

_layouts/default.liquid
```diff
@@ -13,6 +13,7 @@
       <meta http-equiv="refresh" content="3; url={{ redirect }}">
     {% endif %}
     {% include head.liquid %}
+    <script src="{{ '/assets/js/shortcut-key.js' | relative_url }}"></script>
```
george-gca, line 16: Does it affect if this script is added to the end of the body with the other scripts?

assets/js/shortcut-key.js
```diff
@@ -0,0 +1,13 @@
+// Check if the user is on a Mac and update the shortcut key for search accordingly
+(function () {
+  let isMac = navigator.platform.toUpperCase().indexOf('MAC') >= 0;
+  document.addEventListener('readystatechange', function () {
+    if (document.readyState === 'interactive') {
```
george-gca, line 5: Is there any difference between doing this way and doing like this? Just out of curiosity:
```javascript
document.onreadystatechange = () => {
  if (document.readyState === 'complete') {
```

_layouts/default.liquid
ajyey: Nope ! Good idea. Moved it to the bottom of the body.

assets/js/shortcut-key.js
ajyey: Yes, if we wait until the document is fully loaded, you can consistently see the search change from `ctrl k` to `cmd k` in the navbar. Using `interactive` gives us access to manipulate the dom elements earlier in the document loading process as stated [here](https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState)

Its not absolutely perfect.. meaning if you refresh over and over again you'll catch a glimpse of it changing maybe 1 out of 10 times, but its better than waiting until the readystate is complete

I think the way you have it though is cleaner. Updated 👍

george-gca (approved):

PR #2463, opened 2024-05-30
Bugfix: Updates ninja keys text input color so it is always visible

ajyey:
This PR fixes an issue where the search input is not visible in the light theme. This is because `ninja-header-min.js` defaults the text color to white. 

This style change will ensure that the text is always visible regardless of theme selection

## Before Style Change
SCR-20240530-cnxd

## After Style Change
SCR-20240530-coqb

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ Gemfile.lock
 assets/libs/
 node_modules/
 vendor
+.idea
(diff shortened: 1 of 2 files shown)


george-gca (approved):

george-gca:
Thanks for noticing this.

PR #2480, opened 2024-06-05
[Tweak] Update "search filters" displayed on the blog's front page

rstein66:
# [Tweak] Update "search filters" on blog's front page

## Current Behavior

```
1. Go to `blog/`
2. Select "🏷️ Blockquotes" "search filter"

=> `blog/category/blockquotes/` returns 404
```

current-01-blog-filters
current-02-category-blockquotes

<hr>

## Resolution in PR

1. Append 'blockquotes' to [`display_tags`](https://github.com/alshedivat/al-folio/pull/2480/files#diff-ecec67b0e1d7e17a83587c6d27b6baaaa133f42482b07bd3685c77f34b62d883R295)  
2. Replace 'blockquotes' with 'external-services' in `display_categories`  

=> Display 'blockquotes' tag and 'external-services' category on blog's front page

v2-01
<br>  
v2-02
v2-03

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -292,8 +292,8 @@ jekyll-archives:
     tag: "/blog/tag/:name/"
     category: "/blog/category/:name/"
 
-display_tags: ["formatting", "images", "links", "math", "code"] # these tags will be displayed on the front page of your blog
-display_categories: ["blockquotes"] # these categories will be displayed on the front page of your blog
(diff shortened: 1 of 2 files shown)


george-gca (changes requested):
I think it makes more sense to keep `blockquotes` as tag, and just change the [display_categories](https://github.com/rstein66/al-folio/blob/tweak/add-blog-category/_config.yml#L295) for `external-services`, since it is the only other category we have other than `sample-posts`.

george-gca (approved):

Issue #923, opened 2022-10-06
digital bookshelf for al-folio


subhodeeps:
Hi,
I would like to request a new feature.

**Is your feature request related to a problem? Please describe.**

This is feature request to add a `digital bookshelf` to al-folio. Many of us, particularly academics and students, read a lot of books and it would be nice if we could display what we are reading on our website and link them to reviews that we may write for them. Think of it as your own personal (stripped-down) Goodreads running on your own personal website.

**Describe the solution you'd like**

I have been recently toying with this idea and created [jekyll-bookworm](https://github.com/subhodeeps/jekyll-bookworm) for the Minima theme. You can see what it looks like [here](https://subhodeeps.github.io/jekyll-bookworm/). Since I am thinking of shifting my academic website to al-folio from  the (the now abandoned) academicpages theme, I decided to see if I can implement it in al-folio. I have been modestly successful. You can take a look at my amatuer tinkering [here](https://subhodeeps.github.io/al-folio/_pages/books/).

I felt other users of al-folio might like this feature and I would be happy to send a pull request (you can, of course, take this idea ahead on your own).

There are however some issues that needs improvement:

- the reviews are not displayed properly on phones and I can't seem to fix it, in general it can do with a lot aesthetic tweaks
- it would be nice to have pagination in the bookshelf but I don't have the required expertise to do it quickly

**Describe alternatives you've considered**

I could not find an alternative or existing solution to what I was looking for: a digital bookshelf and a book review management system. I would be happy if you can point me to something that can implement this in a more polished fashion.

**Additional context**

I am not an expert in web development and I admit my attempts are probably amateurish but I would be happy to get your inputs and see if something like this can be incorporated into al-folio.

Best,
Subhodeep

asboyer:
@subhodeeps I took a look at your personal website and really liked what you've done with your bookshelf. I made an [attempt](https://www.asboyer.com/favs/books/) to do what you were talking about on this al-folio template. I actually prefer the look of yours, as well the different statuses for each book you're reading.

george-gca:
If you manage to implement this, would you mind sending a PR? Maybe you could also add the subscribe via email that I saw you have in your site to this template? It is a feature that have been requested a while ago (#2097).

asboyer:
Yeah, I can do that. I'll make a PR when I've got those.

subhodeeps:
[...]

PR #2517, opened 2024-06-18
Added support for a newsletter

asboyer:
In reference to idea: https://github.com/alshedivat/al-folio/discussions/2097
In reference to request: https://github.com/alshedivat/al-folio/issues/923#issuecomment-2171924663

Added support to integrate a [loops.so](https://loops.so/) mailing list into the site.

To use, you need to enable `newsletter` in `_config.yml`. You also must specify a loops endpoint (although I think any mailing list endpoint can work), which you can get when you set up a mailing list on loops. More documentation on loops: [here](https://loops.so/docs/forms/custom-form).

Once that is enabled, the behavior is different depending on how you specified your footer to behave in `_config.yml`. If `footer_fixed: true`, then the sign up will appear at the bottom of the about page, as well as at the bottom of blog posts, if you enable `related_posts`.

If `footer_fixed: false`, then the newsletter signup will be in the footer (on every page), like it is in on [my website](https://asboyer.com).

I'm not attached to the placement of the signup, and you can choose to include it wherever you want with `{% include scripts/newsletter.liquid %}`. Also if you include positional variables into that, you can choose how you center the signup. So `{% include scripts/newsletter.liquid left=true %}` positions the signup bar to the left.

Here are some screenshots below:
## Dark version
image

## Light version
image
I think the input field color should probably change to maybe be light for both themes? What do you think? I think the dark background looks cool, but I don't usually see that done like that on other sites.

## Footer fixed
image

image

## Footer not fixed
image

image
To clarify, if footer isn't fixed, the email signup will appear on every page.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -175,6 +175,15 @@ external_sources:
         published_date: 2024-05-14
 
 # -----------------------------------------------------------------------------
+# Newsletter
+# -----------------------------------------------------------------------------
+
+newsletter:
+  enabled: false
+  endpoint: # your loops endpoint (e.g., https://app.loops.so/api/newsletter-form/YOUR-ENDPOINT)
+  # https://loops.so/docs/forms/custom-form
+
+# -----------------------------------------------------------------------------
 # Collections
 # -----------------------------------------------------------------------------
 
--- a/_includes/footer.liquid
+++ b/_includes/footer.liquid
@@ -15,6 +15,10 @@
   </footer>
 {% else %}
   <footer class="sticky-bottom mt-5" role="contentinfo">
+    {% if site.newsletter.enabled %}
+      {% include scripts/newsletter.liquid %}
+    {% endif %}
+
     <div class="container">
       &copy; Copyright {{ site.time | date: '%Y' }}
       {{ site.first_name }}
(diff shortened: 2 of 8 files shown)


asboyer:
Also, can someone mark this as an enhancement/feature? I don't think I have the ability to tag this PR as anything.

george-gca:
This looks really nice. As for the background of the light version, I believe it is best to set to white for readability, or maybe a light gray as long at it doesn't look like a disabled element.

asboyer:
Okay, this is how it looks like in light mode now:

#### Dormant
image

#### Active
image

george-gca (approved):

george-gca:
@asboyer I merged your commit, but just now realized that we forgot about adding this to our README. Would you mind sending a new PR with an explanation added to [CUSTOMIZE.md](https://github.com/alshedivat/al-folio/blob/master/CUSTOMIZE.md)?

asboyer:
> @asboyer I merged your commit, but just now realized that we forgot about adding this to our README. Would you mind sending a new PR with an explanation added to [CUSTOMIZE.md](https://github.com/alshedivat/al-folio/blob/master/CUSTOMIZE.md)?

Yes, I'll whip that up now.

PR #2520, opened 2024-06-19
Add example use of annotation and superscripts in bibtex

garywei944:
image
image
image
image


george-gca (approved):

george-gca:
Better yet. Thanks for your contribution.

PR #2521, opened 2024-06-19
Update CUSTOMIZE.md for Newsletter support

asboyer:
In reference to https://github.com/alshedivat/al-folio/pull/2517 and https://github.com/alshedivat/al-folio/pull/2517#issuecomment-2179244937

### DIFF
--- a/CUSTOMIZE.md
+++ b/CUSTOMIZE.md
@@ -146,3 +146,9 @@ A variety of beautiful theme colors have been selected for you to choose from. T
 ## Adding social media information
 
 You can add your social media links by adding the specified information at the `Social integration` section in the [\_config.yml](_config.yml) file. This information will appear at the bottom of the `About` page.
+
+## Adding a newsletter
+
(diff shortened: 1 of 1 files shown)


CUSTOMIZE.md
```diff
@@ -146,3 +146,9 @@ A variety of beautiful theme colors have been selected for you to choose from. T
 ## Adding social media information
 
 You can add your social media links by adding the specified information at the `Social integration` section in the [\_config.yml](_config.yml) file. This information will appear at the bottom of the `About` page.
+
+## Adding a newsletter
+
+You can add a newsletter subscription form by adding the specified information at the `newsletter` section in the [\_config.yml](_config.yml) file. To set up a newsletter, you can use a service like [Loops.so](https://loops.so/). Once you have set up your newsletter, you can add the form [endpoint](https://loops.so/docs/forms/custom-form) to the `endpoint` field in the `newsletter` section of the [\_config.yml](_config.yml) file.
```
george-gca, line 152: I believe it is best to make it clear that we use only support loops for now. Something like:

```markdown
To set up a newsletter, you can use a service like [Loops.so](https://loops.so/), which is the current supported solution.
```

george-gca (changes requested):

CUSTOMIZE.md
asboyer: Good point

george-gca (approved):

PR #2522, opened 2024-06-19
Fix Altmetric badge not correctly set when Altmetric id is provided

garywei944:
To reproduce the bug:

```bibtex
@inproceedings{Vaswani2017AttentionIA,
  title     = {Attention is All you Need},
  author    = {Ashish Vaswani and Noam M. Shazeer and Niki Parmar and Jakob Uszkoreit and Llion Jones and Aidan N. Gomez and Lukasz Kaiser and Illia Polosukhin},
  booktitle = {Neural Information Processing Systems},
  year      = {2017},
  doi       = {10.48550/arXiv.1706.03762},
  altmetric = {21021191}
}
```

The bug is
1. It seems to be some weird property of the liquid template that [line 252-254](https://github.com/alshedivat/al-folio/blob/8d82670ff170f98e7d7ea5434428234a8216d460/_layouts/bib.liquid#L252-L254) doesn't work at all. According to [this post](https://stackoverflow.com/questions/59887447/liquid-how-to-assign-the-output-of-an-operator-to-a-variable) and [this issue](https://github.com/Shopify/liquid/issues/236), liquid doesn't support assign the output of operator to a variable nor a ternary operator. So based on my console log, the value of `entry_has_altmetric_badge` is always a string value of `entry.altmetric` when altmetric is provided in bibtex.
```liquid
{% assign entry_has_altmetric_badge = entry.altmetric or entry.doi or  entry.eprint or entry.pmid or entry.isbn %}
{% assign entry_has_dimensions_badge = entry.dimensions or entry.doi or entry.pmid %}
{% assign entry_has_google_scholar_badge = entry.google_scholar_id %}
{% if entry_has_altmetric_badge or entry_has_dimensions_badge or entry_has_google_scholar_badge %}
  <div class="badges">
  {% if site.enable_publication_badges.altmetric and entry_has_altmetric_badge %}
    <span
...
```
Note that this could be problematic that a string in liquid is always evaluated as true as long as it is defined regardless if it is "" or "false". [reference](https://shopify.github.io/liquid/basics/truthy-and-falsy/)
2. when altmetric is defined in bibtex, now the order of set attribute to badge is eprint > doi > altmetric id > pmid > ISBN, and the badge doesn't work when an arxiv doi is provided.

I think the expected behavior should be
1. as documented in CUSTOMIZE.md, only render the badge when the entry is set to either "true" or the altmetric id. (It could also implement to always render the badge whenever doi or other related attribute is set, and set altmetric to "false" to disable it)
```md
- `altmetric`: Adds an [Altmetric](https://www.altmetric.com/) badge (Note: if DOI is provided just use `true`, otherwise only add the altmetric identifier here - the link is generated automatically)
```
2. if the almetric id is set, use it first.

### DIFF
--- a/_layouts/bib.liquid
+++ b/_layouts/bib.liquid
@@ -250,17 +250,17 @@
     </div>
     {% if site.enable_publication_badges %}
       {% assign entry_has_altmetric_badge = false %}
-      {% if entry.altmetric != blank and entry.altmetric != 'false' %}
+      {% if entry.altmetric and entry.altmetric != 'false' %}
         {% assign entry_has_altmetric_badge = true %}
       {% endif %}
 
       {% assign entry_has_dimensions_badge = false %}
-      {% if entry.dimensions != blank and entry.dimensions != 'false' %}
+      {% if entry.dimensions and entry.dimensions != 'false' %}
         {% assign entry_has_dimensions_badge = true %}
       {% endif %}
 
       {% assign entry_has_google_scholar_badge = false %}
-      {% if entry.google_scholar_id != blank %}
+      {% if entry.google_scholar_id %}
         {% assign entry_has_google_scholar_badge = true %}
       {% endif %}
       {% if entry_has_altmetric_badge or entry_has_dimensions_badge or entry_has_google_scholar_badge %}
--- a/_layouts/bib.liquid
+++ b/_layouts/bib.liquid
@@ -59,8 +59,8 @@
 
       {%- for author in entry.author_array limit: author_array_limit -%}
         {% assign author_is_self = false %}
-        {%- assign author_last_name = author.last | regex_replace: '[*†‡§¶‖&^]', '' -%}
-        {%- assign author_last_html = author.last | regex_replace: '([*†‡§¶‖&^]+)', '<sup>\1</sup>' -%}
+        {%- assign author_last_name = author.last | regex_replace: '[*∗†‡§¶‖&^]', '' -%}
+        {%- assign author_last_html = author.last | regex_replace: '([*∗†‡§¶‖&^]+)', '<sup>\1</sup>' -%}
         {% if site.scholar.last_name contains author_last_name %}
           {% if site.scholar.first_name contains author.first %}
             {% assign author_is_self = true %}
@@ -112,7 +112,7 @@
             {% assign more_authors_show = more_authors_show | append: ', ' %}
           {% endunless %}
         {%- endfor -%}
-        {% assign more_authors_show = more_authors_show | regex_replace: '([*†‡§¶‖&^]+)', '<sup>\1</sup>' %}
+        {% assign more_authors_show = more_authors_show | regex_replace: '([*∗†‡§¶‖&^]+)', '<sup>\1</sup>' %}
         , and
         <span
           class="more-authors"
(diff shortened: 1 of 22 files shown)


_layouts/bib.liquid
```diff
@@ -261,12 +272,14 @@
               data-hide-less-than="15"
               data-badge-type="2"
               data-badge-popover="right"
-              {% if entry.eprint %}
+              {% if entry.altmetric != blank and entry.altmetric != 'true' %}
```
george-gca, line 275: Shouldn't it be `! = false` here?

_layouts/bib.liquid
```diff
@@ -277,12 +290,12 @@
           {% if site.enable_publication_badges.dimensions and entry_has_dimensions_badge %}
             <span
               class="__dimensions_badge_embed__"
-              {% if entry.doi %}
+              {% if entry.dimensions != blank and entry.dimensions != 'true' %}
```
george-gca, line 293: Shouldn't it be `! = false` here?

_layouts/bib.liquid
```diff
@@ -261,12 +272,14 @@
               data-hide-less-than="15"
               data-badge-type="2"
               data-badge-popover="right"
-              {% if entry.eprint %}
+              {% if entry.altmetric != blank and entry.altmetric != 'true' %}
+                data-altmetric-id="{{ entry.altmetric }}"
+              {% elsif entry.arxiv %}
```
george-gca, line 277: Since we are changing to also check for blank for that field, shouldn't we change it for all of them? What do you think?

george-gca (changes requested):

_layouts/bib.liquid
garywei944: Not really, I think the expected values here are `"true"`, `"false"`, and any altmetric id string. `"true"` means that the user wants to enable the badge. We only want to enter this condition check when the value of the altmetric entry is set to an id, not set to `"true"`.

_layouts/bib.liquid
garywei944: ^ Same as above

_layouts/bib.liquid
garywei944: I think you are right. The check for blank string is not necessary. I previously thought that `entry.altmetic` will be set to an empty string whenever it is provided in the BibTex. A new commit fixed this.

_layouts/bib.liquid
garywei944: Oh if altmetric is `"false"`, then `entry_has_altmetric_badge` won't be set to `true` in [line 254](https://github.com/alshedivat/al-folio/pull/2522/commits/d8917e7836389f21ed471b9f175fafd913ad148b#diff-4ef4ea42cd30776daef2d2cc2da3e79af05331b84aa21bc8e7eea019b935d701R253-R254) and the code wont pass the condition check in [line 268](https://github.com/alshedivat/al-folio/pull/2522/commits/d8917e7836389f21ed471b9f175fafd913ad148b#diff-4ef4ea42cd30776daef2d2cc2da3e79af05331b84aa21bc8e7eea019b935d701R268). So we don't need to worry about "false" in [line 275](https://github.com/alshedivat/al-folio/pull/2522/commits/d8917e7836389f21ed471b9f175fafd913ad148b#diff-4ef4ea42cd30776daef2d2cc2da3e79af05331b84aa21bc8e7eea019b935d701R275)

george-gca (approved):

PR #741, opened 2022-07-04
Update bootstrap and other library dependencies

rohandebsarkar:
This PR is an attempt to move al-folio from Bootstrap v4 to v5. Bootstrap v5 introduced a number of breaking changes, most of them are listed here: https://getbootstrap.com/docs/5.0/migration/.

I have tried to solve most of the breaking changes, but there might be many more that went unnoticed.

Along with bootstrap, I have also updated the other library dependencies to major versions only. This will help in less maintenance of the library versions. And therefore also had to remove the integrity checks. (Also, since we are using a constant major version, there won't be any breaking changes according to [SemVer](https://semver.org/))

I have deployed this branch on [Vercel](https://vercel.com/home) for monitoring any abnormalities. This branch will be continuously deployed on this URL: https://alfolio.vercel.app/

PR #2523, opened 2024-06-20
feat: simple filtering / searching on bibliography

CheariX:
This PR adds a simple filter/search functionality to the bibliography.

It can be used in two ways:

1. Simply enter a search term in the input box.
2. Send a search term via the `location.hash`, e.g., https://alshedivat.github.io/al-folio/publications/#mechanics

**Notes:**

- The search box is optional. It can be simply removed if anyone does not like it.
- Searching via `hash` works without the search box. My idea is to use this functionality to index all BibTeX entries via the `ctrl-k` search and link them via their BibTeX key.
- Searching via `hash` could also be used to set static links on the current page, e.g., to filter specific co-authors, venues, etc.
- I don't know much about the design of the input field. I simply reused the newsletter box style.
- Entering a search term in the box does exact matching. No fuzzy search, no AND/OR logic. I kept it very simple. Maybe anyone else wants to improve it in the future.
- The search looks in all data in the BibTeX entry that is parsed via `bib.liquid`. E.g., it is possible to search for BibTeX keys, titles, authors, years, venues, abstracts, or whatever `bib.liquid` prints.
- I used a 300ms delay before starting to search on the input box.
- Entering search terms in the box does not update the location hash (things could get complex otherwise due to automatically updating each other...)
- If the filter does not find any match in a specific year, the year is also made invisible.

**Screenshot**
screenshot

Looking for feedback.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -51,6 +51,7 @@ navbar_fixed: true
 footer_fixed: true
 search_enabled: true
 socials_in_search: true
+bib_search: true
 
 # Dimensions
 max_width: 930px
--- a/_includes/scripts/misc.liquid
+++ b/_includes/scripts/misc.liquid
@@ -36,6 +36,7 @@
 <script src="{{ '/assets/js/no_defer.js' | relative_url | bust_file_cache }}"></script>
 <script defer src="{{ '/assets/js/common.js' | relative_url | bust_file_cache }}"></script>
 <script defer src="{{ '/assets/js/copy_code.js' | relative_url | bust_file_cache }}" type="text/javascript"></script>
+<script defer src="{{ '/assets/js/bibsearch.js' | relative_url | bust_file_cache }}" type="text/javascript"></script>
 
 <!-- Jupyter Open External Links New Tab -->
 <script defer src="{{ '/assets/js/jupyter_new_tab.js' | relative_url | bust_file_cache }}"></script>
--- a/_includes/scripts/misc.liquid
+++ b/_includes/scripts/misc.liquid
@@ -38,7 +38,9 @@
 <script defer src="{{ '/assets/js/copy_code.js' | relative_url | bust_file_cache }}" type="text/javascript"></script>
 
 <!-- Bibsearch Feature -->
-<script src="{{ '/assets/js/bibsearch.js' | relative_url | bust_file_cache }}" type="module"></script>
+{% if site.search_enabled %}
+  <script src="{{ '/assets/js/bibsearch.js' | relative_url | bust_file_cache }}" type="module"></script>
+{% endif %}
 
 <!-- Jupyter Open External Links New Tab -->
 <script defer src="{{ '/assets/js/jupyter_new_tab.js' | relative_url | bust_file_cache }}"></script>
--- a/_includes/scripts/misc.liquid
+++ b/_includes/scripts/misc.liquid
@@ -36,7 +36,18 @@
 <script src="{{ '/assets/js/no_defer.js' | relative_url | bust_file_cache }}"></script>
 <script defer src="{{ '/assets/js/common.js' | relative_url | bust_file_cache }}"></script>
 <script defer src="{{ '/assets/js/copy_code.js' | relative_url | bust_file_cache }}" type="text/javascript"></script>
+
+<!-- Bibsearch Feature -->
 <script defer src="{{ '/assets/js/bibsearch.js' | relative_url | bust_file_cache }}" type="text/javascript"></script>
+<link
+  rel="modulepreload"
+  href="{{ site.third_party_libraries.highlight-search-term.url.js }}"
+  integrity="{{ site.third_party_libraries.highlight-search-term.itegrity.js }}"
+>
+<script type="module">
+  import { highlightSearchTerm } from '{{ site.third_party_libraries.highlight-search-term.url.js }}';
+  window.highlightSearchTerm = highlightSearchTerm; // export to global scope
+</script>
 
 <!-- Jupyter Open External Links New Tab -->
 <script defer src="{{ '/assets/js/jupyter_new_tab.js' | relative_url | bust_file_cache }}"></script>
--- a/_includes/scripts/misc.liquid
+++ b/_includes/scripts/misc.liquid
@@ -38,16 +38,7 @@
 <script defer src="{{ '/assets/js/copy_code.js' | relative_url | bust_file_cache }}" type="text/javascript"></script>
 
 <!-- Bibsearch Feature -->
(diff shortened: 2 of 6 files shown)


george-gca:
I will try to test it this weekend. Question:

> I used a 300ms delay before starting to search on the input box.

Why?

george-gca:
It would be nice to highlight the search term found, so it would be clear why that item was selected. Something like the current search does:

image

CheariX:
> I will try to test it this weekend. Question:
> 
> > I used a 300ms delay before starting to search on the input box.
> 
> Why?

In contrast to the Ctrl-k search, this one is a bit "hacky". It operates on the HTML DOM rather than on a pretty JSON structure. I was a bit scared that the search could slow down the browser while typing. Waiting 300ms before starting to search reduces unnecessary workload. Without that, the search would be triggered with every key up while typing the search term. E.g, typing "mechan" would trigger the filtering 6 times. With normal typing speed below 300ms, one one filtering is triggered.

300ms seems to be fast enough for me to not get bored. I do not even feel this delay.

I'm open to change this if anyone thinks it is unnecessary.

> It would be nice to highlight the search term found, so it would be clear why that item was selected. Something like the current search does:
> 
> image

This is something that I would also like to have. 

Unfortunately, I have no idea how to do this. Help is appreciated here. Or we leave it for a future PR.

Also, the filter currently searches in "hidden" texts like abstract or bibtex. Should they be opened automatically?

CheariX:
image

I found a very cool library that was easy to integrate and does exactly what we are looking for (ref: [highlight-search-term](https://github.com/marmelab/highlight-search-term)).

Pro:
- Easy integration
- Does not use any clunky `<mark>` or `<span>` elements for highlighting => Does not pollute the DOM.

Con:
- One more library/dependency. Do we want that?
- I think the integration could be improved. My code works, but could be improved. E.g., I import the `highlight-search-term` ES module and then export the `highlightSearchTerm`-function to the global scope. However, my knowledge about doing this the right was is limited. 
- ES Modules cannot be easily imported in our `*.js` files. It would be nice to import the `highlightSearchTerm` directly into `bibsearch.js`. To achive that, I can only use hardcoded URLs (no liquid syntax), so that it would not work with `offline/download` mode. Could maybe be solved by downloading the file directly and putting it into the `assets` folder.
- Currently not [supported in Firefox](https://caniuse.com/?search=CSS.Highlights), but it will get support soon. So, this is not an issue IMHO.

What do you think?

CheariX:
Update.

I integrated `highlight-search-term.js` directly. Now, it can be easily integrated as a module.
The code looks much cleaner. 

Also, I set the old non-highlighting filter function as a fallback in case the browser does not support the API.

I'm pretty happy (=done) with the code now. 

One question left: Can we somehow enable ES6 support in Uglifier (that is the reason why the checks did not pass)? Or do we really want to stay with old ES5 syntax?

george-gca:
I will try to take a proper look at this this weekend.

CheariX:
Thank you, @george-gca. Looking forward to hear your opinion on this PR.

About the failed check:

We can change `node.textContent?.XYZ` (ES6 Syntax) to something like `node.textContent && node.textContent.XYZ` to pass uglify checks, see e.g. https://github.com/alshedivat/al-folio/pull/2523/commits/a0d51f708c4c3eadc252d6a41a872bd06abd939b. However, I’d argue that it would be nice to allow this syntax because it is much better to read.

CheariX:
I changed ES6 syntax to ES5 compatible syntax so that Uglifier does not fail.

In the future, it would be nice to allow ES6 Syntax, see #2548.

@george-gca: is there anything else I can do to push this forward?

george-gca:
I just need to make some time to look throught this PR. I'll do it this weekend for sure.

george-gca:
Sorry for the delay on this. This is a very nice addition indeed.

> Can we somehow enable ES6 support in Uglifier?

I think so. Some code updates might be needed, but it would be nice to keep it as up-to-date as possible with the latest best practices. There was also a huge effort (#741) to update some of the dependencies of the template (aka bootstrap), but that was never finished. I hope someday I will have the time to tackle this before it gets more difficult to do it. Maybe you could help on some of these?

> The search box is optional. It can be simply removed if anyone does not like it.

I believe it is a better idea to create a way to easily disable it, like a setting in `_config.yml`. Something like `bib_search: true`, but it could be another name, I didn't give it much thought. So this way one could avoid drawing the search bar and also import the js files.

CheariX:
> > The search box is optional. It can be simply removed if anyone does not like it.
> 
> I believe it is a better idea to create a way to easily disable it, like a setting in `_config.yml`. Something like `bib_search: true`, but it could be another name, I didn't give it much thought. So this way one could avoid drawing the search bar and also import the js files.

Nice idea. Introduced with https://github.com/alshedivat/al-folio/pull/2523/commits/07d6e619cced7a2256bbe6de582ad68f93cd1553

george-gca (approved):

george-gca:
Thanks for the contribution. It is always nice when a lot more people contribute to the project.

garywei944:
It seems `bibsearch.js` is broken, doesn't it? https://alshedivat.github.io/al-folio/

image

george-gca:
I believe it was just fixed on #2560.

CheariX:
No, it is not fixed there.

If I remember correctly, this issue is caused when you filter single key search or empty searches (don't remember exactly).

It is an upstream issue. I did not put much effort to fix it, since I did not want to change upstream code more than necessary.

Did you experience any downsides besides this error message?

Maybe if I find time, I'll fix it. Or anyone else kicks in 😀

george-gca:
Actually the error happens the moment you open the publications page. You don't even have to start typing anything.

CheariX:
Yes, I think this is because it applies the filter automatically upon page loading.
The default filter is the empty string. If you append something like `#mechanics` to the URL, it is applied automatically.

The same error occurs if you enter a search string, and then delete it (using the keyboard)

george-gca:
Fixed in https://github.com/alshedivat/al-folio/commit/8ffd34c9b49f5496c34e327866817ed023c3edf7.

garywei944:
I can confirm that the feature works fine on my Windows Chrome, iOS Chrome, and Safari browsers. But the highlighting doesn't work on my Windows clean Firefox installation.

image

george-gca:
Hummm it's true also for Linux. Would you mind opening an issue for this?

CheariX:
[Firefox is not supported yet](https://caniuse.com/?search=CSS.Highlights), but it will get support soon. 
I mentioned this in my first. 

It should not depend on the operating system. Only the browser should be relevant.

For this reason, I added a backward-compatible search without highlighting: https://github.com/alshedivat/al-folio/blob/8ffd34c9b49f5496c34e327866817ed023c3edf7/assets/js/bibsearch.js#L17-L24

PR #2525, opened 2024-06-24
Fix repo card heigth for different repo descriptions

tiagolobao:
Hello! I had this minor issue on my website and I saw few other people using this template and having the same issue.

**Brief**
if two repo's in the same row has different number of lines for the descriptions, heights of the cards will not be the same if we don't force the number of lines to be displayed.

**Solution**
By looking at [This issue](https://github.com/anuraghazra/github-readme-stats/issues/2900) I could see that they solved it by adding an new option, `description_lines_count`. This was used on the API request in order to fix the issue.

---

## Issue reproduced:

before

---

## Issue fixed after the commit:

after

### DIFF
--- a/_data/repositories.yml
+++ b/_data/repositories.yml
@@ -2,11 +2,13 @@ github_users:
   - torvalds
   - alshedivat
 
+repo_description_lines_max: 2
+
 github_repos:
   - alshedivat/al-folio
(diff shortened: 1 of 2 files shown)


george-gca (approved):

george-gca:
Thanks for fixing this.

PR #2543, opened 2024-07-04
Add support for translatable/customizable section titles in the cv page

romanescherrer:
This PR adds support for translating or customizing section title in the resume from the `_config.yaml`file. This feature enhances the internationalization of the al-folio theme, allowing users to customize the section titles without modifying the liquid template directly.

# Modification
- added a `sectiontitle`parameter in the `_config.yml`to store translations for section titles.
- Updated `cv.liquid` to use titles from `sectiontitle`if available, defaulting to the original section name if a translation is not provided.
- Updated the documentation in `CUSTOMIZE.md` to explain how to used this new feature.

# Example (in French)

```yaml
# _config.yml 
sectiontitles:
  basics: Informations personnelles
  work: experiences
  education: # Leave it blank for the default title
  publications:
  projects: projets
  volunteer:
  awards:
  certificates:
  skills: Compétences
  languages: langues
  interests: hobbits
  references:
```

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -641,3 +641,18 @@ jsonresume:
   - languages
   - interests
   - references
+
+# Customize the section titles in the cv page, leave it blank for default
+sectiontitles:
+  basics:
+  work:
+  education:
+  publications:
+  projects:
+  volunteer:
+  awards:
+  certificates:
+  skills:
+  languages:
+  interests:
+  references:
--- a/CUSTOMIZE.md
+++ b/CUSTOMIZE.md
@@ -48,6 +48,26 @@ There are currently 2 different ways of generating the CV page content. The firs
 
 What this means is, if there is no resume data defined in [\_config.yml](_config.yml) and loaded via a json file, it will load the contents of [\_data/cv.yml](_data/cv.yml). If you want to use the [\_data/cv.yml](_data/cv.yml) file as the source of your CV, you must delete the [assets/json/resume.json](assets/json/resume.json) file.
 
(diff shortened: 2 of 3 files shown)


george-gca:
For a version of al-folio with support for localization, I recommend [my fork of it](https://github.com/george-gca/multi-language-al-folio), which was created specifically for this.

romanescherrer:
@george-gca, thank you for sharing your impressive work. I'm considering using it for my website :). Nevertheless, I believe it's important to accommodate users who might want to customize the title sections on the CV page. For instance, some might prefer to change 'Basic' to 'Personal Informations' or translate it into another language.

_config.yml
```diff
@@ -641,3 +641,18 @@ jsonresume:
   - languages
   - interests
   - references
+
+# Customize the section titles in the cv page, leave it blank for default
+sectiontitles:
```
george-gca, line 646: I believe it's best to use `sections_titles` instead. `jsonresume` is only in a single word because it is the name of the library.

_config.yml
```diff
@@ -641,3 +641,18 @@ jsonresume:
   - languages
   - interests
   - references
+
+# Customize the section titles in the cv page, leave it blank for default
```
george-gca, line 645: Add to this comment the info about it only being used when the cv is loaded from jsonr. Even though it is in that section in _config.yml, it's better to make it clear.

george-gca (changes requested):

george-gca:
@alshedivat @pourmand1376 @arunoruto any thoughts on this PR?

CheariX:
My five cents: our `config.yml` is growing a lot. Do we really want to have so many config parameters in it? Wouldn't it be better to make it configurable on a per-site basis, e.g. in `cv.md`?

george-gca:
I agree with @CheariX. This would be better as an option in the page itself.

PR #2552, opened 2024-07-09
Fix space before some bib commas

sim642:
These somehow appeared when upgrading from v0.11.0 to v0.12.0.


george-gca (approved):

george-gca:
Thanks for the fix.

PR #2553, opened 2024-07-09
Allow URLs in bib award field

sim642:
If the `award` field in a bib entry is an URL, then link to it directly instead of opening an abstract-like text box.

### DIFF
--- a/_layouts/bib.liquid
+++ b/_layouts/bib.liquid
@@ -185,9 +185,15 @@
     <!-- Links/Buttons -->
     <div class="links">
       {% if entry.award %}
-        <a class="award btn btn-sm z-depth-0" role="button">
-          {%- if entry.award_name %}{{ entry.award_name }}{% else %}Awareded{% endif -%}
-        </a>
+        {% if entry.award contains '://' %}
+          <a href="{{ entry.award }}" class="award btn btn-sm z-depth-0" role="button">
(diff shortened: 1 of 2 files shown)


_layouts/bib.liquid
```diff
@@ -185,9 +185,15 @@
     <!-- Links/Buttons -->
     <div class="links">
       {% if entry.award %}
-        <a class="award btn btn-sm z-depth-0" role="button">
-          {%- if entry.award_name %}{{ entry.award_name }}{% else %}Awareded{% endif -%}
-        </a>
+        {% if entry.award contains '://' %}
```
george-gca, line 188: What if the text someone wrote contains a link inside? Maybe it is best to check if the link in not a url with some other checks, for instance, if it doesn't start with one of `http`, `https`, `ftp` and doesn't contain any spaces.

george-gca (changes requested):

CheariX:
Do we really want to handle this within the `award` field?
The example that @george-gca describes shows how much code complexity we must add to handle this.

Would it make sense to use a different field, like  `award_link`, for this use-case?

sim642:
I can do that, but I just followed what's already common in al-folio: the `contains '://'` for determining if something is an URL or not.

george-gca:
Doing what @CheariX suggested makes sense, since it is more straightforward to understand what the field does. Adding a link from `award` itself seems almost like an easter egg.

Issue #2554, opened 2024-07-09
search function is out of order in a distill style post


LiuYing-1:
### Have you checked that your issue isn't already filed?

- [X] I read through [FAQ](https://github.com/alshedivat/al-folio/blob/master/FAQ.md) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [X] Yes, I have checked that this issue isn't already filed.

### Bug description

Search is not working in this page --> https://alshedivat.github.io/al-folio/blog/2021/distill/

### How to reproduce the bug

1. simply navigate to the https://alshedivat.github.io/al-folio/blog/2021/distill/
2. cannot use search there

Thanks in advance :) 

### Error messages and logs

```
The error message you got, with the full traceback if available. Please paste it between these triple backticks.
```

### What operating system are you using?

Mac

### Where are you seeing the problem on?

Deployed site

PR #2555, opened 2024-07-10
Fix search in Distill style post

scottleechua:
Fixes issue #2554: search function is out of order in a distill style post.


george-gca (approved):

george-gca:
Thanks for noticing this.

PR #2560, opened 2024-07-10
fix: search_enabled -> bib_search

CheariX:
In #2523, I did a copy&paste error with https://github.com/alshedivat/al-folio/commit/07d6e619cced7a2256bbe6de582ad68f93cd1553

I used the global `search_enabled` config key instead of the correct `bib_search` key.

This PR fixed it.

### DIFF
--- a/_includes/bib_search.liquid
+++ b/_includes/bib_search.liquid
@@ -1,0 +1,4 @@
+{% if site.bib_search %}
+  <script src="{{ '/assets/js/bibsearch.js' | relative_url | bust_file_cache }}" type="module"></script>
+  <input type="text" id="bibsearch" spellcheck="false" autocomplete="off" class="search bibsearch-form-input" placeholder="Type to filter">
+{% endif %}
--- a/_includes/scripts/misc.liquid
+++ b/_includes/scripts/misc.liquid
@@ -37,11 +37,6 @@
 <script defer src="{{ '/assets/js/common.js' | relative_url | bust_file_cache }}"></script>
 <script defer src="{{ '/assets/js/copy_code.js' | relative_url | bust_file_cache }}" type="text/javascript"></script>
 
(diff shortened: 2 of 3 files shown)


george-gca:
I think it would be better to move the include of the script from [_includes/scripts/misc.liquid](https://github.com/alshedivat/al-folio/blob/master/_includes/scripts/misc.liquid) to the [publications page](https://github.com/alshedivat/al-folio/blob/master/_pages/publications.md), since only that page uses `bibsearch.js` and currently it is being included in all pages as long as the config is set to use it.

CheariX:
Hm. Good point. In my personal setup, I use the bibsearch box on many pages. So it made sense to include it in `misc.liquid`. But most people maybe only have a single bib page.

george-gca:
Maybe you could change for the template, and keep it as is in your personal site since it makes sense.

CheariX:
Sure. I think, this makes sense for most people.

george-gca:
> An alternative would be create a file in. include/bibsearch.liquid and include it. This would allow other people to easily integrate it. What do you think?

I like this idea most, makes sense.

CheariX:
Done.

(I had some Git hiccups, but I think I could solve it with a force push. )

george-gca (approved):

george-gca:
Thanks for the changes. Just out of curiosity, how are you using this search in other parts of your site? I mean, what code changes should you made to reuse this somewhere?

CheariX:
I manage a lot of content on my site using BibTeX for bibliography management. For example, I list my teaching courses, supervised theses, and press articles. 

I found this to be the easiest method since I use Zotero to manage my papers, theses, and other resources anyway, and Zotero has a great auto-export to BibTeX feature (Better BibTeX extension).

PR #2561, opened 2024-07-10
Fixed title search and truncating if larger than 13 words

### LINKED ISSUE #2459: Search broken due to newline in news

### Bug description

See my details in #2450

### How to reproduce the bug

Add markdown link to an inline news article 

Edit: Initially, I thought this issue is caused to due markdown. But I was wrong, it occurs if the content has more than one line.

### Error messages and logs

_No response_

Running locally with Docker

### More info

I think my report in a closed PR was overlooked. Therefore I created this issue.

george-gca:
Fixes #2459

### DIFF
--- a/_includes/scripts/search.liquid
+++ b/_includes/scripts/search.liquid
@@ -27,11 +27,11 @@
     // add the home and posts menu items
     ninja.data = [
       {%- for page in site.pages -%}
-        {%- if page.permalink == '/' -%}{%- assign about_title = page.title -%}{%- endif -%}
+        {%- if page.permalink == '/' -%}{%- assign about_title = page.title | strip -%}{%- endif -%}
       {%- endfor -%}
       {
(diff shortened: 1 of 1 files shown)


CheariX:
I tested this PR.

If I add the following news:

```
---
layout: post
date: 2024-07-10 10:00:00-0400
inline: true
related_posts: false
---

This news contains a [Link](https://example.com)
and two lines
```

I get:

Bildschirmfoto 2024-07-10 um 22 46 03

Note how the lines are joined.

I think, we need to use `newline_to_br` and them replace it with a whitespace to address this. Maybe, there is a smarter solution.

george-gca:
Hummmm I will test this later.

george-gca:
Fixed in https://github.com/alshedivat/al-folio/commit/d904c52149859386ee2087f87696eed86c399bb5

Issue #2268, opened 2024-03-12
Add IETF social integration


liuchunchi:
### Have you checked that your feature request isn't already filed?

- [X] I read through [FAQ](https://github.com/alshedivat/al-folio/blob/master/FAQ.md) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my feature request.
- [X] Yes, I have checked that this feature request isn't already filed.

### Description & Motivation

Add a IETF social integration item (icon, link) in the main page

### Pitch

IETF, short for [Internet Engineering Task Force](https://www.ietf.org/), is the most important Standards Development Organizations (SDO) for Internet and computer networks. 

Contributors at IETF also has a profile page, for example, [Vint Cerf](https://datatracker.ietf.org/person/Dr.%20Vinton%20G.%20Cerf), [Scott Bradner](https://datatracker.ietf.org/person/sob@sobco.com).

In the main page of al-folio there are icons linked to different external websites, configurable in `_config.yml` as 
```
# -----------------------------------------------------------------------------
# Social integration
# -----------------------------------------------------------------------------

github_username: # your GitHub user name
gitlab_username: # your GitLab user name
x_username: # your X handle
mastodon_username: # your mastodon instance+username in the format instance.tld/@username
linkedin_username: # your LinkedIn user name
telegram_username: # your Telegram user name
scholar_userid: # your Google Scholar ID
```
can we add a `ietf_handle`? The value is usually an email address. The link is usually `https://datatracker.ietf.org/person/<ietf_handle>`, for example [https://datatracker.ietf.org/person/liuchunchi@huawei.com](https://datatracker.ietf.org/person/liuchunchi@huawei.com)

george-gca:
Sure, can you send a PR for this? Just check how this is done for other social integrations.

CheariX:
When starting to implement this, I noticed that there seems to be no IETF icon available.
Currently, al-folio uses [Academicons](https://jpswalsh.github.io/academicons/) and [Fontaweseome](https://fontawesome.com/). I could not find it there, but maybe I missed something?

george-gca:
I believe it is not part of any of them. We also have recently added tabler icons support in #2261. Now I believe the best solution for this case would be to add the SVG code for it.

scottleechua:
[...]

PR #2562, opened 2024-07-11
Enable custom SVG as social icon

scottleechua:
Addresses Issue #2268: Add IETF social integration.

The social integration aspect is straightforward, but [as mentioned](https://github.com/alshedivat/al-folio/issues/2268#issuecomment-2053612898) in the original issue, adding this feature introduces the **use of SVGs as social icons**.

## Current approach
Seeking wider input on how we want to handle SVG social icons. 

In this PR, I made a new social entry `ietf_id` that would show up like this:

Screenshot 2024-07-11 at 10 31 29

I enforced an SVG height of `4rem` to match the styling on the font icons.

It's possible to style the SVG further to match the appearance with the font icons (e.g., enforce monochrome, remove wordmark), but personally I think it wouldn't look as nice.

## Possible Alternative Approach
A possible alternative is to take this as an opportunity to create a more general `svg` entry under social integrations in `_config.yml`, for users to bring their own SVGs. Something like:

```
svg:
  logo: _includes/svg/<LOGO>.svg
  link: https://website.com/profile
```

Then styling would be up to the user.


george-gca:
I don't think I like the idea. The point of using as a font is to make sure the styling of the icons are the same, considering colors, size, etc. If I change the font color in css I would expect to all icons change their color. Also I don't know if I would add this to our `_config.yml`, but currently I have no other suggestion.

I am considering starting a fork of academicons to give support for this kind of issue (since it is being over a year that they made any changes), but I am not sure how hard it is to do this.

scottleechua:
I agree that a font-based solution would absolutely be the best case, yes. I wonder if it's worth reaching out to the owner of Academicons to see what his plans are for the repo. I personally don't mind maintaining it, and would be willing to reach out. What do you think?

george-gca:
I doesn't hurt to try. There are many open requests for icons there, and the last update to the repo was over a year ago.

george-gca:
But I don't think we should close the issue just yet. Maybe we can think of a better way for someone to insert their svgs, maybe a guideline of how it should be, idk.

scottleechua:
I'm also on the fence about this. I think the most helpful guiderail this provides is enforcing the SVG height to be the same as the other social icons. Happy to leave this open so other users with a stake can share their thoughts. 😄

scottleechua:
I've reached out to the Academicons folks for a chat at the end of the month, and will update here on the outcome of that conversation. 👌

Issue #2443, opened 2024-05-26
Altmetric badges don't display for some pubs


etjernst:
### Have you checked that your issue isn't already filed?

- [X] I read through [FAQ](https://github.com/alshedivat/al-folio/blob/master/FAQ.md) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [X] Yes, I have checked that this issue isn't already filed.

### Bug description

For some of my publications, the Altmetric badge doesn't display

### How to reproduce the bug

Despite having the correct doi and `altmetric = {true},` in the .bib file, and the publication having an entry on [Altmetric](https://www.altmetric.com/explorer) when I search for it there, the badge doesn't render on my publications file. 

Example pub for which the altmetric score doesn't render:  [10.1016/j.worlddev.2017.12.002](https://doi.org/10.1016/j.worlddev.2017.12.002)

### What operating system are you using?

Not applicable (e.g. you're using GitHub Pages or other hosting)

### Where are you seeing the problem on?

Deployed site

### More info

The site has deployed without errors, but the badge is still missing
![image](https://github.com/alshedivat/al-folio/assets/14223886/6f82feea-ab58-428f-a72a-d3fde670dd7e)

george-gca:
As you can see [from this code on](https://github.com/alshedivat/al-folio/blob/96c4e613854509c37f36ae0e63a616e6be342547/_layouts/bib.liquid#L239) is where we enable the altmetrics. We use the [altmetric code contained in their documentation](https://badge-docs.altmetric.com/getting-started.html), so there is not much we can do about it. The reasons this might not be working that I can think of are:
1. for some reason this altmetric information in your paper is not yet available
2. there is a limit on the number of badges you can display in your site? Idk

etjernst:
> for some reason this altmetric information in your paper is not yet available
I do know for sure that it has a badge when I look it up directly.

> there is a limit on the number of badges you can display in your site? Idk
Hm that would be weird. It's also not in "order" i.e. the first instance on my publications page is missing the badge and a later one has it.

PR #2565, opened 2024-07-11
Make publication badges always visible

scottleechua:
## The issue
Currently Altmetric and Dimension publication badge elements have non-obvious attributes that hide badges when some conditions are not met ,e.g.:
```
data-hide-no-mentions="true"
data-hide-less-than="15"
```
resulting in seemingly strange behavior where badges are enabled in `config.yml` but don't show up consistently, as reported in #2443 : Altmetric badges don't display for some pubs.

## This PR
- removes these hidden nondisplay conditions in favor of more predictable website behavior;
- adds documentation links to point users interested in customizing badge behavior to the right resources.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -338,7 +338,7 @@ scholar:
 
 # Display different badges withs stats for your publications
 enable_publication_badges:
-  altmetric: true # Altmetric badge (https://www.altmetric.com/products/altmetric-badges/)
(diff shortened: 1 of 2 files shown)


george-gca (approved):

george-gca:
Thanks for noticing this. I never realized those were there.

PR #2571, opened 2024-07-13
Replaced jekyll-minifier that uses uglifier by terser

### LINKED ISSUE #2548: Uglifier: Support ES6 Syntax

### Description & Motivation

Uglifier currently fails when ES6 synatax is used.
See e.g. https://github.com/alshedivat/al-folio/actions/runs/9649917821/job/26614403326#step:5:363 in #2523.

### Pitch

Uglifier should support ES6 syntax to allow better code quality, i.e. easier to read, less to write.

### Alternatives

_No response_

### Additional context

_No response_

george-gca:
Hopefully fixes #2548.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,7 +9,6 @@ group :jekyll_plugins do
     gem 'jekyll-imagemagick'
     gem 'jekyll-jupyter-notebook'
     gem 'jekyll-link-attributes'
-    gem 'jekyll-minifier'
     gem 'jekyll-paginate-v2'
     gem 'jekyll-regex-replace'
     gem 'jekyll-scholar'
@@ -19,6 +18,7 @@ group :jekyll_plugins do
     gem 'jekyll-twitter-plugin'
     gem 'jemoji'
     gem 'mini_racer'
+    gem 'terser'
     gem 'unicode_utils'
     gem 'webrick'
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -50,9 +50,10 @@ GEM
     feedjira (3.2.3)
       loofah (>= 2.3.1, < 3)
       sax-machine (>= 1.0, < 2)
-    ffi (1.17.0)
+    ffi (1.17.0-aarch64-linux-gnu)
     ffi (1.17.0-arm64-darwin)
     ffi (1.17.0-x86_64-darwin)
+    ffi (1.17.0-x86_64-linux-gnu)
     forwardable-extended (2.6.0)
     gemoji (4.1.0)
     google-protobuf (3.25.3-aarch64-linux)
@@ -170,7 +171,6 @@ GEM
       forwardable-extended (~> 2.6)
     public_suffix (6.0.0)
     racc (1.8.0)
-    rake (13.2.1)
     rb-fsevent (0.11.2)
     rb-inotify (0.11.1)
       ffi (~> 1.0)
@@ -178,17 +178,20 @@ GEM
       strscan
     rouge (4.3.0)
     safe_yaml (1.0.5)
-    sass-embedded (1.69.5)
+    sass-embedded (1.69.5-aarch64-linux-gnu)
       google-protobuf (~> 3.23)
-      rake (>= 13.0.0)
     sass-embedded (1.69.5-arm64-darwin)
       google-protobuf (~> 3.23)
     sass-embedded (1.69.5-x86_64-darwin)
+      google-protobuf (~> 3.23)
+    sass-embedded (1.69.5-x86_64-linux-gnu)
       google-protobuf (~> 3.23)
     sax-machine (1.3.2)
     strscan (3.1.0)
     terminal-table (3.0.2)
       unicode-display_width (>= 1.1.1, < 3)
+    terser (1.2.3)
+      execjs (>= 0.3.0, < 3)
     tzinfo (2.0.6)
       concurrent-ruby (~> 1.0)
     uglifier (4.2.0)
@@ -226,6 +229,7 @@ DEPENDENCIES
   jekyll-twitter-plugin
   jemoji
   mini_racer
+  terser
   unicode_utils
   webrick
 
--- a/_config.yml
+++ b/_config.yml
@@ -260,7 +260,6 @@ plugins:
   - jekyll-imagemagick
   - jekyll-jupyter-notebook
   - jekyll-link-attributes
-  - jekyll-minifier
   - jekyll-paginate-v2
   - jekyll-regex-replace
   - jekyll/scholar
@@ -269,6 +268,7 @@ plugins:
   - jekyll-toc
   - jekyll-twitter-plugin
   - jemoji
+  - terser
 
 # Sitemap settings
 defaults:
@@ -281,13 +281,12 @@ sass:
   style: compressed
 
 # -----------------------------------------------------------------------------
-# Jekyll Minifier
+# Terser settings
 # -----------------------------------------------------------------------------
 
-jekyll-minifier:
-  exclude: ["robots.txt", "assets/js/search/*.js"]
-  uglifier_args:
-    harmony: true
+terser:
+  compress:
+    drop_console: true
 
 # -----------------------------------------------------------------------------
 # Jekyll Archives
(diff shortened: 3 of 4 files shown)


george-gca:
@CheariX can you test if this change fixes your issue with ES6?

JoJoDeveloping:
Note that `jekyll-minifier` also minized HTML, CSS and JSON. Your solution no longer does this.

george-gca:
@JoJoDeveloping you are right, so we can't get rid of `jekyll-minifier` just yet.

I made a PR for [jekyll-terser](https://github.com/RobertoJBeltran/jekyll-terser) with my code, but you are right. I wll send a PR for `jekyll-minifier`.

george-gca:
I made a PR for both [jekyll-minifier](https://github.com/digitalsparky/jekyll-minifier/pull/62) and [jekyll-terser](https://github.com/RobertoJBeltran/jekyll-terser). If the second one is accepted, we can drop the plugin I created and use `jekyll-terser` instead. If the first one is accepted, we can drop both my plugin and `jekyll-terser` and use only `jekyll-minifier`.

CheariX:
Thank you @george-gca for taking care of this issue!

When I wrote it initially, I was not aware that al-folio does not use `uglifier` directly but via `jekyll-minifier`. Therefore, changing from `uglifier` to `terser` is more complex than expected.

I hope that your upstream PRs get merged soon.

Anyways, I tested this PR and it worked flawlessly on my site. I vote for merging this PR into `main`. 

Using this PR, I can finally deploy my website again 😀

george-gca:
I was about to ask you this, so the minified js is looking right. Is this also true for the css and html files? And what about when jekyll env is production?

CheariX:
- CSS looks minimized to me too. However, it looks identically to what we had before.
- HTML looks different to the minimized code that we had before. Before, we got only a single line. Now, it seems that at least JavaScript inside `<script>` is not minimized. The HTML itself, however, looks like it was minimized.
- Currently, JavaScript file are also minimized in non-production environments. I added a code suggestion to "fix" this. HTML/CSS is not touched by terser, since it only touches JavaScript files

george-gca:
I thought about adding minification only on production environments, but decided not to since using it on develop environments helped me get it right. Maybe it is a good idea to add a new param somehow to let the user specify if it should run only on production or not, and by default do this.

> I quickly checked where we currently use embedded JavaScript. There are only a very few places (disqus.liquid, mathjax.liquid, misc.liquid, progressBar.liquid, pseudecode.liquid). I'm not sure why these files use embedded JavaScript instead of using *.js files but they could be easily refactored if this is desired. Actually, best practices is also to avoid embedded JavaScript (e.g., deployment of Content-Security Policy is easier, but we don't use it anyway at the moment)

Thanks for the lookup. I think I will handle these guys next in this PR. I will only push it forward when the minification is as close as possible to the one with uglifier.

JoJoDeveloping:
> using it on develop environments helped me get it right.

For me, minifying takes quite a long time, on the order of several tens of seconds in addition to the rest. It makes the build much slower, and therefore I'm happy I don't have to deal with it when developing/prototyping. So please leave the option to have it `production`-only (or invest the effort into making it real fast 🙃).

george-gca:
Another possibility is to migrate the solution to npm-based like already done with `PurgeCSS`. In this case I would use [html-minifier-terser](https://github.com/terser/html-minifier-terser). What are your considerations about this approach?

CheariX:
> > using it on develop environments helped me get it right.
> 
> For me, minifying takes quite a long time, on the order of several tens of seconds in addition to the rest. It makes the build much slower, and therefore I'm happy I don't have to deal with it when developing/prototyping. So please leave the option to have it `production`-only (or invest the effort into making it real fast 🙃).

I'd second this. Minifying takes some time, so I would like to skip this step in development. Also, I find the non-minified code quite nice for feature development. 

> Another possibility is to migrate the solution to npm-based like already done with `PurgeCSS`. In this case I would use [html-minifier-terser](https://github.com/terser/html-minifier-terser). What are your considerations about this approach?

I also thought about alternatives to `jekyll-minifier`. `html-minifier-terser` seems to be the logical solution for this. However, I have no strong opinion about that. At least, it would compress embedded JavaScript (which I think we don't need; we can remove it).

JoJoDeveloping:
Perhaps we should do performance testing between the various alternatives?

george-gca:
Tbh I am not really concerned about performance here, since this will only run on production environment. I am worried about maintainability, both of our code and the ones we are depending on.

For instance, [jekyll-minifier](https://github.com/digitalsparky/jekyll-minifier) last code update was 4 years ago, and it uses under the hood 4 packages: [uglifier](https://github.com/lautis/uglifier) for js (last updated 4 years ago, though [UglifyJS](https://github.com/mishoo/UglifyJS) is heavily maintained), [CSSminify](https://github.com/digitalsparky/cssminify) (last updated 4 years ago), [Htmlcompressor](https://github.com/paolochiodi/htmlcompressor) (last updated 7 years ago), and [JSON::Minify](https://github.com/geoffyoungs/json-minify-rb) (last updated 8 years ago). I am not saying that all these should be continuously updated (since JSON, for instance, is a well-defined standard), but looking at code that haven't seem an update in a while gives a certain insecurity.

Meanwhile [html-minifier-terser](https://github.com/terser/html-minifier-terser) was updated last year. Maybe it is a good moment to migrate to other solution for code minification.

CheariX:
Not stale.

george-gca:
Update: my PR for [jekyll-terser](https://github.com/RobertoJBeltran/jekyll-terser/pull/1) just got accepted. Maybe now I can take a look at this PR again.

george-gca:
> I quickly checked where we currently use embedded JavaScript. There are only a very few places (disqus.liquid, mathjax.liquid, misc.liquid, progressBar.liquid, pseudecode.liquid). I'm not sure why these files use embedded JavaScript instead of using *.js files but they could be easily refactored if this is desired. Actually, best practices is also to avoid embedded JavaScript (e.g., deployment of Content-Security Policy is easier, but we don't use it anyway at the moment)

@CheariX I tried to do a major refactoring to remove as much as possible inline js code. Please test if the current solution is working fine. Also can you give it a try @JoJoDeveloping?

CheariX:
Thank you @george-gca for this amazing work.

I've tested it with my personal page.
It seems to work as expected. HTML and JS are minified. No errors to compression.

However, I found the following bug during the `Install and Build` action:
> Imagemagick: Command returned pid 2651 exit 127 with error sh: 1: convert: not found

The effect is that many preview images (e.g. 480p) are not generated. I can see this on my publication images.

I'm unsure if this is related to this particular PR or to anything else. I didn't merge upstream on my site for a while (since 2024-10-25). I think the netlify preview of this PR does not have these issues.

george-gca:
It is unrelated, and it has been fixed on https://github.com/alshedivat/al-folio/commit/ec7d7c34e8f1a7ef4da2ea37454f470b8b37094b. I just haven't merged it with main yet.

CheariX:
Perfect. Good to know that this one is fixed. Then, this PR looks good to me.

sim642:
> I made a PR for both [jekyll-minifier](https://github.com/digitalsparky/jekyll-minifier/pull/62) and [jekyll-terser](https://github.com/RobertoJBeltran/jekyll-terser). If the second one is accepted, we can drop the plugin I created and use `jekyll-terser` instead. If the first one is accepted, we can drop both my plugin and `jekyll-terser` and use only `jekyll-minifier`.

The jekyll-minifier upgrade would be significantly better. Because jekyll-terser borrows some architecture which assumes 10-year-old Jekyll internals that have a massive limitation: assets from gems like jekyll plugins are mishandled.

george-gca:
Can you elaborate a little bit more @sim642? Which plugins misbehave?

sim642:
jekyll-minifier has recently been updated to also use terser itself, so this should be reverted to some extent at least.
Because al-folio still depends on jekyll-minifier as well, so now it has two different plugins running terser, which is completely redundant.
This would also fix regressions like https://github.com/RobertoJBeltran/jekyll-terser/issues/3.

PR #2576, opened 2024-07-15
Encoded email in CV using jekyll-email-protect

tonideleo:
I added a simple solution where the email provided in the json CV file can also be obfuscated using jekyll-email-protect. In the json file, the email field has now two entries, the first one being the email address and the second being the string to cover up the email.

### DIFF
--- a/_includes/resume/basics.liquid
+++ b/_includes/resume/basics.liquid
@@ -2,8 +2,7 @@
   {% assign skip_basics = 'image,profiles,location' | split: ',' %}
   {% for content in data[1] %}
     {%
-      if (content[1] == "") or (skip_basics contains
-      content[0])
+      if (content[1] == "" ) or (skip_basics contains content[0])
     %}
       {% continue %}
     {% endif %}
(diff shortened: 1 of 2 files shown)


george-gca (changes requested):
1 - you forgot to add the gem to the commit
2 - is this solution still relevant? I mean, it seems a too simple substitution to be made to access the correct email again

george-gca:
It is not that I dislike it, it is more of a question (philosophical maybe?) in itself. With the evolution of all things in the internet, don't bots already handle this solution, since it is simple and apparently common?

george-gca:
> This solution applies similarly to the email button at the bottom pages.

Can you also add it there then?

CheariX:
Well.. I'm not convinced that this adds an level of protection. The Email is simply URL encoded, so calling `decodeURI("%61.%64...")` reveals your Email. I'm pretty sure that phishers are smart enough to handle this.

tonideleo:
Fair point, I was just wondering because it was already used in the footer. Thanks for checking this out!

PR #2617, opened 2024-08-04
[bug-fix] Add padding to default markdown table cells

rstein66:
Default, meaning `pretty_table: false`

## Sample code

```md
|   First Column   |  Second Column  |  Third Column  |
|------------------|-----------------|----------------|
| Sed in.          | Sed non.        | Morbi egestas. |
| Donec facilisis. | Suspendisse eu. | Nulla porta.   |
| Praesent a.      | Interdum et.    | Sed nec.       |
```

### Current result 

current-default

### Proposed result

updated-default

### DIFF
--- a/_posts/2023-03-21-tables.md
+++ b/_posts/2023-03-21-tables.md
@@ -10,7 +10,20 @@ related_posts: true
 pretty_table: true
 ---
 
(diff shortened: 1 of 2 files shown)


george-gca:
Thanks for your contribution.

PR #2622, opened 2024-08-05
[Bug-fix] Make custom blockquote font coloring consistent

rstein66:
Currently, the tip, warning, and danger custom blockquote's font color is not customized when the text is styled as bold, italics, or a list item. As a result, the text is slightly less attractive in light mode and almost illegible in dark mode. 

## Screenshot: Current

current-darkmode

## Screenshot: Proposed

proposed-darkmode

### DIFF
--- a/_sass/_base.scss
+++ b/_sass/_base.scss
@@ -71,7 +71,10 @@ blockquote {
     border-color: var(--global-tip-block);
     background-color: var(--global-tip-block-bg);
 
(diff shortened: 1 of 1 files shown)


george-gca (approved):

george-gca:
Thanks for your contribution. Never used these styles inside the blockquotes, so never really saw these bugs.

PR #2647, opened 2024-08-20
Fix no github_users titling in repositories.md

CSautier:
Inverted order of title and {% if site.data.repositories.github_users %}, so that if there is no github_users, the "GitHub users" title does not appear.


george-gca (approved):

george-gca:
Thanks for changing this. As a matter of fact, I am thinking if it is possible to add cards like these but for GitLab or other online git instances.

PR #2648, opened 2024-08-20
Update _config.yml to add a filtered bibtex keyword

CSautier:
Added the google_scholar_id to filtered keywords


george-gca (changes requested):

george-gca (approved):

george-gca:
Thanks for fixing this.

PR #2649, opened 2024-08-20
Added gh-pages Formatter

KingHowler:
# Added prettier-hmtl.yml
## Purpose
The GitHub Workflow formats the html files on gh-pages. The html files generated are always on a single line. This makes scaling programs a lot more difficult. By formatting the HTML files, al-folio can now be used to generate code which can then be modified to allow for using back-end.

## Errors found
I want to let you know that when I was using prettier for this, it kept crashing and after some debugging I found out that al-folio was generating an invalid tag ```</source>```. ```<source>``` is a self-closing tag and doesn't have a separate closing tag.
Error: ```<source src="URL" type="type"></source>```
Correct: ```<source src="URL" type="type">```

## Workflow Description
1. The workflow starts by checking out the gh-pages branch.
2. Then it finds all ```</source>``` tags in all html files and deletes them.
3. It Installs NodeJS and then Prettier. To make sure the code was executed properly, the workflow checks if prettier is present.
4. Then the workflow runs prettier on all html files present in gh-pages
5. It ends by committing the changes and pushing them to the gh-pages directory

# Example:
> Before
> image

> After
> image

### DIFF
--- a/.github/workflows/prettier-html.yml
+++ b/.github/workflows/prettier-html.yml
@@ -13,14 +13,13 @@ jobs:
           ref: gh-pages
 
       - name: Find and Remove </source> Tags
-        run:  find . -type f -name "*.html" -exec sed -i 's/<\/source>//g' {} +
-      
+        run: find . -type f -name "*.html" -exec sed -i 's/<\/source>//g' {} +
+
       - name: Set up Node.js
         uses: actions/setup-node@v4
 
       - name: Install Prettier
         run: npm install -g prettier
-
 
       - name: Check for Prettier
         run: npx prettier --version || echo "Prettier not found"
--- a/.github/workflows/prettier-html.yml
+++ b/.github/workflows/prettier-html.yml
@@ -1,4 +1,4 @@
-name: Pretty HTML
+name: Prettify gh-pages
 
 on:
   workflow_dispatch:
--- a/.github/workflows/prettier-html.yml
+++ b/.github/workflows/prettier-html.yml
@@ -1,9 +1,6 @@
 name: Pretty HTML
 
 on:
-  push:
-    branches:
-      - gh-pages
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/prettier-html.yml
+++ b/.github/workflows/prettier-html.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout gh-pages branch
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           ref: gh-pages
 
--- a/.github/workflows/prettier-html.yml
+++ b/.github/workflows/prettier-html.yml
@@ -16,7 +16,7 @@ jobs:
         run:  find . -type f -name "*.html" -exec sed -i 's/<\/source>//g' {} +
       
       - name: Set up Node.js
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
 
       - name: Install Prettier
         run: npm install -g prettier
--- a/.github/workflows/prettier-html.yml
+++ b/.github/workflows/prettier-html.yml
@@ -1,0 +1,40 @@
+name: Pretty HTML
+
+on:
+  push:
+    branches:
+      - gh-pages
+  workflow_dispatch:
+
+jobs:
+  format:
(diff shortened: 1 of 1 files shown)


george-gca:
This action is useful when you want to modify your page further **after** your page was built. I believe that most of our users (me included) host their pages on GitHub pages or don't even add changes after the site was built. Therefore, if this action would be added, it should not run by default, but be enabled if the user is willing to do so instead.

george-gca:
Can you pinpoint what creates these `</source>` html tags?

KingHowler:
> Can you pinpoint what creates these `</source>` html tags?

I can try but I I'm not exactly sure I will be able to find it.
And yes that's what I was saying it for. In case someone wants to add a back-end for their al-folio website.

KingHowler:
> This action is useful when you want to modify your page further **after** your page was built. I believe that most of our users (me included) host their pages on GitHub pages or don't even add changes after the site was built. Therefore, if this action would be added, it should not run by default, but be enabled if the user is willing to do so instead.

That's reasonable,
I have removed the following section:
```yaml
  push: 
     branches: 
       - gh-pages
```

The code will now work when manually triggered from GitHub actions

.github/workflows/prettier-html.yml
```diff
@@ -0,0 +1,37 @@
...
+      - name: Checkout gh-pages branch
+        uses: actions/checkout@v3
+        with:
+          ref: gh-pages
+
+      - name: Find and Remove </source> Tags
+        run:  find . -type f -name "*.html" -exec sed -i 's/<\/source>//g' {} +
+      
+      - name: Set up Node.js
+        uses: actions/setup-node@v3
```
george-gca, line 19: Let's use the [latest version](https://github.com/actions/setup-node?tab=readme-ov-file#usage) `v4` here.

.github/workflows/prettier-html.yml
```diff
@@ -0,0 +1,37 @@
...
+
+on:
+  workflow_dispatch:
+
+jobs:
+  format:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout gh-pages branch
+        uses: actions/checkout@v3
```
george-gca, line 11: Let's use the [latest version](https://github.com/actions/checkout?tab=readme-ov-file#usage) `v4` here.

.github/workflows/prettier-html.yml
```diff
@@ -0,0 +1,37 @@
+name: Pretty HTML
```
george-gca, line 1: I think it's best to use a name that is more meaningful. Idk something like `Prettify built site` or something like that that makes clear what exactly is being modified.

.github/workflows/prettier-html.yml
```diff
@@ -0,0 +1,37 @@
...
+
+      - name: Install Prettier
+        run: npm install -g prettier
+
+
+      - name: Check for Prettier
+        run: npx prettier --version || echo "Prettier not found"
+
+      - name: Run Prettier on HTML files
+        run: npx prettier --write '**/*.html'
```
george-gca, line 29: Using this regex also gets the html files in the root directory right? Just to be sure, because I don't remember where in the code using `**/` only gets the files in subdirs.

george-gca (changes requested):

.github/workflows/prettier-html.yml
KingHowler: Yes it gets all of the html files in the root directory as well.
```**/``` means all folders including root and ```*.html``` means all html files

.github/workflows/prettier-html.yml
KingHowler: Does ```HTML Formatter``` sound good?
or perhaps ```Structure HTML```?

george-gca: The point is which html? Because we already have a [prettier](https://github.com/alshedivat/al-folio/blob/master/.github/workflows/prettier.yml) GitHub action that tries to format the code. I think it should make clearer which HTMLs it is formatting, which are the ones that are the result of the page built. Maybe something like `Prettify gh-pages`?

KingHowler: Oo let's go with that.

george-gca (approved):

george-gca (approved):

Issue #2086, opened 2024-01-16
Rename `master` branch to `main`


george-gca:
Many communities, both on [GitHub](https://github.com/github/renaming) and in the [wider Git community](https://sfconservancy.org/news/2020/jun/23/gitbranchname/), have renamed the default branch name of their repository from master to main. I believe it would be nice if [we did the same](https://github.com/github/renaming?tab=readme-ov-file#renaming-existing-branches).

JoaoDiogoDuarte:
Just want to express my support for this.

scottleechua:
+1 here too. Does this create any big blockers to development?

george-gca:
I believe not. The [main changes](https://github.com/github/renaming?tab=readme-ov-file#rename-existing) that will happen for other users are:
- Show a notice to repository contributors, maintainers, and admins on the repository homepage with instructions to update local copies of the repository
- Show a notice to contributors who `git push` to the old branch

scottleechua:
Sounds about right! Perhaps this change can be scheduled to happen along with the 2024 Q4 release?

alshedivat:
added `main` branch and made it default. I kept `master` for posterity / to avoid disruption.

george-gca:
Before closing this issue I think we should update our documentation pointing to `main` whenever it was `master`. I can do that.

PR #2659, opened 2024-08-25
Posts Scheduler

KingHowler:
# Posts Scheduler
### Purpose
The workflow uploads posts that are scheduled for a later date automatically. The ability to schedule posts is a very important tool and I feel like it should also be included in this repository.

### Error **?**
I don't know for certain if this is an error or if it is intentional but al-folio doesn't publish posts which are dated on the future.'

**How I found this out:**
I was making a post but no matter what I did it just wouldn't show up at my website. After starting from scratching and changing everything one-at-a-time to see where the code breaks I found out that the date was causing the post to not be uploaded, the filename was ```2024-08-25-lecture1.md``` and the day I was uploading this post was also 25-August-2024. When I tried to upload it after half a day it was allowed to be uploaded.

After comparing times I found out that at the time of the failed uploading the date was 25-Aug in my region and 24-Aug at UTC. Hence I concluded that only posts which are dated for the present or the past is uploaded on the website

### Another thing to notice:
The website is only uploaded when the ```Deploy``` workflow is run. This means that the user must be online and manually deploy the website.

## Quick Fix
You can solve this issue by having the deploy function run every day. The issue with that is the commit history for ```gh-pages``` branch. Deploying every day will make the commit history very annoyingly big.

## Better Solution
In the main branch, add a new folder ```_scheduled``` and make posts for a later date in there. Then using the workflow I made, the posts should automatically deployed on the date in their names. If there are no files scheduled for today, then there would be no changes made hence keeping the commit history minimal.

### DIFF
--- a/_plugins/external-posts.rb
+++ b/_plugins/external-posts.rb
@@ -42,7 +42,16 @@ module ExternalPosts
     end
 
     def create_document(site, source_name, url, content)
-      slug = content[:title].downcase.strip.gsub(' ', '-').gsub(/[^\w-]/, '')
+      # check if title is composed only of whitespace or foreign characters
+      if content[:title].gsub(/[^\w]/, '').strip.empty?
+        # use the source name and last url segment as fallback
+        slug = "#{source_name.downcase.strip.gsub(' ', '-').gsub(/[^\w-]/, '')}-#{url.split('/').last}"
+      else
+        # parse title from the post or use the source name and last url segment as fallback
+        slug = content[:title].downcase.strip.gsub(' ', '-').gsub(/[^\w-]/, '')
+        slug = "#{source_name.downcase.strip.gsub(' ', '-').gsub(/[^\w-]/, '')}-#{url.split('/').last}" if slug.empty?
+      end
+
       path = site.in_source_dir("_posts/#{slug}.md")
       doc = Jekyll::Document.new(
         path, { :site => site, :collection => site.collections['posts'] }
@@ -80,7 +89,7 @@ module ExternalPosts
       html = HTTParty.get(url).body
       parsed_html = Nokogiri::HTML(html)
 
-      title = parsed_html.at('head title')&.text || ''
+      title = parsed_html.at('head title')&.text.strip || ''
       description = parsed_html.at('head meta[name="description"]')&.attr('content') || ''
       body_content = parsed_html.at('body')&.inner_html || ''
 


george-gca:
> The ability to schedule posts is a very important tool and I feel like it should also be included in this repository.

I agree this would be useful.

> I don't know for certain if this is an error or if it is intentional but al-folio doesn't publish posts which are dated on the future.'

Actually this is a [static sites generator](https://www.cloudflare.com/learning/performance/static-site-generator/) (jekyll in this case) thing. Someone even [opened an issue in jekyll](https://github.com/jekyll/jekyll/issues/6536) a long time ago, but summarizing: static sites generators build a website and then it is done. It can't change a site after it has been built. Since the blog is in a future date, it just ignores it during build.

Now, how would work your solution if a post has a timestamp for later in the day? For example, the post name has `2024-08-27`, but inside the post has something like `date: 2024-08-27 15:09:00`? I believe when your cron job runs it will move the posts correctly to the `_posts/` dir, but during build these posts will be ignored, and will only be included in the next build.

.github/workflows/posts-scheduler.yml
```diff
@@ -0,0 +1,45 @@
...
+jobs:
+  publish_posts:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout main branch
+        uses: actions/checkout@v4
+        with:
+          ref: main
+
+      - name: Set up Node.js environment
```
george-gca, line 18: What is it needed for?

_scheduled/README.md
george-gca, line 1: I think it is nice to add `/` to all directories here, like `_drafts/`.

_scheduled/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+example file: ```2024-08-26-This file will be uploaded on 26 August.md```
+
+## Important Notes:
+- The scheduler uploads posts every 🕛 00:00 UTC
+- It will only upload posts that are scheduled on their respective days
+- It will only upload files which follow the pattern ```yyyy-mm-dd-title.md```
+  - This means that only markdown files will be posted
+  - It means that any markdown which do not follow this pattern (such as this README.md file) will not be posted
+- The scheduler will only post files to the ```_posts``` folder, it will not post to folders like ```_projects``` or ```_news```
```
george-gca, line 16: I would change this to:

- The scheduler works by moving posts from the `_scheduled/` directory to `_posts/`, it will not work for other collections like `_projects/` or `_news/`.

george-gca (changes requested):

KingHowler:
> > The ability to schedule posts is a very important tool and I feel like it should also be included in this repository.
> 
> I agree this would be useful.
> 
> > I don't know for certain if this is an error or if it is intentional but al-folio doesn't publish posts which are dated on the future.'
> 
> Actually this is a [static sites generator](https://www.cloudflare.com/learning/performance/static-site-generator/) (jekyll in this case) thing. Someone even [opened an issue in jekyll](https://github.com/jekyll/jekyll/issues/6536) a long time ago, but summarizing: static sites generators build a website and then it is done. It can't change a site after it has been built. Since the blog is in a future date, it just ignores it during build.
> 
> Now, how would work your solution if a post has a timestamp for later in the day? For example, the post name has `2024-08-27`, but inside the post has something like `date: 2024-08-27 15:09:00`? I believe when your cron job runs it will move the posts correctly to the `_posts/` dir, but during build these posts will be ignored, and will only be included in the next build.

That is certainly an interesting case, but we can't run deploy once every hour, it will clutter up the actions tab too much.

This can be solved with a bit of sacrifice. Instead of running the scheduler at 00:00 we can run it at 23:59

This way all of the posts will be deployed. On that specific day but only right before midnight which will delay it by alot.

Another option is to use delays but that will cause the workflow to run for potential upto 7-8 hours and that will cause problems for GitHub itself

A fourth option which I think may have a good chance of working is by separating this workflow into 2
- A scanner which runs every 00:00 utc and modifies the publisher. The scanner will also run once everytime after the publisher finishes
- A publisher which will move the files and push

KingHowler:
> Maybe we could also fetch the timestamp from the post itself and use it somehow? For example, by doing:
> 
> ```bash
> sed -ne '/---/,/---/{/---/N;p}' 2015-03-15-formatting-and-links.md | grep "date: " | cut -c7-
> ```
> 
> We can obtain the timestamp `2015-03-15 16:40:16` from inside the post.

Yeah but that only helps with comparison. The difficult part is to trigger the event as minimally as possible.

I've got an idea but I am a bit busy for a few days. I'll send a diagram of a possible solution this Thursday.

george-gca:
I believe the easiest solution would be to trigger the action twice a day (00:00 and 23:59), and check each time which posts based on the timestamp should be moved to `_posts/`.

_scheduled/README.md
KingHowler: Ok

CheariX:
> Maybe we could also fetch the timestamp from the post itself and use it somehow? For example, by doing:
> 
> ```shell
> sed -ne '/---/,/---/{/---/N;p}' 2015-03-15-formatting-and-links.md | grep "date: " | cut -c7-
> ```
> 
> We can obtain the timestamp `2015-03-15 16:40:16` from inside the post.

Just my few cents: instead of `grep` + `cut`, one could also use `awk '/^date:/ {print $2}'` to print the date only (If I understood correctly, the time is not necessary. Otherwiese `print $2, $3`). This is maybe less error-prone in case there are simple formatting issues, like multiple whitespaces after the colon.

george-gca:
@CheariX the time is necessary.

@KingHowler then the action itself would change its own cron schedule? It is feasible, but I think it is kind of overcharged. For example, if someone adds like 5 scheduled posts for the same day, but for different timestamps, the action would run 5 times, and it seems a bit excessive. What do you think?

KingHowler:
> @CheariX the time is necessary.
> 
> @KingHowler then the action itself would change its own cron schedule? It is feasible, but I think it is kind of overcharged. For example, if someone adds like 5 scheduled posts for the same day, but for different timestamps, the action would run 5 times, and it seems a bit excessive. What do you think?

Well yeah, it does seem excessive but that's only to ensure that each post is posted on the exact time it was said to be on.

Other than that we can just run it once every day at 23:59. I'm fine with either way as I don't use time in my website, only date.

KingHowler:
> @CheariX the time is necessary.
> 
> @KingHowler then the action itself would change its own cron schedule? It is feasible, but I think it is kind of overcharged. For example, if someone adds like 5 scheduled posts for the same day, but for different timestamps, the action would run 5 times, and it seems a bit excessive. What do you think?

Another thing I would like to mention is that this highly depends on the user. Like I said I don't use time but there may be some people who think it's crucial for their post to be published at the exact time.

How about we make a config.txt file?

We can build both versions. An action that runs daily and an action that runs at specific times. A third action, the scanner, will check the config.txt and run the appropriate action according to it.

This makes it customizable for the user.

george-gca:
I am more inclined into doing a single one that checks everyday (ignoring time) and add an explanation of how one would do that considering the timestamp in our FAQ or CUSTOMIZE, for example.

KingHowler:
> I am more inclined into doing a single one that checks everyday (ignoring time) and add an explanation of how one would do that considering the timestamp in our FAQ or CUSTOMIZE. What do you think?

The current code can fulfill those requirements, but as the site doesn't publish posts dated for a later time, it's essential to have the scheduler work exactly at 23:59.

Earlier you proposed that we should run it at 00:00 and 23:59. If we run it at 00:00, all files dated (ignoring time) will be sent to ```_post/``` but deploy won't upload them. When we run it at 23:59 later, all the files will already have been moved to ```_posts/``` and there will be nothing to push.

Without pushing the deploy action won't trigger and we will have to wait for the next scheduled post (that could be the next day, could be next month maybe even next year)

So running it twice a day will cause the scheduler to not work properly.

george-gca:
> The current code can fulfill those requirements, but as the site doesn't publish posts dated for a later time, it's essential to have the scheduler work exactly at 23:59.

We can simply ignore the time like you currently use, and give instructions for adding the time if someone wishes it to.

> Earlier you proposed that we should run it at 00:00 and 23:59. If we run it at 00:00, all files dated (ignoring time) will be sent to _post/ but deploy won't upload them. When we run it at 23:59 later, all the files will already have been moved to _posts/ and there will be nothing to push.

My mistake, I meant 12:00 and 23:59. That way we would have maybe 2 pushes per day. But it is ok for me doing once a day and gathering all posts for that day.

.github/workflows/posts-scheduler.yml
```diff
@@ -0,0 +1,40 @@
...
+  workflow_dispatch:
+
+jobs:
+  publish_posts:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout main branch
+        uses: actions/checkout@v4
+        with:
+          ref: main
```
george-gca, line 16: Can you change this for `master`? In al-folio our main branch is still named master, at least until #2086 is closed.

KingHowler:
> I believe I can merge this PR now. Would you mind sending another PR with the information about how one could add support for time in this action @KingHowler, maybe in CUSTOMIZE.md or FAQ.md? I believe it would be useful for some users.

Wait, let me set trigger time to 23:59 first

_scheduled/README.md
george-gca, line 1: Can you also add this to our `CUSTOMIZE.md`? Or maybe move this to `CUSTOMIZE.md` and on our `README.md` just cite that this feature is available and for more information point to this section in `CUSTOMIZE.md`.

george-gca (changes requested):

KingHowler:
@george-gca I have set the trigger time to 23:59 and will send a PR for CUSTOMIZE.md

Here's a few things I'd like you to know before you merge this

- This code completely ignores time given in the post, it makes up for it by running just before midnight
- This code runs exactly once every day at 23:59
- This code works with files using the following name pattern only: ```yyyy-mm-dd-title.md```
- It moves files from ```_scheduled/``` to ```_posts``` all other directories are unaffected by the action

george-gca:
I just added a few more things to fix.

_scheduled/README.md
KingHowler: I can add it to CUSTOMIZE.md, the reason I added this to a README.md inside ```_scheduled/``` is as a placeholder, to ensure that the folder is not removed from GitHub which removes empty folders

george-gca: But I believe this way we end up having information in places where people might not look for it.

I think it is best if you add this information to `CUSTOMIZE.md`, in our `README.md` just cite that this feature is available and for more information point to this section in `CUSTOMIZE.md`, and to keep this directory created (not sure if really needed, since we also don't create an empty directory for `_drafts/`), you could simply add an empty [.gitkeep](https://www.freecodecamp.org/news/what-is-gitkeep/) file.

But now thinking about it, I have another suggestion. Commit this workflow completed commented, since this is something I believe not many users will use, but it is useful and not straightforward to do it, and specify in the information added that this file exists and you can just uncomment it. What do you think?

@CheariX @pourmand1376 what do you think?

KingHowler: Instead of commenting the entire action, we can replace the ```.yml``` extension with ```.txt``` that way github won't recognize it as an action.

I am removing the ```README.md``` from the ```_scheduled``` folder. I will send another pull request adding the information regarding scheduler to the ```CUSTOMIZE.md```. Meanwhile you can merge this pull request

PR #2678, opened 2024-09-03
Fix conditional rendering of tag and category section

gsoykan:
### Overview
This PR fixes an issue where unnecessary horizontal lines were displayed when there were no tags or categories present. The tag and category container is now conditionally rendered, ensuring it only appears when there are tags or categories to display.

no tags meaning, in _config.yml
```
display_tags: []
display_categories: []
```

### Before and After
The difference is illustrated in the images below:
- **First Image (Fixed)**: Shows the correct behavior with no extra lines when tags or categories are absent.
- **Second Image (Current)**: Demonstrates the issue with unwanted horizontal lines appearing when no tags or categories are present.

image
image

### Impact
This change improves the visual consistency and cleanliness of the theme by preventing unnecessary elements from being rendered, particularly in cases where there are no tags or categories defined.

### DIFF
--- a/_pages/blog.md
+++ b/_pages/blog.md
@@ -29,7 +29,7 @@ pagination:
   </div>
   {% endif %}
 
-{% if site.display_tags or site.display_categories %}
+{% if site.display_tags.size > 0 or site.display_categories.size > 0 %}
 
   <div class="tag-category-list">
     <ul class="p-0 m-0">
--- a/_pages/blog.md
+++ b/_pages/blog.md
@@ -29,7 +29,7 @@ pagination:
   </div>
   {% endif %}
 
-{% if site.display_tags.size > 0 or site.display_categories.size > 0 %}
+{% if site.display_tags and site.display_tags.size > 0 or site.display_categories and site.display_categories.size > 0 %}
 
   <div class="tag-category-list">
     <ul class="p-0 m-0">


_pages/blog.md
```diff
@@ -29,7 +29,7 @@ pagination:
   </div>
   {% endif %}
 
-{% if site.display_tags or site.display_categories %}
+{% if site.display_tags.size > 0 or site.display_categories.size > 0 %}
```
george-gca, line 32: What if someone removes `display_tags` and `display_categories` from `_config.yml`? Maybe we should use both checks.

george-gca (changes requested):

_pages/blog.md
KingHowler: Wouldn't it be better to use the old check?
It doesn't make sense to have an empty ```display_tags``` or ```display_categories``` section. If you don't want to display any tags or categories then just remove the entire section

george-gca: I think the same, but I believe adding an empty list check wouldn't hurt.

KingHowler: true

gsoykan: Hey all,
Thanks for the feedback! 🙌 I’ve updated the conditional check as suggested to handle cases where display_tags and display_categories might be missing in _config.yml. The new condition now ensures both the existence and size of these fields are checked.

george-gca (approved):

PR #2688, opened 2024-09-09
Fixed "All contributors not showing on README.md"

KingHowler:
# In README.md
## All Contributors Section

**Out of the 216 contributors, the page only shows around 100**
By adding an additional parameter ```max``` It now shows all of them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -452,7 +452,7 @@ Our most active contributors are welcome to join the maintainers team. If you ar
 ### All Contributors
 
 <a href="https://contrib.rocks">
-  <img src="https://contrib.rocks/image?repo=alshedivat/al-folio" />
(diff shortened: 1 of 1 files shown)


george-gca:
I am currently not sure if we should add 200+ heads in our README (seems a bit excessive) or just change the header to `Top contributors`.

KingHowler:
> I am currently not sure if we should add 200+ heads in our README (seems a bit excessive) or just change the header to `Top contributors`.

yeah but the issue is, that all the contributors are not ordered by number of commits, I have 2 commits under my name, but instead of me people with only 1 commits are shown.

I totally agree with you on the fact that it is excessive, that's why we should move it to the end of the README.md.

PR #2689, opened 2024-09-10
Added new feature to allow user to display individual repositories in their Projects pages

KingHowler:
# New Feature
## The user can now display individual GitHub repositories in their Project Page

The theme changes according to the site theme

Light Mode
Light Mode

Dark Mode
Dark Mode

### DIFF
--- a/CUSTOMIZE.md
+++ b/CUSTOMIZE.md
@@ -66,6 +66,17 @@ If you want to create blog posts that are not ready to be published, but you wan
 
 You can create new projects by adding new Markdown files in the [\_projects](_projects/) directory. The easiest way to do this is to copy an existing project and modify it.
 
+You can also display your project's GitHub Repo Card, if it has any, on the project page.
+
+You simply need to add the following to your Markdown file's Frontmatter :
+```YAML
+repo:
+  owner_username: User #The username of the owner of the repository
+  repo_name: myRepo #The name of the repository
+  display_owner: false #Boolean for whether you want to show the username of the owner or not
+  max_lines_description: 3 #Integer telling the maximum number of lines the description of the repo should have in the card
+```
+
 ## Adding some news
 
 You can add news in the about page by adding new Markdown files in the [\_news](_news/) directory. There are currently two types of news: inline news and news with a link. News with a link take you to a new page while inline news are displayed directly in the about page. The easiest way to create yours is to copy an existing news and modify it.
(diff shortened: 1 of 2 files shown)


KingHowler:
@george-gca kindly review this one too

KingHowler:
@george-gca, I am having trouble with scaling the card automatically for each device size.

What are you doing to make these cards scale automatically?
image

george-gca:
> What are you doing to make these cards scale automatically?

Nothing special really. You can check the classes used here

https://github.com/alshedivat/al-folio/blob/b30b3f4ec0c3223366c06183b49bdb8f0a95664c/_pages/repositories.md?plain=1#L42

About this feature, not sure it would be a nice addition. I think this is more useful as a [shared idea](https://github.com/alshedivat/al-folio/discussions/categories/show-and-tell) on how to achieve this rather than adding it to the repository.

For example, what if the user wants to change the position where the repository card is displayed? Maybe in the end of the page, or in the middle between some paragraphs, after properly introducing the problem it is trying to solve? What if there is more than one repository linked to the project that the user wants to display? This adds too much variables and personal preferences that I think are out of the scope of the template as a basis for personal customization.

KingHowler:
Yeah I figured out the scaling issue and the how to add multiple repositories while maintaining the standard design of the website.
That only leaves the issue with the user wanting to display in the end. I've got an easy solution for that as well, we can add it to customizing.md

KingHowler:
@george-gca, how does it seem now?

george-gca:
I don't think this is too much of a burden for a user to add in their page, so I don't believe it should be added to the template. Like you said, simply by adding this to a page

```liquid
<div class="repositories d-flex flex-wrap flex-md-row flex-column justify-content-between align-items-center">
  {% include repository/repo.liquid repository=Username1/Repo1 %}
  {% include repository/repo.liquid repository=Username2/Repo2 %}
</div>
```

solves this. And this can be added straight to the project itself, it doesn't need to be in `_layouts/page.liquid`, which is a layout shared by other pages. By adding this directly into the project gives the user the freedom to specify where exactly it wants the card to be shown, instead of being limited by the position defined in the layout.

KingHowler:
Then I should remove the changes to the liquid file and modify the CUSTOMIZING.md accordingly, is that what you mean @george-gca?

KingHowler:
So that the user know how to add repository cards anywhere

KingHowler:
👍

Issue #2668, opened 2024-08-28
Publication preview image not enlarged on click 


WYF99:
### Have you checked that your issue isn't already filed?

- [X] I read through [FAQ](https://github.com/alshedivat/al-folio/blob/master/FAQ.md) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [X] Yes, I have checked that this issue isn't already filed.

### Bug description

Publication preview thumbnails are not enlarged and displayed at original size upon mouse click

### How to reproduce the bug

1. Go to any publication entry with a preview image which originally is of larger size than the thumbnail (e.g., any image from [my personal site](https://wangyi.fan/publications/)).
2. Click on the thumbnail to zoom in.
3. The image is centered but still displayed at the same small (and unreadable) size.

### Error messages and logs

No error message.

### What operating system are you using?

Not applicable (e.g. you're using GitHub Pages or other hosting)

### Where are you seeing the problem on?

Deployed site

### More info

A publication thumbnail is displayed at a reduced size by default, which is supposed to be centered, enlarged, and displayed at its original size upon click. In reality, the thumbnail is simply centered without any size change.

 According to @george-gca from #2662 
> imagemagick creates a reduced version of those images, since they are displayed in this location at a really small size, and simply doesn't use the original version.

jcblum:
I was struggling with the same thing, so I dug into it. I think this is not technically a bug, but instead the interaction of non-obvious expected behavior from two different plugins:

1. `jekyll-imagemagick` only scales images down (not up)
2. `medium-zoom` calculates the zoomed size based on what's in `sizes`, not the source image's actual dimensions

## `jekyll-imagemagick` only scales images down (not up)

<details>
<summary>All the details are in here…</summary>

It turns out that small original images aren't being resized at all, because `jekyll-imagemagick` is hard-coded to request that Image Magick shrink images, not enlarge them. This is a big part of why the tiny `gif` preview images in the example site content don't enlarge.

In the [jekyll-imagemagick source](https://gitlab.com/emmmile/jekyll-imagemagick/-/blob/bd15ef28d59fe3321e5e90e2756f4ff747089eb5/lib/jekyll-imagemagick/convert.rb#L26-36),  the line that constructs the Image Magick `-resize` command is:

> <code>cmd += "-resize \"#{long_edge}<mark>></mark>\" #{resize_flags} "</code>

[...]

PR #2692, opened 2024-09-12
Added lightbox2 functionality complete with an example (project_10)

Picsiri:


george-gca:
What is this for?

Picsiri:
What is a feature for? Anything you use it for.

I mostly needed it where I put 12 images in a row. With this you can actually see all of them. Also, in general I like if an image pops up when I click on it.

I am not sure about it but I guess imagemagick downscales the embedded images? In that case pop up does not enlarge the image but if you upload HD images...

george-gca:
> What is a feature for? Anything you use it for.

With this you assume we know what lightbox is for.

> I mostly needed it where I put 12 images in a row. With this you can actually see all of them. Also, in general I like if an image pops up when I click on it.

Do you have a working example or screenshots? In [this example](https://alshedivat.github.io/al-folio/blog/2015/images/) the zoomable images pops when you click on them.

> I am not sure about it but I guess imagemagick downscales the embedded images?

Not entirely sure, but I believe this is only currently happening on publication images like reported in #2668.

Picsiri:
Should I cancel the PR?

alshedivat:
just catching up here: what's the advantage of lightbox2 over zoomable images we already have?

george-gca:
Can you answer this @Picsiri?

Picsiri:
sorry, I don't usually check my GitHub ...

For example you can group images, there are more mini image setting options.

I made a pretty looking gallery with 6 images in a row that are unrecognizable on that scale but it is clearly indicated that they belong together and you can view them in a carousel.

_config.yml
george-gca, line 1: I think this should be set per page, since not all pages may include images. So this should be, for example, in the project itself.

_includes/head.liquid
```diff
@@ -46,6 +46,25 @@
...
 
+<!-- Add scripts/style for Lightbox2 -->
+{% if site.lightbox.enabled %}
+  <link
+    defer
+    rel="stylesheet"
+    type="text/css"
+    href="{{ '/assets/css/lightbox.min.css' | relative_url | bust_file_cache }}"
+  >
+  <script src="{{ '/assets/js/lightbox-plus-jquery.min.js' | relative_url | bust_file_cache }}"></script>
```
george-gca, line 57: Maybe this line and the next script tag should be included in [_includes/scripts/imageLayouts.liquid](https://github.com/alshedivat/al-folio/blob/main/_includes/scripts/imageLayouts.liquid) instead.

assets/js/lightbox-plus-jquery.min.js
george-gca, line 1: We don't need the version with jquery, since we already include jquery in our code. Also, I think it would be better to include it from [jsdelivr](https://www.jsdelivr.com/package/npm/lightbox2), like we do with a lot of our js dependencies. Check, for example, how we include Leaflet in [_includes/scripts/leaflet.liquid](https://github.com/alshedivat/al-folio/blob/main/_includes/scripts/leaflet.liquid) and [_config.yml](https://github.com/alshedivat/al-folio/blob/b0d70d4dd6db8920aaddb0c2d0abcab15b85dc78/_config.yml#L527)

george-gca (changes requested):

assets/js/lightbox-plus-jquery.min.js
Picsiri: Thanks for checking it out. I will try to understand and complete the instructions.

george-gca: Would you mind if I create this initial part, then you further help me customize it? I can create the base functionality, then you create the options part. This way you can also create a single file for other gallery options I just found out about.

george-gca: @Picsiri check #2877.

PR #2698, opened 2024-09-13
Fixed .webp src creation for svg and other files

### LINKED ISSUE #2660: Vector images not generating webp files

### Bug description

al-folio doesn't convert ".svg" files to ".webp" files. ".svg" files are small and their quality isn't lost when being displayed so not converting it is the best option. I should also mention that the images are displayed on the website without any faults. The issue which I want to bring forth is the error logs. The workflow "Check for broken links on site" produces an error log which should not be generated.

### How to reproduce the bug

- Add a ```.svg``` file in ```assets/img/```
- Include this image in a post
- Push changes and check the ```Check for broken links on site```

### Error messages and logs

```
✗ [ERR] file:///home/runner/work/kinghowler.github.io/kinghowler.github.io/_site/assets/img/dec/logic-gates/NOR.svg-800.webp | Failed: Cannot find file
```

Deployed site

### More info

I suggest to create an exception case to not search for converted .svg files

KingHowler:
Added a default srcset in case extension is other than the following:
- .jpg
- .jpeg
- .png
- .tiff
- .gif

fixed #2660

### DIFF
--- a/_includes/figure.liquid
+++ b/_includes/figure.liquid
@@ -1,4 +1,6 @@
-{% assign img_path = include.path | remove: '.jpg' | remove: '.jpeg' | remove: '.png' | remove: '.tiff' | remove: '.gif' %}
+{% assign parts = include.path | split: '.' %}
+{% assign img_path = parts.first %}
+{% assign ext = parts.last %}
 
 <figure
   {% if include.slot %}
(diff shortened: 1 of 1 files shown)


_includes/figure.liquid
```diff
@@ -1,4 +1,6 @@
-{% assign img_path = include.path | remove: '.jpg' | remove: '.jpeg' | remove: '.png' | remove: '.tiff' | remove: '.gif' %}
+{% assign parts = include.path | split: '.' %}
```
george-gca, line 1: This will lead to unexpected results if the image name have other `.` than the extension (e.g. `aaa.bbb.jpg`).

_includes/figure.liquid
```diff
@@ -14,13 +16,17 @@
     {% if site.imagemagick.enabled %}
       <source
         class="responsive-img-srcset"
-        srcset="{% for i in site.imagemagick.widths %}{{ img_path | relative_url }}-{{ i }}.webp {{i}}w,{% endfor %}"
+        {% if ext == '.jpg' or ext == '.jpeg' or ext == '.png' or ext == '.tiff' or ext == '.gif' %}
```
george-gca, line 19: Since the path was splitted with `.`, I believe that the `.` will not be part of the `ext` anymore. Have you tested this?

george-gca (changes requested):

_includes/figure.liquid
KingHowler: I have tested it, the code works as expected

_includes/figure.liquid
KingHowler: done

george-gca (approved):

george-gca:
Thanks for the fix. Did you find if this fixes the faulty `</source>` tags?

KingHowler:
Unfortunately they can't be fixed. It's not an issue with al-folio, but an issue with the language ```Liquid```

What we can do is use a GitHub action to correct it. I have already done that in the ```gh-pages formater``` workflow.

PR #2702, opened 2024-09-16
Created a new layout for GitHub projects

MUmarShahbaz:
# Preview
Screenshot 2024-09-16 205708
Screenshot 2024-09-16 205724
Screenshot 2024-09-16 205832
Screenshot 2024-09-16 205819

## Additions
1. Add repositories, type you repo name in the front matter as the following
```YAML
repositories:
  - KingHowler/Arduino-Language-Support
  - KingHowler/Oscilloscope-Online
  - alshedivat/al-folio
```
2. Add icons to your title, to show which languages and tools were used
```YAML
icons:
  - https://skillicons.dev/icons?i=processing
  - javascript/javascript-original.svg # Note: You don't have to specify the full link to devicons, only the foldername and filename
  - html5/html5-original.svg
```

### DIFF
--- a/_includes/icon.liquid
+++ b/_includes/icon.liquid
@@ -19,7 +19,7 @@
   {% if include.height %}
     height="{{ include.height }}"
   {% else %}
-    height="40"
+    height="2.5rem"
   {% endif %}
 >
 {% if include.spaceAfter %}
--- a/_includes/icon.liquid
+++ b/_includes/icon.liquid
@@ -1,0 +1,7 @@
+{% assign addIcon = include.icon %}
+{% if addIcon contains 'http' %}
+  <img src="{{ addIcon }}" height="40">
+{% else %}
+  <img src="https://cdn.jsdelivr.net/gh/devicons/devicon/icons/{{ addIcon }}" height="40">
+{% endif %}
+<img width="12">
--- a/_includes/icon.liquid
+++ b/_includes/icon.liquid
@@ -1,7 +1,27 @@
 {% assign addIcon = include.icon %}
-{% if addIcon contains 'http' %}
-  <img src="{{ addIcon }}" height="40">
-{% else %}
-  <img src="https://cdn.jsdelivr.net/gh/devicons/devicon/icons/{{ addIcon }}" height="40">
+<img
+  {% if addIcon contains 'http' %}
+    src="{{ addIcon }}"
+  {% else %}
+    src="https://cdn.jsdelivr.net/gh/devicons/devicon/icons/{{ addIcon }}"
+  {% endif %}
+  {% if include.alt %}
+    alt="{{ include.alt }}"
+  {% endif %}
+  {% if include.title %}
+    title="{{ include.title }}"
+  {% endif %}
+  {% if include.width %}
+    width="{{ include.width }}"
+  {% else %}
+    width="auto"
+  {% endif %}
+  {% if include.height %}
+    height="{{ include.height }}"
+  {% else %}
+    height="40"
+  {% endif %}
+>
+{% if include.spaceAfter %}
+  <img width="{{ include.width }}">
 {% endif %}
-<img width="12">
--- a/_layouts/github-project.liquid
+++ b/_layouts/github-project.liquid
@@ -11,10 +11,10 @@ layout: default
 <div class="post">
   <header class="post-header">
     <div>
-      <h1 class="post-title">
+      <h1 class="post-title d-flex flex-wrap flex-md-row flex-column justify-content-between align-items-center">
         {{ page.title }}
         {% if page.icons %}
-          <div style="display: inline-block; float : right;">
+          <div>
             {% for iconLink in page.icons %}
               {% include icon.liquid icon=iconLink %}
             {% endfor %}
--- a/_layouts/github-project.liquid
+++ b/_layouts/github-project.liquid
@@ -1,0 +1,49 @@
+---
+layout: default
+---
+{% if page._styles %}
+  <!-- Page/Post style -->
+  <style type="text/css">
+    {{ page._styles }}
+  </style>
+{% endif %}
+
+<div class="post">
+  <header class="post-header">
+    <div>
+      <h1 class="post-title">
+        {{ page.title }}
+        {% if page.icons %}
+          <div style="display: inline-block; float : right;">
+            {% for iconLink in page.icons %}
+              {% include icon.liquid icon=iconLink %}
+            {% endfor %}
+          </div>
+        {% endif %}
+      </h1>
+    </div>
+    <p class="post-description">{{ page.description }}</p>
+    {% if page.repositories %}
+      <div class="repositories d-flex flex-wrap flex-md-row flex-column justify-content-between align-items-center">
+        {% for repo in page.repositories %}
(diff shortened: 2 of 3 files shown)


KingHowler:
# Fixed layout issues for mobile phones
## Preview
51a87086-d065-489c-826e-d0aac984007c
b6acec2a-6b56-4a8b-8693-124b1f01e792
5fde2080-7703-48da-99cb-59b2fd5eb257
c5f4ca7a-8837-4cc9-8316-fb0b0e1b4aea

KingHowler:
@george-gca kindly review this as well

KingHowler:
I know this was previously discussed, but this is the updated version. I am using this for my [projects section](https://kinghowler.github.io/projects/) and it felt like a waste for it to not be a feature.

Next I am thinking about adding some default badges to include
for example I am using this in my about page :

#### Light Mode
image

#### Dark Mode
image

_includes/icon.liquid
```diff
@@ -0,0 +1,27 @@
...
+    width="auto"
+  {% endif %}
+  {% if include.height %}
+    height="{{ include.height }}"
+  {% else %}
+    height="40"
+  {% endif %}
+>
+{% if include.spaceAfter %}
+  <img width="{{ include.width }}">
```
george-gca, line 26: Is this a hack?

_includes/icon.liquid
```diff
@@ -0,0 +1,27 @@
+{% assign addIcon = include.icon %}
```
george-gca, line 1: Why not use `include.icon` directly in the code?

_includes/icon.liquid
```diff
@@ -0,0 +1,27 @@
...
+<img
+  {% if addIcon contains 'http' %}
+    src="{{ addIcon }}"
+  {% else %}
+    src="https://cdn.jsdelivr.net/gh/devicons/devicon/icons/{{ addIcon }}"
```
george-gca, line 6: I have a suggestion: what about creating another input to `icon.liquid`, for example `devicon`, and then set a precedence order. For example, use `devicon` if defined, else use `icon` as the src. This way we could even add other possibilities. Maybe also add [simple-icon](https://simpleicons.org/) as another parameter, which is the icon set used by [shields.io](https://shields.io/badges/static-badge).

george-gca:
> I was thinking about adding a little bit more like these so that they can make them easily? or should they do this themselves?

I believe they should add them themselves.

I am not sure about the layout of the icons in the desktop version, but tbh I don't currently have a counter-proposal. But I like the idea of this being a different layout available for the users. Can you create (or maybe modify) a project so it uses this layout, so it can be an example of usage?

KingHowler:
> > I was thinking about adding a little bit more like these so that they can make them easily? or should they do this themselves?
> 
> I believe they should add them themselves.
> 
> I am not sure about the layout of the icons in the desktop version, but tbh I don't currently have a counter-proposal. Have you tested it in the description as smaller icons maybe? But I like the idea of this being a different layout available for the users. Can you create (or maybe modify) a project so it uses this layout, so it can be an example of usage?

Then should I center align the title and add the icons below it? Like in the mobile version?

george-gca:
I think the current desktop version doesn't look very professional, but the mobile one looks nice. I agree to not center align it in the desktop version. What about putting it after the description and before the repo card?

KingHowler:
> I think the current desktop version doesn't look very professional, but the mobile one looks nice. I agree to not center align it in the desktop version. What about putting it after the description and before the repo card?

That was how I did it initially but it looked like I was just dumping everything to the left side

KingHowler:
This is the new design I have come up with
## PC Version
image

## Mobile Version
image

The icons and the repositories can be any number the user wills it to be

The ```{{ page.icon }}``` shown in the project card at the all projects page is displayed as an icon on the page

KingHowler:
If this one is better, should I add the description before or after the icons?
image

KingHowler:
I am also thinking about adding a divider right before the repositories, it will keep it separated from the center-aligned content

KingHowler:
I am using the design on my website, you can take a look directly if you want to
https://kinghowler.github.io/projects/

_includes/icon.liquid
KingHowler: Yeah, I have removed it now

_includes/icon.liquid
KingHowler: I was thinking to store them in variables so I can facilitate using them in the frontmatter and user calls, I found a workaround to that and now I am using include.icon directly

_layouts/github-project.liquid
```diff
@@ -0,0 +1,49 @@
...
+            {% for iconLink in page.icons %}
+              {% include icon.liquid icon=iconLink %}
+            {% endfor %}
+          </div>
+        {% endif %}
+      </h1>
+    </div>
+    <p class="post-description">{{ page.description }}</p>
+    {% if page.repositories %}
+      <div class="repositories d-flex flex-wrap flex-md-row flex-column justify-content-between align-items-center">
```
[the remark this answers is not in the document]
KingHowler: yes, I have done that in the new design. Waiting for approval.

george-gca:
Can you make it look like this?

image

I believe left aligned keeps the layout more similar to the other ones we are already using. If the user provides an image it could be shown in the left previous to the title like in this image, if not, it could be hidden and the title and everything would be left aligned.

Also I agree to adding a divider after the repo cards.

KingHowler:
I'll try to make it look like that

MUmarShahbaz:
Oh wow, I had my exams and completely forgot about this. I'll get working on this one.

MUmarShahbaz:
@george-gca I've edited according to your reqiurements.

# Preview
image

Requesting for a review. Also idk what the new 4 failed checks are, there were only 3 when I made the PR.

_includes/icon.liquid
```diff
@@ -0,0 +1,27 @@
+{% assign addIcon = include.icon %}
+<img
+  {% if addIcon contains 'http' %}
```
george-gca, line 3: It is best to search for `://` since it could include files from other sources (e.g. ftp://)

george-gca (changes requested):

george-gca:
2 more things:

1. Can you make that the image ends with the icons? In your example the icons are positioned at half the image height. Like this:

image

2. Can you create or modify one of the projects so that it showcases this new template? You can link to this repository as the code example, and even add the information you showed in [your previous comment](https://github.com/alshedivat/al-folio/pull/2702#issuecomment-2353465531) so users get to know how to do this.

MUmarShahbaz:
It's a bit different actually, It's not that the icons are placed at the middle of the image, the Entire Right section starts with the the top of the Image. I could add a simple padding-top to resolve it but I wanted to leave it so that we can use Multiline descriptions. If the description uses multiple line it will automatically shift the icons below.

Or I could have the right section be aligned to the bottom of the image.

george-gca:
Can't you instead make the image size relative to the right section, starting it and ending it with the section, but always scaling it? This way the image could be always squared and it would increase its size if the user added a large description.

MUmarShahbaz:
Interesting, I'll look into it

Issue #2682, opened 2024-09-05
Deployment Failure Due to Nokogiri Gem Incompatibility with Ruby 3.2.2


deepjyotk:
### Have you checked that your issue isn't already filed?

- [X] I read through [FAQ](https://github.com/alshedivat/al-folio/blob/master/FAQ.md) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [X] Yes, I have checked that this issue isn't already filed.

### Bug description

During deployment, the build process fails while attempting to install gems via Bundler. The failure occurs because the nokogiri gem (version 1.13.10-x86_64-linux) has a requirement of Ruby < 3.2.dev, >= 2.6, which is incompatible with the current Ruby version 3.2.2. As a result, the deployment is halted with an exit code 5.

### How to reproduce the bug

- Push the code to github

### Error messages and logs

<img width="1400" alt="image" src="https://github.com/user-attachments/assets/8cdabfcc-a2eb-40ae-b4c7-0a99f14a647a">

`Run ruby/setup-ruby@v1
Modifying PATH
Downloading Ruby
Extracting  Ruby
Print Ruby version
Installing Bundler
> bundle install
/opt/hostedtoolcache/Ruby/3.2.2/x64/bin/bundle config --local path /home/runner/work/deepjyotk.github.io/deepjyotk.github.io/vendor/bundle
/opt/hostedtoolcache/Ruby/3.2.2/x64/bin/bundle config --local deployment true
Cache key: setup-ruby-bundler-cache-v6-ubuntu-22.04-x64-ruby-3.2.2-wd-/home/runner/work/deepjyotk.github.io/deepjyotk.github.io-with--without--only--Gemfile.lock-0deeb3d141a9828c8aa70d1[5](https://github.com/deepjyotk/deepjyotk.github.io/actions/runs/10715535115/job/29711185717#step:3:6)dddc9183df3ab5dddf9[8](https://github.com/deepjyotk/deepjyotk.github.io/actions/runs/10715535115/job/29711185717#step:3:10)11b6a988bb4f3066ca73
/opt/hostedtoolcache/Ruby/3.2.2/x64/bin/bundle install --jobs 4
Fetching gem metadata from https://rubygems.org/.......
nokogiri-1.13.10-x86_64-linux requires ruby version < 3.2.dev, >= 2.6, which is
incompatible with the current version, 3.2.2
[...]

Issue #2544, opened 2024-07-04
Deployment failed with Gemfile.lock (created by local setup)


justinhou95:
### Have you checked that your issue isn't already filed?

- [X] I read through [FAQ](https://github.com/alshedivat/al-folio/blob/master/FAQ.md) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [X] Yes, I have checked that this issue isn't already filed.

### Bug description

Deployment failed with Gemfile.lock

### How to reproduce the bug

1. Local setup using Docker creates Gemfile.lock
2. Uploading Gemfile.lock causes error in deployment action

### How to solve it
4. Solved by adding Gemfile.lock into .gitignore

### Error messages and logs

```
bundle install
/opt/hostedtoolcache/Ruby/3.2.2/x64/bin/bundle config --local path /home/runner/work/justinhou95.github.io/justinhou95.github.io/vendor/bundle
/opt/hostedtoolcache/Ruby/3.2.2/x64/bin/bundle config --local deployment true
Cache key: setup-ruby-bundler-cache-v6-ubuntu-22.04-x64-ruby-3.2.2-wd-/home/runner/work/justinhou95.github.io/justinhou95.github.io-with--without--only--Gemfile.lock-03358d48efe98caee97dcb4e12a28f649[26](https://github.com/justinhou95/justinhou95.github.io/actions/runs/9794626471/job/27045011792#step:3:32)5ecc3d62e7fddd6bef7fe0b35c8ce
Cache Size: ~155 MB (162713749 B)
/usr/bin/tar -xf /home/runner/work/_temp/cd43a20f-23d2-49fb-ba6d-763e124ad1f6/cache.tzst -P -C /home/runner/work/justinhou95.github.io/justinhou95.github.io --use-compress-program unzstd
Received 16[27](https://github.com/justinhou95/justinhou95.github.io/actions/runs/9794626471/job/27045011792#step:3:33)13749 of 162713749 (100.0%), 155.0 MBs/sec
Cache restored successfully
[...]

PR #2705, opened 2024-09-18
Additional instructions for Apple Silicon Users

0xAdriaTorralba:
Trying to deal with building the website using Github Actions, I found a problem related to my local development build being an Apple Silicon Mac. 

I found a workaround for this, and I'm sharing and adding the step-by-step to fix the problem, in case you find suitable to include this for the rest of Apple Silicon Mac users :)

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -60,6 +60,32 @@ Note that when you run it for the first time, it will download a docker image of
 Now, feel free to customize the theme however you like (don't forget to change the name!). Also, your changes should be automatically rendered in real-time (or maybe after a few seconds).
 
 > Beta: You can also use the slimmed docker image with a size below 100MBs and exact same functionality. Just use `docker compose -f docker-compose-slim.yml up`
+### Important note for Apple Silicon Users
+
+Since the introduction of M-powered Macs, the target of local and production (remote) builds *may* be different. For this reason, Apple Silicon Macs may be doing additional steps in order to have the Github Actions to finish successfully. 
+
+If after you commit your changes to Github and try to build the website, you face an issue similar to this:
+
+```
+Your bundle only supports platforms ["aarch64-linux-gnu"] but your local
+platform is x86_64-linux. Add the current platform to the lockfile with
+`bundle lock --add-platform x86_64-linux` and try again.
+```
+
+Then, you must do the following:
+
+1. Connect to your docker container (`docker exec -it <container_name> sh`)
+2. Run the following commands *inside* the container
+    ```bash
+    bundle lock --add-platform ruby
+    bundle lock --add-platform x86_64-linux
+    ```
+3. Remove Gemfile.lock *outside* the container (`rm -rf Gemfile.lock`)
+4. Run `bundle install` *inside* the container.
+5. After that, you should have a new `Gemfile.lock` capable of handling builds for both local and production environments.
+6. Push the new `Gemfile.lock` and you should be good to go.
+
+Find the source for this fix [here](https://stackoverflow.com/questions/43429685/how-can-i-resolve-your-bundle-only-supports-platforms-x86-mingw32-but-your).
 
 ### Build your own docker image
 


george-gca:
Can you verify this @pourmand1376? I don't have access to a Mac.

pourmand1376:
Hey Guys. 

I tested on mac and I really couldn't figure out the issue. There is no issue here. 

Here's what I've done. 

```
docker compose pull
docker compose up
```

Is there any way to replicate your issue?

0xAdriaTorralba:
The issue is not related to local development, but when you try to deploy this website to Github Actions, see [here](https://github.com/0xAdriaTorralba/0xAdriaTorralba.github.io/actions/runs/10913865415) an instance of a failed build when you push your changes from local.

Again, local development is just fine using the `docker compose`.

george-gca:
Maybe this is related to #2682, #2544?

pourmand1376:
Can you explain how you updated your Gemfile.lock file and what was the reason?

0xAdriaTorralba:
> Maybe this is related to #2682, #2544?

Yes, it seems so. I got those issues before updaing the Gemfile.lock

> Can you explain how you updated your Gemfile.lock file and what was the reason?

The way I managed to update the Gemfile.lock is by doing the step-by-step I added on `INSTALL.md`, I paste this again here:

1. Connect to your docker container (`docker exec -it <container_name> sh`)
2. Run the following commands *inside* the container
    ```bash
    bundle lock --add-platform ruby
    bundle lock --add-platform x86_64-linux
    ```
3. Remove Gemfile.lock *outside* the container (`rm -rf Gemfile.lock`)
4. Run `bundle install` *inside* the container.
5. After that, you should have a new `Gemfile.lock` capable of handling builds for both local and production environments.
6. Push the new `Gemfile.lock` and you should be good to go.

The problem here is the following: the first time you run `docker compose pull; docker compose up` it is populated using the LOCAL development environment, so, it lists and adds the proper items to Gemfile.lock for THAT (local) platform (in my case, ARM-based ones). HOWEVER, it may be the case that the PRODUCTION environment architecture is DIFFERENT from your local one (this is my case, and also the cases on issues #2682, #2554). In this case, you need to rebundle the Gemfile.lock to consider this platform target. Is it clear now?

george-gca:
Somehow its Gemfile.lock is apparently being overwritten and commited to the repo, which I believe it shouldn't. We have [Gemfile.lock](https://github.com/alshedivat/al-folio/blob/master/Gemfile.lock) versioned here and also [added to the .gitignore](https://github.com/alshedivat/al-folio/blob/046545983f0864b62e883105955717cbf298561c/.gitignore#L9). Also if you look at the versioned Gemfile.lock we have

https://github.com/alshedivat/al-folio/blob/046545983f0864b62e883105955717cbf298561c/Gemfile.lock#L215

which is the version you just locked into.

pourmand1376:
@0xAdriaTorralba 

OK. Thanks for explanation. I think a better workaround is that you don't commit `Gemfile.lock` to the repository and it remains the same instead of changing it everytime. Does that solve your problem?

For that before any commit do this:
```
git restore Gemfile.lock
```

@george-gca And About `.gitignore`. Adding a file to gitignore won't work if we already have that file tracked in repository.

pourmand1376:
Thanks for mentioning that. I think we need to add another PR or change this PR somehow to make that change everytime. 

We should add this command somewhere to be run after finishing local deployment:
```
git restore Gemfile.lock
```

george-gca:
Maybe we should do [something](https://stackoverflow.com/a/13631525) like `git update-index --assume-unchanged` somewhere? Idk if it would solve the issue, since this flag can't be pushed to the repo.

pourmand1376:
I merged the change since I'm sure that line was problematic and I should've changed that since we've added `Gemfile.lock` into repository.

pourmand1376:
I merged my change. It should solve this problem.

pourmand1376:
This is not needed anymore.

PR #2707, opened 2024-09-19
Update entry_point.sh

pourmand1376:


george-gca:
Can you check if this fixes it @0xAdriaTorralba?

0xAdriaTorralba:
> Can you check if this fixes it @0xAdriaTorralba?

Will do ASAP.

0xAdriaTorralba:
> Can you check if this fixes it @0xAdriaTorralba?

Will do ASAP.

0xAdriaTorralba:
Tried the fix (reverted to the "old" `Gemfile.lock`) and copied the `entry_point.sh` proposed here and it [fails](https://github.com/0xAdriaTorralba/0xAdriaTorralba.github.io/actions/runs/11032169912/job/30640463715).

pourmand1376:
Thanks for testing it. 

if you want to test it, you have to replace the `docker-compose.yaml` too. This is because the `entry_point.sh` is built into image and will change only when a new image is pushed. 

```
# this file uses prebuilt image in dockerhub
services:
  jekyll:
    build: .
    ports:
      - 8080:8080
      - 35729:35729
    volumes:
      - .:/srv/jekyll
    environment:
      - JEKYLL_ENV=development
```

I think if you revert the Gemfile into the previous version and then commit. After that, you shouldn't have any problems when having local deployments.

pourmand1376:
I merged the change since I'm pretty much sure that line was problematic and I should've changed that since we've added `Gemfile.lock` into repository.

george-gca:
@0xAdriaTorralba can you test again in the same way? Now the docker image have the updated `entry_point.sh`.

pourmand1376:
Yep. Just make sure to pull new image.

Issue #2733, opened 2024-09-30
Version Issue from Docker Compose


jagar2:
### Have you checked that your issue isn't already filed?

- [X] I read through [FAQ](https://github.com/alshedivat/al-folio/blob/master/FAQ.md) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [X] Yes, I have checked that this issue isn't already filed.

### Bug description

Main Branch Does not build

### How to reproduce the bug

When I git clone the repo on a clean branch and run docker compose up the website does not build

### Error messages and logs

```
jekyll-1  | Gemfile.lock is tracked by git, keeping it intact
jekyll-1  | /usr/local/bundle/gems/bundler-2.5.7/lib/bundler/definition.rb:596:in `materialize': Could not find sass-embedded-1.79.3-x86_64-linux-gnu in locally installed gems (Bundler::GemNotFound)
jekyll-1  |     from /usr/local/bundle/gems/bundler-2.5.7/lib/bundler/definition.rb:203:in `specs'
jekyll-1  |     from /usr/local/bundle/gems/bundler-2.5.7/lib/bundler/definition.rb:270:in `specs_for'
jekyll-1  |     from /usr/local/bundle/gems/bundler-2.5.7/lib/bundler/runtime.rb:18:in `setup'
jekyll-1  |     from /usr/local/bundle/gems/bundler-2.5.7/lib/bundler.rb:162:in `setup'
jekyll-1  |     from /usr/local/bundle/gems/jekyll-4.3.4/lib/jekyll/plugin_manager.rb:52:in `require_from_bundler'
jekyll-1  |     from /usr/local/bundle/gems/jekyll-4.3.4/exe/jekyll:11:in `<top (required)>'
jekyll-1  |     from /usr/local/bundle/bin/jekyll:25:in `load'
jekyll-1  |     from /usr/local/bundle/bin/jekyll:25:in `<main>'
```

### What operating system are you using?

Linux

### Where are you seeing the problem on?

Running locally with Docker

dylanhogan:
I'm getting a similar error, also running locally with Docker on Linux:

```
jekyll-1  | Gemfile.lock is tracked by git, keeping it intact
[...]

PR #2740, opened 2024-10-02
Add Image Version to Docker-Compose to pull the relevant image

pourmand1376:
This is to address 
- #2733 

Since a lot of times, the only problem is that docker image is not consistent with the build. We have to somehow incorporate image version into repository. 

I don't insist to provide it this way. Maybe there are other automatic ways which are better. 

We can also calculate the relevant tag from git.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
 # this file uses prebuilt image in dockerhub
 services:
   jekyll:
-    image: amirpourmand/al-folio:latest
+    image: amirpourmand/al-folio:v0.12.0
     build: .
     ports:
       - 8080:8080


george-gca:
I don't think I understand the problem. The issue is that the version of the libraries in `Gemfile.lock` are different from the ones installed in the `Dockerfile`? If so, I think the easiest solution we could give is to every time we update the dependencies we also update the version of the docker image and change it where it is used.

Now, I am not sure if this is the problem. Currently our users can use 2 different docker images:
1. from https://github.com/alshedivat/al-folio/blob/c0d53e631630b19328f30f9e0da900ff1161eb27/docker-compose.yml#L4
2. from https://github.com/alshedivat/al-folio/blob/c0d53e631630b19328f30f9e0da900ff1161eb27/.devcontainer/devcontainer.json#L5

Since I believe most recent users simply use dev containers because it is the easiest way to start, the problem might be in that path, but this is still to be confirmed.

pourmand1376:
The issue is exactly what happened right now. 

The main branch has changes that docker image is corrupt. So, people should stick to the tagged version until the problem is solved. and we release a new version.

pourmand1376:
I applied this so that the current problem with docker image be solved via this approach.

PR #2750, opened 2024-10-05
fix: light-name-color -> color-name-light; added light variants if mi…

CheariX:
I found that some color names used a different naming scheme than other.

E.g. `$light-cyan-color` should be `$cyan-color-light` according to the other `light` and `dark` variants.
Also, I added light variants in case they were missing.

### DIFF
--- a/_sass/_variables.scss
+++ b/_sass/_variables.scss
@@ -8,20 +8,25 @@
 // Colors
 $red-color: #ff3636 !default;
 $red-color-dark: #b71c1c !default;
+$red-color-light: color.adjust($red-color, $lightness: 25%);
 $orange-color: #f29105 !default;
 $blue-color: #0076df !default;
 $blue-color-dark: #00369f !default;
+$blue-color-light: color.adjust($blue-color, $lightness: 25%);
 $cyan-color: #2698ba !default;
-$light-cyan-color: color.adjust($cyan-color, $lightness: 25%);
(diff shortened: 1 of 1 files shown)


george-gca:
I don't think they are exactly light variants, some of them might be just a `light cyan` instead of a `cyan for light mode`, for example.

Also if you changed the color name check if it isn't used anywhere. If it is, you should also change it there. If it isn't, why does it even exist?

CheariX:
> I don't think they are exactly light variants, some of them might be just a `light cyan` instead of a `cyan for light mode`, for example.

Oh, this is good to know. I did not know that the `light` variants are supposed to be `cyan for light mode` instead of `light cyan`. Thank you.

So this particular `light-cyan-color` is correctly named and used in a dark theme? Then we would need a `light-cyan-color-light` for the `light` theme?
Also, there are color names like `green-color-bright`, which sound to me like what this `cyan` should have been defined.

These namings are really confusing. 😃

> Also if you changed the color name check if it isn't used anywhere. If it is, you should also change it there. If it isn't, why does it even exist? So maybe we should delete it.

[I checked it before](https://github.com/search?q=repo%3Aalshedivat%2Fal-folio%20light-cyan-color&type=code) creating this PR but I could not find any references.

> 
> There will be one day that one shall look at all our sass code and clean/organize it.
> 

😀

TBH, I don't care too much about this particular PR. It was just something that I stumbled upon during my last master merge.
If you don't like it, feel free to close/reject it.

Issue #2757, opened 2024-10-09
Latex equations not rendering in a jupyter notebook


m-julian:
### Have you checked that your issue isn't already filed?

- [X] I read through [FAQ](https://github.com/alshedivat/al-folio/blob/master/FAQ.md) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [X] Yes, I have checked that this issue isn't already filed.

### Bug description

Latex equations not rendering in a jupyter notebook which is inserted with the jekyll-jupyter-notebook plugin

### How to reproduce the bug

Hello,

Recently I have tried to make a GitHub pages and use the [jekyll-jupyter-notebook](https://github.com/red-data-tools/jekyll-jupyter-notebook) plugin to insert jupyter notbooks into my website. I ran into the issue where Latex equations in the jupyter notebooks were not rendering. I opened up an issue in the jekyll-jupyter notebook repo already https://github.com/red-data-tools/jekyll-jupyter-notebook/issues/14 but the issue seems to be related to the `nbconvert` package that is in the Dockerfile.

I have made a test repo at https://github.com/m-julian/test_equations_jekyll-jupyter-notebook which reproduces the issue.

### Error messages and logs

No error message, but the notebook equations were not rendering because Mathjax was not being loaded correctly, see https://github.com/red-data-tools/jekyll-jupyter-notebook/issues/14 for details.

### What operating system are you using?

Mac

### Where are you seeing the problem on?

Running locally with Docker (docker compose)

PR #2758, opened 2024-10-09
update dockerfile to render jupyter notebook equations

m-julian:
Hello,

This pr should address the issue where the equations in jupter notebooks were not being rendered correctly in issue https://github.com/alshedivat/al-folio/issues/2757 .

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,11 +8,10 @@ RUN apt-get update -y && apt-get install -y --no-install-recommends \
     imagemagick \
     build-essential \
     zlib1g-dev \
-    pipx \
+    python3-pip \
     inotify-tools procps && \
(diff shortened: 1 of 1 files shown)


george-gca:
Thanks for finding a fix for this. Can you test it with [python3-pip](https://pkgs.org/download/python3-pip) package instead of [pipx](https://pkgs.org/download/pipx), and installing the packages with it? [pipx](https://pipx.pypa.io/stable/comparisons/) relies on pip (and venv), so using only `pip` is better.

m-julian:
Hi @george-gca, I've tested it on my end with pip and it works. I did have to add the `--break-system-packages` option because otherwise it does not allow installation of packages without making a virtual environment https://peps.python.org/pep-0668/.

george-gca (approved):

Issue #2774, opened 2024-10-14
Add support for apple-touch-icon


BenoitGaston:
### Have you checked that your feature request isn't already filed?

- [X] I read through [FAQ](https://github.com/alshedivat/al-folio/blob/master/FAQ.md) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my feature request.
- [X] Yes, I have checked that this feature request isn't already filed.

### Description & Motivation

Adding this line:` <link rel="apple-touch-icon" href=link_to_my_icon_asset />`  below the `shortcut icon` line inside `index.html` will make that: adding a shortcut on an iOS device will use the chosen icon instead of a generic one.

See iOS icon without this line on the left and with this line on the right:

![iOs icon](https://github.com/user-attachments/assets/4b1d3d06-0def-4101-afa0-5bc2e03c9d28)

The change is very simple.

george-gca:
I don't get it. Can you explain it further?

BenoitGaston:
Sure. First sorry for the poor wording that reflect my ignorance  of html.

When using Safari  on an iOS device, one has the option to  `Add a website icon to your Home Screen` (see [here](https://support.apple.com/guide/iphone/bookmark-a-website-iph42ab2f3a7/ios)). 

However, this icon needs to be referenced as  `apple-touch-icon` in the html file (see [here](https://webhint.io/docs/user-guide/hints/hint-apple-touch-icons/)). If this reference is missing an icon with the web site initial is created by iOS.

After building, and deploying my webpage (to my own domain not to gihub), I added the icon on iOS and  obtained the icon shared above on the left (with just my initial).

So what I did was to modify the file `_site/index.html` to add this line:

```
<link rel="apple-touch-icon" href="/assets/img/the_name_of_myshortcut_image.jpg>
```

below this line
```
<link rel="shortcut icon" href="/assets/img/the_name_of_myshortcut_image.jpg>
```

and deploy again to my own domain.

And now I am able to get the icon above on the right (with my short cut picture) when adding a shortcut to my iPhone home screen.

I hope this makes sense.

Best,

BenoitGaston:
Sorry if my question is naive but the only place in the code where I can see a reference to `shortcut icon` is [here](https://github.com/alshedivat/al-folio/blob/0af4aecf50426fc1e35696ca4c47940737175b26/_includes/head.liquid#L60).

[...]

PR #2783, opened 2024-10-17
apple-touch-icon support-issue-#2774

BenoitGaston:
Add apple-touch-icon to create iOS home screen shortcuts.

Issue https://github.com/alshedivat/al-folio/issues/2774


george-gca:
There is some piece of code still missing here.

Issue #2788, opened 2024-10-19
Failed to build on Netlify


HenryF23:
### Have you checked that your issue isn't already filed?

- [X] I read through [FAQ](https://github.com/alshedivat/al-folio/blob/master/FAQ.md) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [X] Yes, I have checked that this issue isn't already filed.

### Bug description

Failed to build on Netlify

### How to reproduce the bug

1. Clone the repository.
2. Set the environment variable RUBY_VERSION to 3.2.2 on Netlify.
3. Deploy on Netlify.

### Error messages and logs

```
3:30:20 AM: $ bundle exec jekyll build
3:30:21 AM: Failed during stage 'building site': Build script returned non-zero exit code: 2 (https://ntl.fyi/exit-code-2)
3:30:21 AM: /opt/build/cache/bundle/ruby/3.2.0/gems/mini_racer-0.15.0/lib/mini_racer.rb:22:in `load': /opt/build/cache/bundle/ruby/3.2.0/extensions/x86_64-linux/3.2.0/mini_racer-0.15.0/mini_racer_extension.so: undefined symbol: __libc_single_threaded (LoadError)
	from /opt/build/cache/bundle/ruby/3.2.0/gems/mini_racer-0.15.0/lib/mini_racer.rb:22:in `<top (required)>'
	from /opt/buildhome/.rvm/gems/ruby-3.2.2/gems/bundler-2.5.7/lib/bundler/runtime.rb:60:in `require'
	from /opt/buildhome/.rvm/gems/ruby-3.2.2/gems/bundler-2.5.7/lib/bundler/runtime.rb:60:in `block (2 levels) in require'
	from /opt/buildhome/.rvm/gems/ruby-3.2.2/gems/bundler-2.5.7/lib/bundler/runtime.rb:55:in `each'
	from /opt/buildhome/.rvm/gems/ruby-3.2.2/gems/bundler-2.5.7/lib/bundler/runtime.rb:55:in `block in require'
	from /opt/buildhome/.rvm/gems/ruby-3.2.2/gems/bundler-2.5.7/lib/bundler/runtime.rb:44:in `each'
	from /opt/buildhome/.rvm/gems/ruby-3.2.2/gems/bundler-2.5.7/lib/bundler/runtime.rb:44:in `require'
[...]

Issue #2432, opened 2024-05-21
Add support for deploys based on PRs to make it easier to test them


george-gca:
### Have you checked that your feature request isn't already filed?

- [X] I read through [FAQ](https://github.com/alshedivat/al-folio/blob/master/FAQ.md) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my feature request.
- [X] Yes, I have checked that this feature request isn't already filed.

### Description & Motivation

To have a deployed version of the site ready to be tested based on a PR would make it much simpler for us (maintainers) to validate them.

### Alternatives

I was able to make it work for Netlify only.

### Additional context

Suggested in #1981, made it work for a forked repo and shown in [#2415](https://github.com/alshedivat/al-folio/pull/2415#issuecomment-2118105841).

Steps needed:
1. Create an account in [Netlify](https://www.netlify.com/)
2. Add a [Deploy with Git](https://docs.netlify.com/site-deploys/create-deploys/#deploy-with-git) from this template (note: you should give access to Netlify to your repositories, but this can be limited by repo)
3. In the new deployed site, go to `Site configuration` -> `Build & deploy` -> `Continuous deployment` -> `Build settings`. Set the `Build command` as `sed -i "s/^\(baseurl: \).*$/baseurl:/" _config.yml && bundle exec jekyll build` and the `Publish directory` as `_site`. This can be seen in the screenshot below:
  
  ![image](https://github.com/alshedivat/al-folio/assets/31376482/29ea713c-d2e9-4a87-b666-e5f842e86670)
  
4. Enable [Deploy Previews](https://docs.netlify.com/site-deploys/deploy-previews/)

alshedivat:
@george-gca, that sounds like a great idea. let me set this up. also, appreciate all the features you've been shipping! 🚀

george-gca:
The best way to test this is to create a PR with a really obvious change and see if after a while the Netlify bot adds the link to the built page in the PR.

george-gca:
How is it going? Need any help?

george-gca:
Any updates on this @alshedivat?

CheariX:
Is there anything we can do to help with this?
Or is this something that only @alshedivat can do?

This feature would help so much with PRs!

george-gca:
There are changes that only the owner of the repo can do such as this one and #2086.

george-gca:
Can you take a look at this @alshedivat?

alshedivat:
[...]

PR #2791, opened 2024-10-21
Removed mini_racer dependency, updated dependencies and tabler-icons

george-gca:
Since `mini_racer` is causing issue in #2432 and #2788 (both related to Netlify) and [currently it doesn't seem we have a reason for having it](https://github.com/search?q=repo%3Aalshedivat%2Fal-folio%20mini_racer&type=code) as a dependency, I am removing it to see if it fixes the issues.


george-gca:
@alshedivat the Netlify preview just worked after I removed `mini_racer` dependency 🎉 

Do you have any idea why we had it in the first place?

pourmand1376:
Nope. I have no idea. 

I like the preview btw ;)

PR #2793, opened 2024-10-22
Renamed all references to master branch to main

george-gca:
Finishes #2086


CheariX:
Uff.

How are users supposed to apply these changes?
E.g. old forked templates still follow `master`, see e.g. https://github.com/CheariX/al-folio/
Also, if usered cloned the repo, do they automatically get a change in the name too?
How about deployment? Does it affect exting github pages deployment?

PR #2798, opened 2024-10-22
Update the INSTALL.md file to include steps for deploying al-folio on Netlify.

HenryF23:
Update CUSTOMIZE.md to include steps to deploy al-folio on Netlify.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -134,7 +134,7 @@ If you need to manually re-deploy your website to GitHub pages, go to Actions, c
 1. [Use this template -> Create a new repository](https://github.com/new?template_name=al-folio&template_owner=alshedivat).
 2. Netlify: **Add new site** -> **Import an existing project** -> **GitHub** and give Netlify access to the repository you just created.
 3. Netlify: In the deploy settings
-    - Set **Branch to deploy** to `master`
+    - Set **Branch to deploy** to `main`
     - **Base directory** is empty
     - Set **Build command** to `sed -i "s/^\(baseurl: \).*$/baseurl:/" _config.yml && bundle exec jekyll build`
     - Set **Publish directory** to `_site`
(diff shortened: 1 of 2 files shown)


george-gca (changes requested):
Sorry for the delay on this. Can you move this to `INSTALL.md`? More specifically, above `Deployment to another hosting server (non GitHub Pages)`. I think it makes more sense there. Also, 2 more changes:
1. change branch `master` to `main`, since we are not using master anymore
2. make it as an enumeration. I think it is easier to follow

HenryF23:
> Sorry for the delay on this. Can you move this to `INSTALL.md`? More specifically, above `Deployment to another hosting server (non GitHub Pages)`. I think it makes more sense there. Also, 2 more changes:
> 
> 1. change branch `master` to `main`, since we are not using master anymore
> 2. make it as an enumeration. I think it is easier to follow

I have moved the documentation to `INSTALL.md` and used enumeration for better readability. The branch should already be `main`

INSTALL.md
```diff
@@ -129,6 +129,25 @@ If you keep your site on another branch, open `.github/workflows/deploy.yml` **o
 
 If you need to manually re-deploy your website to GitHub pages, go to Actions, click "Deploy" in the left sidebar, then "Run workflow."
 
+### Deploy on [Netlify](https://www.netlify.com/)
+
+1. [Use this template -> Create a new repository](https://github.com/new?template_name=al-folio&template_owner=alshedivat).
+2. Netlify: **Add new site** -> **Import an existing project** -> **GitHub** and give Netlify access to the repository you just created.
+3. Netlify: In the deploy settings
+    - Set **Branch to deploy** to `master`
```
george-gca, line 137: The branch here should be `main`.

george-gca (changes requested):

INSTALL.md
HenryF23: > The branch here should be `main`.

I see. Thank you for pointing that out. It's now fixed.

george-gca (approved):

george-gca:
Thanks for your contribution.

Issue #2795, opened 2024-10-22
Problem with docker compose


angel-ayala:
### Have you checked that your issue isn't already filed?

- [X] I read through [FAQ](https://github.com/alshedivat/al-folio/blob/main/FAQ.md) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [X] Yes, I have checked that this issue isn't already filed.

### Bug description

I'm facing problems with the docker setup, when I initialize the container it says: Could not find a JavaScript runtime. I'm not sure if have nodejs on the system but if is a docker container all should be wrapped on it right?

### How to reproduce the bug

1. Download repository and enter to the repo's folder
2. `docker compose pull`
3. `docker compose up`

### Error messages and logs

```
[+] Running 1/0
 ✔ Container al-folio-jekyll-1  Created                                                                                        0.0s 
Attaching to jekyll-1
jekyll-1  | Resolving dependencies...
jekyll-1  | /usr/local/bundle/gems/execjs-2.9.1/lib/execjs/runtimes.rb:68:in `autodetect': Could not find a JavaScript runtime. See https://github.com/rails/execjs for a list of available runtimes. (ExecJS::RuntimeUnavailable)
jekyll-1  | 	from /usr/local/bundle/gems/execjs-2.9.1/lib/execjs.rb:5:in `<module:ExecJS>'
jekyll-1  | 	from /usr/local/bundle/gems/execjs-2.9.1/lib/execjs.rb:4:in `<top (required)>'
jekyll-1  | 	from /usr/local/lib/ruby/3.3.0/bundled_gems.rb:74:in `require'
jekyll-1  | 	from /usr/local/lib/ruby/3.3.0/bundled_gems.rb:74:in `block (2 levels) in replace_require'
jekyll-1  | 	from /usr/local/bundle/gems/uglifier-4.2.0/lib/uglifier.rb:5:in `<top (required)>'
jekyll-1  | 	from /usr/local/lib/ruby/3.3.0/bundled_gems.rb:74:in `require'
jekyll-1  | 	from /usr/local/lib/ruby/3.3.0/bundled_gems.rb:74:in `block (2 levels) in replace_require'
jekyll-1  | 	from /usr/local/bundle/gems/jekyll-minifier-0.1.10/lib/jekyll-minifier.rb:1:in `<top (required)>'
[...]

PR #2799, opened 2024-10-23
Fix docker compose issue

kartikey2001:
Fixed the `docker compose` issue, when trying to run the repository, locally.

#2795 

<hr>

<h3>To test these out: </h3>

Run:
`docker compose pull`
`docker compose build`
`docker compose up`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,5 @@
 FROM ruby:latest
+RUN apt-get update && apt-get install -y nodejs npm
 ENV DEBIAN_FRONTEND noninteractive
 
 Label MAINTAINER Amir Pourmand
@@ -38,5 +39,7 @@ RUN bundle install --no-cache
 EXPOSE 8080
 
 COPY bin/entry_point.sh /tmp/entry_point.sh
+RUN chmod +x /tmp/entry_point.sh
 
+ENTRYPOINT ["/bin/bash", "/tmp/entry_point.sh"]
 CMD ["/tmp/entry_point.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,7 +40,5 @@ RUN bundle install --no-cache
 EXPOSE 8080
 
 COPY bin/entry_point.sh /tmp/entry_point.sh
-RUN chmod +x /tmp/entry_point.sh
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -10,3 +10,4 @@ services:
       - .:/srv/jekyll
     environment:
       - JEKYLL_ENV=development
+    entrypoint: /tmp/entry_point.sh
```
pourmand1376, line 13: I don't think this line is needed.

bin/entry_point.sh
```diff
@@ -32,3 +35,5 @@ while true; do
         start_jekyll
     fi
 done
+
+exec "$@"
```
pourmand1376, line 39: What does this do?

Dockerfile
```diff
@@ -38,5 +39,7 @@ RUN bundle install --no-cache
 EXPOSE 8080
 
 COPY bin/entry_point.sh /tmp/entry_point.sh
+RUN chmod +x /tmp/entry_point.sh
 
+ENTRYPOINT ["/bin/bash", "/tmp/entry_point.sh"]
```
pourmand1376, line 44: Theses two lines shouldn't be needed in my opinion

Dockerfile
```diff
@@ -1,4 +1,5 @@
 FROM ruby:latest
+RUN apt-get update && apt-get install -y nodejs npm
```
pourmand1376, line 2: Let's add this to the line that is installing needed packages.

It would be like:
```
RUN apt-get update -y && apt-get install -y --no-install-recommends \
    locales \
    nodejs npm \
    imagemagick \
    build-essential \
    zlib1g-dev \
    python3-pip \
    inotify-tools procps && \
    apt-get clean && rm -rf /var/lib/apt/lists/* /var/cache/apt/archives/* && \
    pip install nbconvert --break-system-packages
```

george-gca:
I will paste that here so everyone (including me) who gets here knows what this is all about:

https://gist.github.com/mohanpedala/1e2ff5661761d3abd0385e8223e16425?permalink_comment_id=3935570

PR #2809, opened 2024-10-25
Upate docker-compose to be easier to use for permission bug

pourmand1376:
Hey @george-gca ,

I am thinking about what you've done in `docker-compose.yaml`. 

Does it solve the problem if we just do this change in docker-compose?

- https://stackoverflow.com/questions/48727548/how-to-configure-docker-compose-yml-to-up-a-container-as-root

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,15 +3,7 @@ services:
   jekyll:
     image: amirpourmand/al-folio:v0.12.1
     build: .
-    # uncomment these if you are having this issue with the build:
-    # /usr/local/bundle/gems/jekyll-4.3.4/lib/jekyll/site.rb:509:in `initialize': Permission denied @ rb_sysopen - /srv/jekyll/.jekyll-cache/.gitignore (Errno::EACCES)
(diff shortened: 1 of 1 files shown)


george-gca:
Sorry for the huge delay. I'll take a look at it.

george-gca:
No, it doesn't. Same error:

```
/usr/local/bundle/gems/jekyll-4.3.4/lib/jekyll/cache.rb:181:in `initialize': Permission denied @ rb_sysopen - /srv/jekyll/.jekyll-cache/Jekyll/Cache/Jekyll--Cache/b7/9606fb3afea5bd1609ed40b622142f1c98125abcfe89a76a661b0e8e343910 (Errno::EACCES)
```

pourmand1376:
How can I replicate this?

george-gca:
Tbh I have no idea. I tried to use it on my laboratory server, where we use docker to run experiments. The docker installation there have some restrictions, to avoid someone being able to run code inside docker as root. And I have no idea how they did this.

Issue #2825, opened 2024-11-04
Ability to Add Resume/CV With Link


pranjalagg:
### Have you checked that your feature request isn't already filed?

- [X] I read through [FAQ](https://github.com/alshedivat/al-folio/blob/main/FAQ.md) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my feature request.
- [X] Yes, I have checked that this feature request isn't already filed.

### Description & Motivation

It would a good addition if there is a way in which if I add a link to my resume in the `cv.md` file, the cv page on the website automatically redirects to the link. And if I have a pdf file in the `assests` directory, it points to that.

### Pitch

People might have different ways of maintaining their resumes/CVs. I like to keep my different versions in the cloud. So when there is an update to my resume and I have to update it on my website, I need to download the pdf and update it in the `assets/pdf` directory. I find it more convenient to just update the link of my resume in the `cv.md` file (since I already have them in the cloud with shareable links).

### Alternatives

I think the solution to this should be a straightforward if/else condition. I can work on it and raise a PR.

Edit: Just raised a PR that solves this (#2826)

PR #2826, opened 2024-11-04
Support for Resume/CV Links

pranjalagg:
Solves #2825

### DIFF
--- a/_layouts/cv.liquid
+++ b/_layouts/cv.liquid
@@ -57,7 +57,7 @@ layout: default
             {% if page.cv_pdf contains '://' %}
               href="{{ page.cv_pdf }}"
             {% else %}
-              href="{{ page.cv_pdf | prepend: 'assets/pdf/' | relative_url}}"
+              href="{{ page.cv_pdf | prepend: 'assets/pdf/' | relative_url }}"
             {% endif %}
             target="_blank"
             rel="noopener noreferrer"
--- a/_layouts/cv.liquid
+++ b/_layouts/cv.liquid
@@ -53,12 +53,13 @@ layout: default
       <h1 class="post-title">
         {{ page.title }}
         {% if page.cv_pdf %}
-          {% if page.cv_pdf contains "http" %}
+          {% if page.cv_pdf contains 'http' %}
             <a
               href="{{ page.cv_pdf }}"
               target="_blank"
               rel="noopener noreferrer"
-              class="float-right">
+              class="float-right"
+            >
               <i class="fa-solid fa-file-pdf"></i>
             </a>
           {% else %}
@@ -66,7 +67,8 @@ layout: default
               href="{{ page.cv_pdf | prepend: 'assets/pdf/' | relative_url}}"
               target="_blank"
               rel="noopener noreferrer"
-              class="float-right">
+              class="float-right"
+            >
               <i class="fa-solid fa-file-pdf"></i>
             </a>
           {% endif %}
--- a/_layouts/cv.liquid
+++ b/_layouts/cv.liquid
@@ -53,13 +53,23 @@ layout: default
       <h1 class="post-title">
         {{ page.title }}
         {% if page.cv_pdf %}
-          <a
-            href="{{ page.cv_pdf | prepend: 'assets/pdf/' | relative_url}}"
-            target="_blank"
-            rel="noopener noreferrer"
-            class="float-right"
-            ><i class="fa-solid fa-file-pdf"></i
-          ></a>
+          {% if page.cv_pdf contains "http" %}
+            <a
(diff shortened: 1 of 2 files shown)


_pages/cv.md
```diff
@@ -4,7 +4,7 @@ permalink: /cv/
 title: cv
 nav: true
 nav_order: 5
-cv_pdf: example_pdf.pdf
+cv_pdf: http://alberteinstein.eurocv.eu/
```
george-gca, line 7: I think we should keep the local pdf file on the template.

_layouts/cv.liquid
george-gca, line 1: Can you make the if else only for the `href`? Since the rest of the `a` tag is the same.

george-gca (changes requested):

_layouts/cv.liquid
pranjalagg: Sure.

_pages/cv.md
pranjalagg: You're probably right. I just wanted to include it so that users know that there is an option to add links for resume as well.
Maybe we can add a comment to indicate that?

_layouts/cv.liquid
```diff
@@ -54,12 +54,17 @@ layout: default
         {{ page.title }}
         {% if page.cv_pdf %}
           <a
-            href="{{ page.cv_pdf | prepend: 'assets/pdf/' | relative_url}}"
+            {% if page.cv_pdf contains 'http' %}
```
george-gca, line 57: Maybe it is best to search for `://` instead of `http`, since it is possible to have the http string, and also to support, for example, storing the cv file in a `ftp://`.

george-gca (changes requested):

_layouts/cv.liquid
pranjalagg: Makes sense.

_pages/cv.md
george-gca: I agree.

_layouts/cv.liquid
```diff
@@ -54,12 +54,17 @@ layout: default
         {{ page.title }}
         {% if page.cv_pdf %}
           <a
-            href="{{ page.cv_pdf | prepend: 'assets/pdf/' | relative_url}}"
+            {% if page.cv_pdf contains '://' %}
+              href="{{ page.cv_pdf }}"
+            {% else %}
+              href="{{ page.cv_pdf | prepend: 'assets/pdf/' | relative_url}}"
```
george-gca, line 60: Just one more thing, can you add a space between `relative_url` and `}}`?

george-gca (changes requested):

_layouts/cv.liquid
pranjalagg: Sure, done!

george-gca (approved):

pranjalagg:
Hey @george-gca,
I see the Prettier check failing, but when I run `npx prettier . --check` or `npx prettier . --write` locally, I don't get any issues in any file. Any idea what might be the problem here?

george-gca:
Unfortunately the prettier check is not doing what it was supposed to. It was supposed to add here information about why it failed, but it is not doing it. You can still check it manually though.

Just click on the [failed action](https://github.com/alshedivat/al-folio/actions/runs/12094429501/job/33734615376?pr=2826) above, expand the `Run peter-evans/repository-dispatch@v2` line and download the generated file:

image

This is an html diff of what changes must be made. It is a zip file because GitHub doesn't allow to store anything else.

pranjalagg:
Thanks for that @george-gca,
I downloaded the HTML diff and checked it. It suggests changes to a file that I didn't even touch in this PR. Moreover, the change it suggests have already implemented in my branch, so not sure what the issue is.
On further investigation, I saw the Prettier check being failed on the main branch as well, and the HTML diff also suggests the same change (pasted below).
image

_pages/cv.md
george-gca: Just missing this point here @pranjalagg. Also I just fixed the prettier issue on main. Rebase it and it should be fine now.

pranjalagg: Done.

_pages/cv.md
```diff
@@ -4,7 +4,7 @@ permalink: /cv/
 title: cv
 nav: true
 nav_order: 5
-cv_pdf: example_pdf.pdf
+cv_pdf: example_pdf.pdf # you can replace it with a link to your resume as well
```
george-gca, line 7: I think it should be more explicit, like `you can also use external links here`.

george-gca (changes requested):

george-gca (approved):

george-gca:
Thanks for your contribution.

PR #2856, opened 2024-11-25
Bluesky URL prefix should not be appended

gdalle:
This PR modifies

- `_config.yml` to rename `bluesky_url` into `bluesky_id`
- `_includes/social.liquid` to handle the renaming in the display of social media icons

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -73,7 +73,7 @@ og_image: # The site-wide (default for all links) Open Graph preview image
 
 acm_id: # your dl.acm.org/profile/id
 blogger_url: # your blogger URL
-bluesky_url: # your bluesky id (like yourname.bsky.social)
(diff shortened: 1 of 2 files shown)


george-gca:
Actually this is intended. [Bluesky uses](https://docs.bsky.app/docs/advanced-guides/atproto) the [AT Protocol](https://atproto.com/), which is a decentralized social media network. This means that it is possible to have different services implementing this, like it is with Mastodon, and consequently urls other than `https://bsky.app/profile/` for your profile.

The correct change it would be to remove this part from `_includes/social.liquid` and put the whole url in `_config.yml`. Can you do that?

gdalle:
Sure, that works too! I just wanted to fix the current state, which was slightly confusing

gdalle:
I made the appropriate change and renamed the PR

george-gca (approved):

PR #2920, opened 2024-12-29
Revert padding from #2617 for CV tables

sim642:
PR #2617 added a noticeable amount of right padding to all table cells. This might be fine for tables in content (e.g. posts), but creates odd spaces in the CV page, which is largely made up of tables.


george-gca (approved):

george-gca:
Thanks for fixing this.

PR #2932, opened 2025-01-01
Fix Line Ending Issues for Shell Scripts in Docker for Windows 10

manuelsh:
Shell scripts like entry_point.sh caused errors in Docker due to CRLF line endings on Windows (`exec /tmp/entry_point.sh: no such file or directory`).

Added `.gitattributes` to enforce LF line endings for *.sh files.

This resolves the issue and ensures consistent behavior across platforms.

### DIFF
--- a/.gitattibutes
+++ b/.gitattibutes
@@ -1,0 +1,2 @@
+# Force LF line endings, needed for Docker to work on Windows
+*.sh text eol=lf


manuelsh:
@george-gca this change fixes an issue with the docker starting in windows.

george-gca (approved):

george-gca:
I didn't know about this feature. Thanks for fixing this. I had no idea this was still happening on our side. Is the file with the right line feed in the repo?

manuelsh:
The file has the right ending but when it is cloned in windows 10 and there is not this .gitattributes file with the settings added, it will change the line ending and it won't work. I believe you mentioned somewhere that the docker image was not working well in windows 10.

I have tested it already in my forked repo and it works, so it should be fine. Once it is merged, I am happy to test it from scratch (clone and generate the image).

PR #2945, opened 2025-01-08
Remove internal copy of jekyll-terser

alshedivat:
# Summary

since https://github.com/RobertoJBeltran/jekyll-terser/pull/1 has been merged, I'm removing the internal copy jekyll-terser and switching to https://github.com/RobertoJBeltran/jekyll-terser instead. also, refactored `Gemfile` a bit.


george-gca:
Oh yes, I didn't remove it because `jekyll-terser` is not yet on rubygems. I even opened an [issue there](https://github.com/RobertoJBeltran/jekyll-terser/issues/2) about it. But using it from git is a nicer solution.

PR #2956, opened 2025-01-15
Fix issue causing no color inversion when rendering equations

### LINKED ISSUE #2915: Color: distill-style in-line math in dark mode remains black

- [X] I read through [FAQ](https://github.com/alshedivat/al-folio/blob/main/FAQ.md) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.

### Bug description

With dark mode on, the inline math with `$...$` of a distill-style blog post remains black, but should be white.

### How to reproduce the bug

The example distill-style post on the example site can illustrate this. 

1. Go to [the example post, "Equation" section](https://alshedivat.github.io/al-folio/blog/2021/distill/#equations). 
2. In paragraph 2, the equation rendered by $E = mc^2$ is black under dark mode.

### Error messages and logs

```
```
image

Mac

Running locally with Docker (docker compose), Deployed site

### More info

@george-gca suggested that this happens because the distill template for this case with `$...$` imports another css, which overwrites the color part. You can see this happening with the browser debug.

Weirdly enough, adding
```
span.katex {
  color: var(--global-text-color) !important;
}
```
to `_sass/_distill.scss` doesn't solve this.

khamkarsuraj:
## Problem Description

When dark mode is enabled, inline math expressions rendered using `$...$` in distill-style blog posts are displayed in black text. This makes them difficult to read against the dark background. The issue occurs because the default styles for MathJax-rendered content do not account for dark mode and set the text color to black.

## Proposed Solution

This PR updates the MathJax setup to dynamically apply inline styles to ensure the text color of MathJax-rendered content inherits the parent container’s color. Specifically:
- Added a custom action in the _MathJax.options.renderActions_ section to inject a CSS rule that sets _.mjx-container_ to inherit its color from its parent element.
- Ensures that inline math expressions render correctly in both light and dark modes without explicitly setting a fixed color.

## Changes Made
Modified the _mathjax-setup.js_ file to include:
- Custom inline styles injected via JavaScript for _.mjx-container_ elements.
- Explicit handling of _inlineMath_ delimiters to ensure consistency in rendering.

## Testing and Validation
- Tested the fix in both light and dark modes to confirm that inline math expressions render in the correct color.
- Verified that the changes do not introduce regressions in light mode or affect block math expressions rendered using `$$...$$`.

## Impact
- Improves readability of inline math expressions in dark mode.
- Provides a consistent user experience for distill-style blog posts across light and dark themes.
- Resolves #2915

### DIFF
--- a/assets/js/mathjax-setup.js
+++ b/assets/js/mathjax-setup.js
@@ -1,5 +1,26 @@
 window.MathJax = {
   tex: {
     tags: "ams",
+    inlineMath: [
+      ["$", "$"],
+      ["\\(", "\\)"],
+    ],
+  },
+  options: {
+    renderActions: {
+      addCss: [
+        200,
+        function (doc) {
+          const style = document.createElement("style");
+          style.innerHTML = `
+          .mjx-container {
+            color: inherit;
+          }
+        `;
+          document.head.appendChild(style);
+        },
+        "",
+      ],
+    },
   },
 };


george-gca (approved):

george-gca:
Thanks for the fix.

PR #2964, opened 2025-01-17
Typo in jekyll-scholar plugin

chloevoyer:
I suspect this is a typo and the plugin should be jekyll-scholar and not jekyll/scholar?


chloevoyer:
Seems like I might have been wrong and this isn't a typo, it just doesn't follow the formatting for other plugins. Sorry bout that!

PR #2968, opened 2025-01-19
Upgrade al-folio to v1.0

alshedivat:
## Overview

This PR delivers the `v1.0` architecture and runtime upgrade for `al-folio`.

`al-folio` moves from a monolithic theme repo to a **thin starter + plugin ecosystem** model, while preserving visual identity and core behavior.

### 1) Starter becomes orchestration-only

`al-folio` is now responsible for:

- starter wiring (`Gemfile`, `_config.yml`)
- sample/demo content
- user-facing docs
- visual regression and cross-plugin integration tests

Component/runtime ownership is moved to dedicated gems.

### 2) Runtime and feature ownership split into plugins

`al_folio_core` now provides shared runtime/layout primitives and contracts, and delegates feature modules to owning gems.

Key plugin ownership in v1:

- `al_folio_core`: core theme runtime/contracts
- `al_folio_distill`: Distill layouts/runtime
- `al_folio_cv`: CV rendering
- `al_search`: search runtime and modal behavior
- `al_icons`: icon runtime (Font Awesome, Academicons, Scholar Icons)
- `al_cookie`: cookie-consent runtime
- `al_img_tools`: image/gallery/lightbox tooling
- `al_math`: math/TikZ runtime integration
- additional `al-*` plugins for analytics/charts/citations/comments/newsletter/external posts/upgrade tooling

### 3) Styling/runtime modernization

- Tailwind-first v1 runtime is established.
- Legacy Bootstrap behavior is isolated behind optional `al_folio_bootstrap_compat` for migration windows.
- jQuery-dependent runtime paths are removed/replaced in plugin-owned implementations.

### 4) Test ownership policy is enforced

- `al-folio` keeps only starter-level integration and visual integrity checks.
- Component correctness tests are owned by the corresponding plugin repos.

This mirrors runtime ownership and reduces duplication/drift.

### 5) Upgrade and migration workflow

`al_folio_upgrade` is integrated for audit/apply/report migration flows, including:

- config contract checks
- ownership drift detection
- local override drift audit/diff/acknowledgement workflow
- legacy marker detection
- plugin wiring validation

### 6) Documentation and contributor model for v1

Docs are updated to reflect modular v1 behavior and contribution routing.

New governance additions include:

- plugin ownership/boundary guidance (`docs/BOUNDARIES.md`)
- plugin naming convention (theme-coupled vs reusable plugin tracks)
- featured-vs-bundled plugin distinction
- plugin catalog source + page (`_data/featured_plugins.yml`, `_pages/plugins.md`)
- plugin feature proposal issue template
- PR template updates for ownership routing
- agent-facing setup and migration skills

## Compatibility and migration notes

- Plugin integration in starter is via `Gemfile` + `_config.yml`.
- `al-folio` starter intentionally does not use a gemspec in v1.
- Transitional Bootstrap compatibility remains available via `al_folio_bootstrap_compat`.
- Existing front-matter contracts are preserved where possible; runtime implementations are now plugin-owned.
- Customized sites can track local override drift with `bundle exec al-folio upgrade overrides audit` and `.al-folio-overrides.yml`.

## Validation strategy in this PR

- visual parity checks against the v0.16.3 baseline
- starter integration suites (comments, plugin toggles, distill, compat, upgrade CLI)
- style contract checks and repository linting

## Scope intent

This is the foundational v1 architecture PR.

Post-merge feature work should generally target owning plugin repos first, with `al-folio` limited to starter wiring, docs, examples, and integration/visual validation.

_Written by Codex on behalf of @alshedivat._


george-gca:
I decided to already move my contributions to this PR to the main repository, since it will already make it a little simpler. Also, I think the external-posts and analytics would be useful for more jekyll users, so maybe they should be created as `jekyll-` plugins?

alshedivat:
agreed. instead of having these changes stuck in this dev branch, best to merge them into main. will see if I can find time to do it this weekend.

george-gca:
@alshedivat I saw that you started support for Tailwind in #3509. Think you can also finish #3481? I started a test with GitHub Copilot in #3481 and one with Jules in #3482 to see which one could better handle updating our dependencies, but I haven't finished it. I decided to go with the Copilot one, but I am still iterating/testing the solution. There are still some layout differences, and a thorough test is needed for the blog posts.

alshedivat:
@george-gca, thanks for experimenting with copilot and jules for bootstrap upgrade.

i'm envisioning `v1.0` as a modularization upgrade of `al-folio`, and potentially a revamp of all the underlying styles. migrating to tailwind would be ideal b/c it's more flexible and less opinionated about visual appearance compared to bootstrap. so it's easier to exactly reproduce al-folio's visual appearance with tailwind. that said, migrating from bootstrap to tailwind is a pretty substantial change, so my PR is an experiment to see if the latest version codex is able to execute such migration while fully preserving visual appearance and functionality.

alshedivat:
Addressed latest review + CI feedback in commits 8828f478 and 596763a7:\n\n- Increased visual regression timeout budget to 20 minutes.\n- Updated parity helper to capture full-page screenshots by default for parity checks.\n- Added a safe fallback for very long pages to avoid Playwright's 32767px screenshot limit.\n- Kept distill baseline parity on viewport capture (to avoid long-page image hard limit while preserving the dedicated distill stability checks).\n- Hardened broken-link checks with retries and replaced the flaky demo redirect in _projects/3_project.md.\n\nCurrent status: required CI checks are green on the latest head.

george-gca:
I will try to find some time this week to properly review this. After all, it is a HUGE change.

dfuchss:
Hi @alshedivat, hi @george-gca, from my perspective, it may be interesting to address this question [here](https://github.com/alshedivat/al-folio/discussions/3545) in the FAQs or somewhere as part of v1.0 :)

Most people that I know and that are using al-folio have their own tweaks / customizations; so far this was straightforward to keep up-to-date with upstream - thanks to git :)
With the upcoming changes, I guess, this get's rather tricky (even when you use subtrees - the solution I mentioned in the discussions). Maybe you have a better solution/idea for those people. Perhaps it would even be an option to have a de-modularized version of the repository that simply contains the subtrees and aggregates the changes via CI (so basically an artificial, full-blown al-folio that is read-only and fetches upstream changes from all al-folio-modules) .. That was an idea I had 10 minutes ago :)

alshedivat:
thank you, @dfuchss, this is a great question. just wanted to note here that this PR won't be merged until everything is fully battle-tested and all important use cases are addressed. we'll certainly need to make sure upgrading + retaining existing customizations in most cases should be possible. (switch from bootstrap to tailwind will make this a bit more challenging, but the current design of v1.0 supports bootstrap backward compatibility.)

alshedivat:
Pushed the refreshed `v1.0-dev` head at `a721f324` after the migration rehearsal and local validation pass.

What changed:

- Merged current `origin/main` into `v1.0-dev`.
- Removed the tracked generated `output/playwright/**` artifacts and added ignore coverage for `output/playwright/` plus `test-results/`.
- Added the real migration rehearsal log for `dfuchss/fuchss.org`: https://github.com/alshedivat/al-folio/blob/v1.0-dev/MIGRATION_REHEARSAL_FUCHSS_ORG.md
- Updated migration/customization guidance in `FAQ.md`, `INSTALL.md`, and `CUSTOMIZE.md`, including local layout/include overrides, plugin pinning/fork guidance, Docker release behavior, and a v1 path ownership table.

Rehearsal result:

- Used a disposable local migration copy of `dfuchss/fuchss.org`.
- Initial `al-folio upgrade audit --no-fail`: 4 blocking issues, all caused by stale local copies of legacy runtime includes (`_includes/head.liquid`, `_includes/scripts.liquid`).
- After removing stale plugin-owned runtime files and old helper plugins/assets while keeping site-owned overrides: 0 blocking audit issues.
- Migrated site built successfully; checked key generated pages including home, CV, publications, repositories, projects, and Matrix `.well-known` pages.

Local validation completed:

- `npm ci`
- `npm run lint:prettier`
- `npm run lint:style-contract`
- `bundle exec jekyll build --baseurl /al-folio`
- `bash test/integration_comments.sh`
- `bash test/integration_plugin_toggles.sh`
- `bash test/integration_distill.sh`
- `bash test/integration_bootstrap_compat.sh`
- `bash test/integration_upgrade_cli.sh`
- `npx playwright install chromium webkit`
- `npm run test:visual` (`29 passed`, `25 skipped` baseline-gated tests)
- Docker Compose smoke: build/start succeeded, `http://localhost:8080/al-folio/` returned HTTP 200, then the container was stopped.

Fresh GitHub checks are now running on the pushed head.

_Written by Codex on behalf of @alshedivat._

alshedivat:
Follow-up after the migration rehearsal/docs push:

- Removed two unstable sample-post outbound links that were blocking `link-checker` on fresh CI:
  - stale MathJax comparison URL returning 403
  - Lightbox2 project URL returning 429
- Re-ran local hygiene for the follow-up edits: `git diff --check` and `npm run lint:prettier`.
- Pushed `4e381a79`; the refreshed PR check suite is now green. Netlify header/page/redirect checks are skipped as expected, and all GitHub Actions jobs pass.

_Written by Codex on behalf of @alshedivat._

alshedivat:
Final release-prep update before merge:

- Opened the real customized-site migration rehearsal PR: https://github.com/dfuchss/fuchss.org/pull/1
- Published `al_folio_upgrade` 1.0.2 with broader drift detection for plugin-owned local paths.
- Published `al_folio_core` 1.0.9 after the final starter build exposed a Jupyter notebook / `jekyll-minifier` interrupted-write edge case; the starter now pins this release.
- Reorganized long-form docs under `docs/`, updated agent/Copilot/Claude/Gemini guidance for v1 plugin ownership, and removed the local fuchss rehearsal note from this branch.
- Added weekly grouped Dependabot coverage across the plugin repos.

Local validation on the latest `v1.0-dev` head included:

```text
npm run lint:prettier
npm run lint:style-contract
bundle exec al-folio upgrade audit --no-fail
bundle exec jekyll build --baseurl /al-folio
```

The Jekyll build was run with Jupyter/nbconvert available and with network access for external post feeds; it passes after clearing a stale generated `_site` placeholder file from the local macOS filesystem.

The latest push also fixes the broken-link failures caused by the docs move; GitHub CI has been refreshed on the new head.

_Written by Codex on behalf of @alshedivat._

dfuchss:
Thanks for the detailed response and for using my site as the rehearsal target; that's really appreciated and very helpful for understanding the migration path!

I just wanted to follow up on one aspect of my original concern that I think wasn't fully addressed: tracking upstream changes in files I've locally overridden.

With the pre-v1 workflow, if I had a local copy of e.g. _includes/repository.liquid and upstream changed it, git merge would surface a conflict; giving me an explicit signal to review and reconcile the change. With v1.0, if I keep a local override and the owning gem updates that same file, a bump to Gemfile.lock is the only signal I get; with no diff, no conflict, and no indication that my override may now be stale.

This is also what I had in mind with my earlier suggestion of a "de-modularized / full" read-only aggregated repository that vendors all plugin content via subtrees; it would restore the git-based diff/conflict workflow for people with customized forks.

I completely understand that this is a hard problem given the new architecture, and I'm not expecting a perfect solution right away. But I was wondering: is there a planned approach for this tracking gap, or is the current idea that users manually diff their overrides against gem changelogs on each update?

alshedivat:
Thanks @dfuchss, this concern is valid. The intended answer is no longer “manually diff gem changelogs.” I added an explicit override reconciliation workflow in `al_folio_upgrade` 1.0.3 and pinned it in this PR.

The new workflow is:

```text
bundle exec al-folio upgrade overrides audit
bundle exec al-folio upgrade overrides diff PATH
bundle exec al-folio upgrade overrides accept PATH
```

What it does:

- detects local files that shadow installed `al-*` / `al_folio_*` plugin files;
- reports the owning gem, installed gem version, upstream path, and local path;
- stores reviewed baselines in `.al-folio-overrides.yml` with upstream/local checksums;
- after a future gem update, marks an override stale when the upstream plugin file changed since the last acknowledgement.

I also updated the migration docs, agent instructions, and the fuchss.org rehearsal PR to use this. Running the new audit on the fuchss.org migration branch reports 78 local overrides, including the local jQuery-era JS files behind the rendering issues you noticed. That is exactly the missing signal you described: the branch can build, but the override audit now tells users which local copies still need reconciliation.

I still think a generated read-only aggregate repo could be useful later for people who strongly prefer a Git-native full-tree diff, but I would treat that as a post-v1 convenience. The release-blocking fix is the tool-based override drift signal, because it works directly in customized starter forks and with Bundler-pinned plugin versions.

_Written by Codex on behalf of @alshedivat._

Issue #2787, opened 2024-10-19
RenderCV integration


m3lixir:
### Have you checked that your feature request isn't already filed?

- [X] I read through [FAQ](https://github.com/alshedivat/al-folio/blob/master/FAQ.md) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my feature request.
- [X] Yes, I have checked that this feature request isn't already filed.

### Description & Motivation

CV page content is generated from either a json or yml. json input conforms to JSON Resume schema (https://github.com/alshedivat/al-folio/issues/1513), however there is no schema defined for yml input. I propose we add in RenderCV support and add functionality so that the pdf of the CV can be generated on your current cv.yml (on each push).

- https://github.com/sinaatalay/rendercv
  - https://github.com/sinaatalay/rendercv/blob/main/schema.json
- https://github.com/sinaatalay/rendercv-pipeline

gmsundar15:
If RenderCV support is added, it would be great if the JSON resume could be used to automatically generate the downloadable PDF of the CV using RenderCV with each push, just like the yml file. 

To achieve this, JSON resume can be easily [converted into the CV section of RenderCV](https://github.com/jsonresume/jsonresume.org/tree/master/packages/converters/jsonresume-to-rendercv) with the RenderCV design, locale, and settings being specified using the `--design`, `--locale-catalog` and `--rendercv-settings` runtime options.

george-gca:
If you know how to do this, we are welcoming PRs.

gmsundar15:
I did try creating a worker to achieve this but there is an open bug report on RenderCV when using external YAML files which causes my approach to fail. I will give it another go once this bug has been fixed.

https://github.com/rendercv/rendercv/issues/272

george-gca:
Does it work when using a path without spaces? Because I don't think this is much of an issue, since it is recommended to avoid spaces in directories name. If it works in a path without spaces (like `my design/design.yaml`), then it is fine. We just add a line in README to warn about avoiding spaces.

gmsundar15:
I tried running it locally and get the same error when run without any spaces in the absolute file path. It seems like these runtime options are just completely broken.

george-gca:
I sent a PR there (https://github.com/rendercv/rendercv/pull/310). After it gets accepted you can move on with the code.

gmsundar15:
[...]

PR #2969, opened 2025-01-19
Migrate cv.yml to RenderCV schema

gmsundar15:
The cv.yml will follow the RenderCV schema as the first step towards auto-generating CV PDFs from cv.yml or resume.json as discussed in #2787. The YAML file also supports markdown formatting to make it RenderCV friendly. 

example_pdf.pdf updated to pdf generated by RenderCV using the example cv.yml file.

The existing cv.yml format would be deprecated after this PR.


gmsundar15:
The deployed preview defaults to resume.json, so you can actually see the preview there. But there is an issue with `list-group-item:first-child` having a top border when it shouldn't have that. I am unable to debug this.
2025-01-19T22:20:41,110015214+00:00

george-gca:
I think I can took a look at it tomorrow.

_layouts/cv.liquid
```diff
@@ -28,26 +28,88 @@ layout: default
...
-          </div>
+        {% assign cv = site.data.cv.cv %}
+        {% include cv/gen-info.liquid %}
+        {% for entry in cv.sections %}
+          {% unless entry[0] == 'summary' %}
+            {% if entry[0] == 'education' %}
+              <a class="anchor" id="{{ entry[0] }}"></a>
+              <div class="card mt-3 p-3">
+                <h3 class="card-title font-weight-medium">{{ entry[0] | capitalize }}</h3>
+                {% include cv/education.liquid %}
```
george-gca, line 39: Not really sure why, but adding a `<div>` surrounding this line fixes the extra horizontal line in your layout. I just compared your rendered HTML with the original one and found that it had a `<div>` less. The same for the other sections with that extra line.

george-gca:
Just tried to create an action to validate the yml against the schema, but it is failing. But maybe is because [this schema](https://docs.rendercv.com/user_guide/structure_of_the_yaml_input_file/) is valid only to be used inside code editors?

gmsundar15:
The JSON Resume to RenderCV converter I was looking at was successfully using the same URL to validate the schema. https://github.com/guruor/jsonresume-to-rendercv/blob/main/jsonresume_to_rendercv/converter.py

It seems that a new version of RenderCV has been released a couple of days ago with some changes and this cv.yml does not follow the new schema. This could be causing the errors. 
I will take a look at this soon.

_layouts/cv.liquid
george-gca: Also take a look at this.

gmsundar15: I have fixed this. I'm also not sure why this was causing an issue.

gmsundar15:
Sorry for the delay.

I have fixed the rendering issues and updated the schema. However, the issues with validating the schema using actions were not caused by the file not following the schema, as the old file was still following the schema as per my editor.  The update to the schema only added new entries, which I wasn't using in the old file.

george-gca:
Have you figured it out why is it failing the schema action?

gmsundar15:
My guess is that using a dedicated YAML validator instead of using a JSON validator should work. Also, the validator should not strictly object to extra tags which are not defined in the schema but used for rendering the web CV.
I will give this a try and see if it works.

jiuntian:
Thank you for your efforts on this! This improvement is valuable to the community. Regarding schema validation, I tested it and found that only a few small modifications are needed to make it work correctly.

1. Add the missing `required` variable( /properties/design/required ), as this seems to incompatible with JSON schema format (of validator).
```diff
5390c5390,5393
-       "title": "Design"
---
+       "title": "Design",
+       "required": [
+         "theme"
+       ]

```
2. Enable `fixSchemas` to handle certain unsupported formats (e.g., path, which is of type string).
```yaml
    - name: Validate cv.yml 🧪
      uses: cardinalby/schema-validator-action@v3
      with:
        file: "_data/cv.yml"
        mode: "lax"
        schema: "rendercv-schema.json"
        fixSchemas: true
```
3. Fix the cv.yml to ensure passing validation and successful CV generation.
The version included in this PR failed to generate a valid CV using rendercv. However, when I tested with the example resume YAML from the RenderCV repository, it passed validation and successfully generated a CV. 


These are my patchy workarounds to make the CV YAML validation work. It might not be perfect, so feel free to refine or suggest improvements. I hope this contributes to the development of RenderCV support, as it is highly anticipated and could bring significant benefits to this project.

george-gca:
Thanks @jiuntian, I'll take a look at this later.

jiuntian:
Thanks @george-gca 

While I trying to made this to work on my own website, I also notice another incorrect variable name that made the sorting in CV failed.

Can check the following changes:
https://github.com/jiuntian/al-folio/commit/71c43689c86991d7d614bdbcfc78f169ddbe87dc

jiuntian:
RenderCV renders CV sections dynamically based on the entry type, as described in the [RenderCV documentation](https://docs.rendercv.com/user_guide/structure_of_the_yaml_input_file/#educationentry).I believe we can adopt a similar approach, instead of hardcoding specific sections. This would eliminate the need to modify the layout whenever the user wish to add a new section, especially since we’re planning to use the RenderCV schema for cv.yml.

I’ve implemented a dynamic layout for this project that renders CV sections based on the entry type defined by the RenderCV schema. Each entry type is detected by its mandatory fields and rendered using its own Liquid layout file.

You can view the commit on my personal website here: https://github.com/jiuntian/al-folio/commit/a89099613243124f122e19df415722e107f825eb. This implemented all currently available entry types: [EducationEntry](https://docs.rendercv.com/user_guide/structure_of_the_yaml_input_file/#educationentry), [ExperienceEntry](https://docs.rendercv.com/user_guide/structure_of_the_yaml_input_file/#experienceentry), [PublicationEntry](https://docs.rendercv.com/user_guide/structure_of_the_yaml_input_file/#publicationentry), [NormalEntry](https://docs.rendercv.com/user_guide/structure_of_the_yaml_input_file/#normalentry), [OneLineEntry](https://docs.rendercv.com/user_guide/structure_of_the_yaml_input_file/#onelineentry), [BulletEntry](https://docs.rendercv.com/user_guide/structure_of_the_yaml_input_file/#bulletentry), and [TextEntry](https://docs.rendercv.com/user_guide/structure_of_the_yaml_input_file/#textentry). Let me know if you think this proposal is useful, we can discuss how best to integrate it.

george-gca:
@jiuntian I think that your solution would be really useful. Are you still willing to submit a PR?

jiuntian:
@george-gca I’d be happy to submit a PR once I have the capacity. I’ll keep you posted!

neuroconvergent:
I see the point in using the native section system of RenderCV rather than imitating the JSONresume like i did. I will now push a workflow for auto generation of RenderCV PDF which can be integrated into your solution.  

As for the schema validation, the addition of #properties/design/required is one of the problems but there are many problems associated with it, such as a typo in the main code for section_title type, there is a type named "with-parial-line" instead of the type "with-partial-line" which is specified in the schema. Currently, using a local schema file with version lock of RenderCV in the workflows seems to be for the best. I will raise an issue on RenderCV repo to improve the schema in the meantime.

george-gca:
Closing it in favor of #3462.

PR #2933, opened 2025-01-01
Howto remove pages

george-gca:
Closes #2930

PR #2978, opened 2025-01-23
[Unbundle] Changed to use plugin `jekyll-3rd-party-libraries` plugin

george-gca:
I agree that it is a good idea to unbundle some of the plugins from `al-folio`, specially since they can be useful to more jekyll users overall. I have unbundled 2 plugins and created a gem for each one of them, so they can be easily integrated in any jekyll site. This is the first one of them.


alshedivat:
@george-gca, thanks for contributing to unbundling, I'll take a closer look at these third party libraries.

I'll write up a more detailed plan in #2968 for how to best go about unbundling some of the functionality. for some of the common jekyll functionality (like cache busting or various 3rd party libraries) it makes sense to create general jekyll plugins (prefixing names of the ruby gems with `jekyll-`). for al-folio specific functionality (even things that depend on 3rd libraries), I decided to prefix plugin names with `al-` just to indicate that it's part of `al-folio` ecosystem. also, I created https://github.com/al-org-dev for all these unbundled al-folio-related repos.

george-gca:
These plugins I created by moving code that can be more broadly used to their own plugins. Also created gems for them for easy installation.

Do you plan to move also post related features to plugins? I saw that you created #2984, but idk how hard it is to create a completely separated gem that can be also used for more sites from them. I mean, by ust installing a gem without you having to manually copy some js files to the repo that will use it.

Also I kept thinking about how could we avoid the merge that happens everytime a new functionality is added to the template. The best solution so far is to add these files to `excludes` in `_config.yml`, like recommended in https://github.com/alshedivat/al-folio/pull/2933#issuecomment-2571271117.

alshedivat:
makes sense, thanks. and good question about moving post-related features to plugins, I'll look into this.

PR #2979, opened 2025-01-23
Changed to use plugin `jekyll-cache-bust` plugin

george-gca:


alshedivat (approved):
LGTM, thanks for turning this into a standalone plugin.

Issue #1828, opened 2023-10-13
On Adding Hashnode RSS: Liquid Exception: Numerical argument is out of domain - sqrt in /SwayamInSync.github.io/_layouts/post.html


SwayamInSync:
Made these changes to re-create the error
```yml
# External sources.
# If you have blog posts published on medium.com or other external sources,
# you can display them in your blog by adding a link to the RSS feed.
external_sources:
  - name: Hashnode.com
    rss_url: https://swayam-blog.hashnode.dev/rss.xml
```

pskadasi:
when this template is used for the first time, we are getting similar error directing to `_layouts/post.html`, due to this webpage is not being deployted. can some one please tackle this bug.

rahulkumar-aws:
Hi,
Hope it's the right place to report, I am also facing a similar issue.

`external_sources:
  - name: medium.com
    rss_url: https://medium.com/@rahulkumar.aws/feed
`

error:
`
Liquid Exception: Numerical argument is out of domain - "sqrt" in /Users/rahulkumar/Documents/GitHub/rahulkumar-aws.io/_layouts/post.html
`
`
/Users/rahulkumar/.rbenv/versions/3.0.0/lib/ruby/gems/3.0.0/gems/classifier-reborn-2.3.0/lib/classifier-reborn/extensions/vector.rb:64:in `sqrt': Numerical argument is out of domain - "sqrt" (Math::DomainError)
`

This error the stops server from running.

rudresh-v:
Same Issue!

YifanJiang233:
I find this issue is caused by the `related_blog_post` with the usage of `--lsi` (latent semantic indexing)  option. If there is no need for related posts, one can simply fix the issue by setting `enable` as false in `_config.yml`.

Otherwise, it is rather tricky to fix this issue. After tracing back the out of domain error, I find it is due to the accumulated numerical error of SVD used in [classfier-reborn](https://jekyll.github.io/classifier-reborn/). In short, they self-crafted a pure Ruby SVD algorithm and it is not very robust. The orthonormal matrices they introduced might not be exact due to floating-point error and this could lead to a negative value inside the square root of line 64  [here](https://github.com/jekyll/classifier-reborn/blob/d8872fe18ff80fa6f6a4b66384a3a9a58090a493/lib/classifier-reborn/extensions/vector.rb). One can fix the issue locally by changing line 64 to 
`s << Math.sqrt(qrot[r, r].abs)`
[...]

PR #2983, opened 2025-01-25
fixed indexing of external posts

niebles:
This should fix several issues with indexing external posts, including #1828.

In short, I found that the issue with indexing was that the index builder was receiving 'empty' documents. To fix that, I'm setting the document content to be the post content as retrieved from the rss feed or the text extracted from the external page.

I've tested with various blog sources and it seems to be working as expected now.


alshedivat (approved):
LGTM, thanks for this fix!

PR #3002, opened 2025-02-05
Allows for standard Jekyll code highlighting with distill template

jacobaustin123:
As described in https://github.com/alshedivat/al-folio/issues/300, standard Markdown code highlighting doesn't work in the Distill template because it is clobbered by the Distill "Prism" syntax highlighting. This PR will allow d-code tags (distill-specific code tags) to function with the Distill syntax highlighter, but will default `highlight` and Markdown code blocks to use the Jekyll/Rouge highlighter.

This PR also removes overrides.js, which is currently broken because jquery isn't loaded.


jacobaustin123:
Before: 
image

After:

image

george-gca:
It is missing some prettier fixes.

jacobaustin123:
I pushed a set of prettier fixes. This should now pass presubmits.

jacobaustin123:
Note that fwiw prettier is causing a bunch of bad formatting in the md file: dedenting the code blocks and unnecessarily escaping the *.

PR #3015, opened 2025-02-13
fix language-not-found issues on repo page with simplified and traditional Chinese

jiaye-wu:
The author of github-readme-stats uses the non-standard code "cn" for simplified Chinese, see [here](https://github.com/anuraghazra/github-readme-stats?tab=readme-ov-file#available-locales). While for traditional Chinese the author only provides "zh-tw". The github-readme-stats functions on repo page will break if the user sets the site language to ALL variants of Chinese except for zh-tw. This hack is to make all sub-variants of simplified Chinese fall back to "cn" and all sub-variants of traditional Chinese fall back to "zh-tw".

This patch fixes the problem and has been tested locally & with GitHub pages.

### DIFF
--- a/_includes/repository/repo.liquid
+++ b/_includes/repository/repo.liquid
@@ -18,11 +18,11 @@
         {% assign lang = 'cn' %}
       {% when 'tw', 'hk', 'mo', 'hant' %}
         {% assign lang = 'zh-tw' %}
-
     {% endcase %}
-{% endcase %}
 
 {% comment %} Add a new language using when... if needed {% endcomment %}
+{% endcase %}
+
 {% comment %} If you still encounter language-display issues, check the available locale codes in github-readme-stats (different from ISO-639 standard used in your website): https://github.com/anuraghazra/github-readme-stats?tab=readme-ov-file#available-locales {% endcomment %}
 
 {% if site.data.repositories.repo_description_lines_max %}
--- a/_includes/repository/repo.liquid
+++ b/_includes/repository/repo.liquid
@@ -5,6 +5,20 @@
   {% assign show_owner = true %}
 {% endif %}
 
+{% assign lang = site.active_lang | split: '-' | first %}
+{% if lang == 'pt' %}
+  {% assign lang = site.active_lang %}
+{% endif %}
+{% if lang == 'zh' %}
+  {% assign lang_last = site.active_lang | split: '-' | last %}
+  {% if lang_last == 'cn' or lang_last == 'sg' or lang_last == 'my' or lang_last == 'hans' %}
+    {% assign lang = 'cn' %}
+  {% elsif lang_last == 'tw' or lang_last == 'hk' or lang_last == 'mo' or lang_last == 'hant' %}
+    {% assign lang = 'zh-tw' %}
+  {% endif %}
+{% endif %}
+{% comment %} If you still encounter language-display issues, check the available locale codes in github-readme-stats (different from ISO-639 standard used in your website): https://github.com/anuraghazra/github-readme-stats?tab=readme-ov-file#available-locales {% endcomment %}
+
 {% if site.data.repositories.repo_description_lines_max %}
   {% assign max_lines = site.data.repositories.repo_description_lines_max %}
 {% else %}
--- a/_includes/repository/repo.liquid
+++ b/_includes/repository/repo.liquid
@@ -5,24 +5,24 @@
   {% assign show_owner = true %}
 {% endif %}
 
-{% assign lang = site.active_lang | split: '-' | first %}
+{% assign lang = site.lang | split: '-' | first %}
 
 {% case lang %}
   {% when 'pt' %}
     {% assign lang = site.active_lang %}
 
   {% when 'zh' %}
-    {% assign lang_last = site.active_lang | split: '-' | last %}
+    {% assign lang_last = site.lang | split: '-' | last %}
     {% case lang_last %}
       {% when 'cn', 'sg', 'my', 'hans' %}
         {% assign lang = 'cn' %}
       {% when 'tw', 'hk', 'mo', 'hant' %}
         {% assign lang = 'zh-tw' %}
-
-        {% comment %} when... add a new language here if needed {% endcomment %}
+        
     {% endcase %}
 {% endcase %}
 
+{% comment %} Add a new language using when... if needed {% endcomment %}
 {% comment %} If you still encounter language-display issues, check the available locale codes in github-readme-stats (different from ISO-639 standard used in your website): https://github.com/anuraghazra/github-readme-stats?tab=readme-ov-file#available-locales {% endcomment %}
 
 {% if site.data.repositories.repo_description_lines_max %}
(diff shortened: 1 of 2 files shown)


_includes/repository/repo.liquid
```diff
@@ -5,6 +5,20 @@
   {% assign show_owner = true %}
 {% endif %}
 
+{% assign lang = site.active_lang | split: '-' | first %}
+{% if lang == 'pt' %}
+  {% assign lang = site.active_lang %}
+{% endif %}
+{% if lang == 'zh' %}
```
george-gca, line 12: Shouldn't here also be an `elsif`?

_includes/repository/repo_user.liquid
```diff
@@ -1,3 +1,17 @@
+{% assign lang = site.active_lang | split: '-' | first %}
+{% if lang == 'pt' %}
+  {% assign lang = site.active_lang %}
+{% endif %}
+{% if lang == 'zh' %}
```
george-gca, line 5: Shouldn't here also be an `elsif`?

george-gca (changes requested):

_includes/repository/repo.liquid
jiaye-wu: I prefer to keep each language (pt, zh, fr, ...) on its own case of `if`, then when more languages are added, the code won't be stacked with too many levels of `elsif`. I am no expert on liquid, do you expect the code to work differently if we leave it this way?

_includes/repository/repo_user.liquid
jiaye-wu: I prefer to keep each language (pt, zh, fr, ...) on its own case of `if`, then when more languages are added, the code won't be stacked with too many levels of `elsif`. I am no expert on liquid, do you expect the code to work differently if we leave it this way?

_includes/repository/repo.liquid
jiaye-wu: To make the code more readable and eaiser to insert other language, I opted to use the `case-when` structure. Tested on my own website.

_includes/repository/repo_user.liquid
jiaye-wu: To make the code more readable and eaiser to insert other language, I opted to use the `case-when` structure. Tested on my own website.

_includes/repository/repo.liquid
```diff
@@ -5,6 +5,26 @@
...
+
+  {% when 'zh' %}
+    {% assign lang_last = site.active_lang | split: '-' | last %}
+    {% case lang_last %}
+      {% when 'cn', 'sg', 'my', 'hans' %}
+        {% assign lang = 'cn' %}
+      {% when 'tw', 'hk', 'mo', 'hant' %}
+        {% assign lang = 'zh-tw' %}
+
+        {% comment %} when... add a new language here if needed {% endcomment %}
```
george-gca, line 22: This comment must be after the `endcase`, if not the language would be added inside the `zh` case.

_includes/repository/repo_user.liquid
```diff
@@ -1,3 +1,23 @@
...
+
+  {% when 'zh' %}
+    {% assign lang_last = site.active_lang | split: '-' | last %}
+    {% case lang_last %}
+      {% when 'cn', 'sg', 'my', 'hans' %}
+        {% assign lang = 'cn' %}
+      {% when 'tw', 'hk', 'mo', 'hant' %}
+        {% assign lang = 'zh-tw' %}
+
+        {% comment %} when... add a new language here if needed {% endcomment %}
```
george-gca, line 15: This comment must be after the `endcase`, if not the language would be added inside the `zh` case.

george-gca (changes requested):

george-gca:
Also, you are missing the use of the `lang` variable in both scripts, like `&locale={{ lang }}`.

_includes/repository/repo_user.liquid
george-gca: The case of using `elsif` instead of a sequence of `ifs` was to avoid checking the second if if it already entered the first condition. But it looks fine with `case-when`.

jiaye-wu:
> You should use `site.lang` instead of `site.active_lang` here. `site.active_lang` is a [polyglot feature](https://github.com/untra/polyglot?tab=readme-ov-file#features). In the `al-folio` we use the `lang` defined in `_config.yml`.
> 
> https://github.com/alshedivat/al-folio/blob/10d007a98c63043b1e849994eca80098a2c2611b/_config.yml#L18
> 
> Also, you are missing the use of the `lang` variable in both scripts, like `&locale={{ lang }}`.

Sorry, I don't understand your last sentence. Could you please do the modification after I fix `site.active_lang`?

_includes/repository/repo_user.liquid
george-gca: Can you move the comments to inside the outer `case`? Since there is where the `when` must be added, so it is clear to the user.

_includes/repository/repo.liquid
```diff
@@ -5,6 +5,26 @@
...
+      {% when 'cn', 'sg', 'my', 'hans' %}
+        {% assign lang = 'cn' %}
+      {% when 'tw', 'hk', 'mo', 'hant' %}
+        {% assign lang = 'zh-tw' %}
+
+    {% endcase %}
+{% endcase %}
+
+{% comment %} Add a new language using when... if needed {% endcomment %}
+{% comment %} If you still encounter language-display issues, check the available locale codes in github-readme-stats (different from ISO-639 standard used in your website): https://github.com/anuraghazra/github-readme-stats?tab=readme-ov-file#available-locales {% endcomment %}
```
george-gca, line 26: Can you move the comments to inside the outer `case`? Since there is where the `when` must be added, so it is clear to the user.

_includes/repository/repo_user.liquid
```diff
@@ -1,14 +1,34 @@
...
+      {% when 'cn', 'sg', 'my', 'hans' %}
+        {% assign lang = 'cn' %}
+      {% when 'tw', 'hk', 'mo', 'hant' %}
+        {% assign lang = 'zh-tw' %}
+
+    {% endcase %}
+{% endcase %}
+
+{% comment %} Add a new language using when... if needed {% endcomment %}
+{% comment %} If you still encounter language-display issues, check the available locale codes in github-readme-stats (different from ISO-639 standard used in your website): https://github.com/anuraghazra/github-readme-stats?tab=readme-ov-file#available-locales {% endcomment %}
```
george-gca, line 19: Can you move the comments to inside the outer `case`? Since there is where the `when` must be added, so it is clear to the user.

george-gca (changes requested):

_includes/repository/repo.liquid
jiaye-wu: Done.

_includes/repository/repo_user.liquid
jiaye-wu: Done.

george-gca (approved):

george-gca (approved):

george-gca:
Thanks for your contribution!

PR #3054, opened 2025-03-11
[BugFix] exclude `CUSTOMIZE.md` in broken link check workflow

fedebotu:
As per the title, we should exclude `CUSTOMIZE.md` in the broken link check workflow.

If something was deleted from `_pages`, `_posts`, `_projects` etc (in my case, I deleted the `dropdown.md`), then the `link-checker` workflow would fail since it checks these template links from the `CUSTOMIZE.md` file. Since the latter only contains instructions, we can exclude it from the link checker :)


george-gca:
I am not sure if this is the best solution. I understand what you are saying, but this check in `CUSTOMIZE.md` also helps us make sure that our documentation points to the right places. I only removed `README.md` because it points to a lot of outside links that might not exist anymore. I think the best solution might be for you to also exclude `CUSTOMIZE.md`.

fedebotu:
> but this check in CUSTOMIZE.md also helps us make sure that our documentation points to the right places

I see what you mean. However, since `CUSTOMIZE.md` is part of the documentation, it has to be checked only by the upstream repo `alshedivat/al-folio/`, while users don't need such checks, right?

> I think the best solution might be for you to also exclude CUSTOMIZE.md.

So you are saying that my proposed change (excluding the file) is fine? Perhaps I didn't fully get your point :eyes:

george-gca:
> I see what you mean. However, since `CUSTOMIZE.md` is part of the documentation, it has to be checked only by the upstream repo `alshedivat/al-folio/`, while users don't need such checks, right?

Right, but I am not sure if we can do something to enable this only for the main repo.

> So you are saying that my proposed change (excluding the file) is fine? Perhaps I didn't fully get your point 👀

I think that in your repo you should exclude the `CUSTOMIZE.md` file, and we should skip this PR. Unless there is some solution to check for these files only on our repo.

Maybe we should create a different check broken links that would run only on the main repo? Since this check is useful for the user when it is referencing files locally. What do you think?

thoklei:
This may be overkill, but wouldn't it be better to check the final site itself instead of just the files? In my fork, I've changed the workflow to build the whole thing, serve it locally, and run lychee against the served website. This takes a minute, but has the advantage that I don't need to exclude anything, and all internal links are checked properly as well. This seems like a cleaner solution. Happy to share how I did it, if there is interest.

george-gca:
We already have this check in our repo in [.github/workflows/broken-links-site.yml](https://github.com/alshedivat/al-folio/blob/main/.github/workflows/broken-links-site.yml). What I think it is best is to only enable this check if this is the main repo. Instead of adding `CUSTOMIZE.md` to the ignore, add this as a condition:

https://github.com/alshedivat/al-folio/blob/f68e88c5ef64c0e3b31626da71c333a7fddf1468/.github/workflows/deploy-image.yml#L20

george-gca:
Closing this in favor of #3118.

PR #3070, opened 2025-03-13
Fix typo in CUSTOMIZE.md

jiuntian:
`excludes` should be `exclude`. `excludes` with `s` does not work with `_config.yml`, which itself uses `exclude` and `include`.


george-gca:
Thanks for fixing this.

PR #3074, opened 2025-03-14
Update INSTALL.md

pourmand1376:
Again, I saw this version problem with current version of our docker image. Supposedly, there is some json package problem which I dealt with. 

But here is the overall solution to this problem for whoever is having it.


george-gca:
Actually this json might be my fault. Dependabot updated json version in #3066, but I think I forgot to update the docker image tag. Releasing a new version will fix this json problem. Will do it now.

pourmand1376:
Thanks 🙏

PR #3089, opened 2025-03-22
Re-add ability to exclude posts from search

pranjalagg:
Earlier there was a tag in the `_config.yml` file to exclude the posts from the search results. I could no longer find that option and hence modified the `search.liquid.js` file.
It basically checks if the `posts_in_search` tag is true or not.


george-gca:
I don't remember if we had this or not, but it is a nice addition anyway. Thanks.

george-gca (approved):

PR #3105, opened 2025-03-31
Simple

RyanCaldwell360:
The al-folio template was more complicated and challenging to get up and running. This merge is just for a very basic start using Jekyll minima.


RyanCaldwell360:
I apologize, this was meant to be for my forked repo. Definitely don't want to accept this.

PR #3107, opened 2025-04-02
Fix "comma and and" before last author in publication list

tom111:
When formatting the list of authors of a paper, there is a comma and the word 'and' before the last author. With 3 or more authors this can be interpreted as Oxford comma, but with two authors it is just wrong.

I applied this patch locally to get rid of the last comma.

Before:
before

After:
after

### DIFF
--- a/_layouts/bib.liquid
+++ b/_layouts/bib.liquid
@@ -79,8 +79,8 @@
         {%- endif -%}
 
         {%- if forloop.length > 1 -%}
-          {% if forloop.first == false %},&nbsp;{% endif %}
-          {%- if forloop.last and author_array_limit == author_array_size %}and&nbsp;{% endif -%}
+          {% if forloop.first == false and forloop.last == false%},&nbsp;{% endif %}
+          {%- if forloop.last and author_array_limit == author_array_size %}&nbsp;and&nbsp;{% endif -%}
         {% endif %}
         {%- if author_is_self -%}
           <em>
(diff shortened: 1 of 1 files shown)


tom111:
Too bad, this creates a problem with the folded authors beyond the limit:

grafik

I'll have to look into this. Of course we want a ", " between Planck and Schrödinger and maybe no comma before the "and" still.

tom111:
I've pushed a second patch. Now it just resolves the problem that

> "Author, and Author" 

is wrong. There should be no ", and" between two items ever.
The Oxford comma is preserved as it seems to be desired. Thus

> "Author, Author, and Author" 

will still print like this. 

I also took extra care with the whitespace, not introducing any double spaces. Hope I caught all. ✌️

One could of course also make the Oxford comma (the last comma that goes with the 'and') an option, but I've not worked on this.

tom111:
While I was at it, I fixed two other bugs in [1eb6e89](https://github.com/alshedivat/al-folio/pull/3107/commits/1eb6e898330df50d04ccc5406acdb286508a330e)

- The whitespaces after the comma in "Author, Author, ..." should be breakable. Does not make sense to have non-breaking whitespace there. This just leads to breaks between the first and the last name of the author which are inferior.

- There was a extra space bug in the case that a co-authors name appears in coauthors.yml which was fixed by more whitespace stripping.

alshedivat (approved):
LGTM, thanks for this fix!

Issue #3129, opened 2025-04-13
The platform-dependent shortcut key next to Search icon shifts top nav links over briefly on load


dylanhawley:
### Have you checked that your issue isn't already filed?

- [x] I read through [FAQ](https://github.com/alshedivat/al-folio/blob/main/FAQ.md) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [x] Yes, I have checked that this issue isn't already filed.

### Bug description

The platform-dependent shortcut key next to Search icon shifts top nav links over briefly on load

### How to reproduce the bug

1. Be on MacOS
2. Click any link on the top nav bar of an al-folio templated website
3. When the new page loads, observe for a brief moment that the top nav bar next shifts over. This is due to the fact that the page renders "Ctrl k" text by default, but then switches to " Cmd k" if the browser is on a Mac. This logic is slow enough that I can see the text change happen as the page renders. Not sure how to fix, but I know the cause

### Error messages and logs

```
N/A
```

### What operating system are you using?

Mac

### Where are you seeing the problem on?

Deployed site

george-gca:
This is weird, since we are not deferring the loading of the [assets/js/shortcut-key.js](https://github.com/alshedivat/al-folio/blob/main/assets/js/shortcut-key.js), as can be seen here:

https://github.com/alshedivat/al-folio/blob/15fad29f23a80f5f2ea24f421435e2dc76df5e33/_includes/scripts.liquid#L331

I don't know how could we fix it.

dylanhawley:
I pushed a fix

dylanhawley:
> Not really, I don't use Mac. Maybe [@pourmand1376](https://github.com/pourmand1376) could replicate it?

I pushed a fix

PR #3136, opened 2025-04-15
Load Javascripts in header, instead of body.

dylanhawley:
To avoid a flash of unstyled content issue, `shorcut-key.js` must be loaded in the header. We achieve this by loading all Javascript in `scripts.liquid` in the header.

This resolves issue #3129, which only appears on MacOS.

*Note:* This issue could also be resolved by loading only `<script src="{{ '/assets/js/shortcut-key.js' | relative_url | bust_file_cache }}"></script> ` in the header, instead of all of the javascript in `scripts.liquid`.

### DIFF
--- a/_layouts/default.liquid
+++ b/_layouts/default.liquid
@@ -13,6 +13,9 @@
       <meta http-equiv="refresh" content="3; url={{ redirect }}">
     {% endif %}
     {% include head.liquid %}
+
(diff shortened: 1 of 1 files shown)


george-gca:
@dylanhawley can you change so that only the shortcut key script is loaded in head instead of everything?

PR #3137, opened 2025-04-16
CVSectionRename

Bodysbobb:
**Addressing issue #2274** by making the section name changeable. 
Allows users to rename CV sections (from resume.json) via _config.yml. The key refers to the original section name, and the label specifies the new display name shown on the webpage.
RenameCVSection

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -73,6 +73,7 @@ GEM
     ffi (1.17.1-arm-linux-gnu)
     ffi (1.17.1-arm-linux-musl)
     ffi (1.17.1-arm64-darwin)
+    ffi (1.17.1-x64-mingw-ucrt)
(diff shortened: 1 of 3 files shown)


Bodysbobb:
I’ve developed an alternative approach that involves labeling the section name directly within the resume.json. I personally prefer this method, as it allows for further development—specifically, automatically generating a downloadable PDF file from the same resume.json while inheriting the section names seamlessly.

The current approach in this pull request has limitations when it comes to inheriting customized section names from the configuration.

The new approach would look like this:
"work_label": "Work Experience",
"work": [
  { ... },
  { ... }
],

"education_label": "Education",
"education": [
  ...
]
Screenshot 2025-04-18 044105

george-gca:
This is interesting. How is this aligned with RenderCV schema?

Issue #2167, opened 2024-02-07
[Feature Request] automatically generate pulications in the CV from .bib file


mingwei-liu:
### Have you checked that your feature request isn't already filed?

- [X] I read through [FAQ](https://github.com/alshedivat/al-folio/blob/master/FAQ.md) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my feature request.
- [X] Yes, I have checked that this feature request isn't already filed.

### Description & Motivation

Thanks for the great work! When creating the cv page, I have to add the publications manually one by one. Is it possible to use the previous pubilications.bib file to automatically generate it, similar to the the publications pages?

george-gca:
Yes it is. For this you should create another file like [_layouts/bib.liquid](https://github.com/alshedivat/al-folio/blob/master/_layouts/bib.liquid) that will be responsible for displaying the bib information in the correct format, and then use the bibliography tag in the cv page, overriding the default bibliography template by passing the `--template` or `-T` parameter. For more information see [jekyll-scholar](https://github.com/inukshuk/jekyll-scholar?tab=readme-ov-file#bibliography-template) documentation.

If you do this modification, would you mind sending a PR?

mingwei-liu:
Sure. I will give it a try.

gcostantino:
hello, I am trying to do that as well, but I don't manage to "print" the publications inside a "publicatiosn" tag inside the cv.yml. I had tried something like:

- title: Publications
  contents:
    bibliography_template: bib

but this does not work. It just prints "{"bibliography_template"=>"bib"}" inside my cv. Could you please help me? Thanks!

george-gca:
To make this work you must implement it like I said before.

HangJung97:
Any update on this? This feature would be useful.

george-gca:
We are accepting PRs on this matter. If you manage to implement this, it will be a good contribution.

PR #3141, opened 2025-04-17
Bib2Cv addressing #2167

Bodysbobb:
I noticed that someone (issue #2167) was looking for this feature, which I had implemented last month.

This update automatically links a `.bib` file (e.g., `papers.bib`), used for the Publications page, to the **Publications** section of the **CV** page—ensuring consistent formatting with the CV layout.  

Key features added to the CV’s Publications section via this plugin include:
- Hyperlinks to the same URLs as on the main Publications page  
- Underlined author names, following the same format  
- Separation of working papers, Master and Phd thesis, and published papers  

I hope this is helpful, and thank you so much for providing such a great template!  
I’ve also added other features like PDF/HTML conversion for the CV page, including a **Download PDF** button.


george-gca:
This is a very good addition to our template. Would you mind updating your PR?

Gemfile.lock
george-gca, line 1: Please avoid uploading Gemfile.lock unless it is essential for the PR.

assets/json/resume.json.bak
george-gca, line 1: I believe this file is not needed here.

george-gca (changes requested):
Wouldn't it be better to remove all publications from `resume.json` and keep then only on `papers.bib`, leaving only one to be used as an example? Also you should add a flag to avoid a publication to be added to the cv, maybe the user would want that?

Bodysbobb:
Thanks, I almost forgot about it. I will update it when I'm free.

Issue #872, opened 2022-09-07
Add calendar integration


chris-tkinter:
**Is your feature request related to a problem? Please describe.**
Hello,
Thank you for the awesome project!
I think it would be nice to add calendar support in al-folio. In fact calendars like google calendar support iframe so I think it is not hard to add, simply pasting the iframe code to any page should work. However I wonder how to seamlessly integrate into the current layout. 
Thanks!

PR #3144, opened 2025-04-19
Add calendar iframe and togglAdd calendar iframe and toggle buttone button

LucasLiu6:
### Description

This PR adds a Google Calendar integration feature as described in issue #872. Users can embed their own calendar using an iframe configured via `_config.yml`, and the calendar is toggleable through a responsive button. This component is styled to visually blend into the theme and supports responsive display across desktop and mobile.

### Changes
- Added `calendar.liquid` in `_includes`
- Modified `about.md` to include the toggleable calendar block
- Introduced `calendar` config section in `_config.yml`
- Appended calendar-related styles in `_sass/_base.scss`

### Configuration
To enable this feature, add the following to `_config.yml`:
```yaml
calendar:
  enabled: true
  calendar_id: your_calendar_id@group.calendar.google.com
  timezone: UTC
  style: "border:0; width:100%; height:600px;"

### DIFF
--- a/README.md
+++ b/README.md
@@ -485,3 +485,22 @@ Our most active contributors are welcome to join the maintainers team. If you ar
 The theme is available as open source under the terms of the [MIT License](https://github.com/alshedivat/al-folio/blob/main/LICENSE).
 
 Originally, **al-folio** was based on the [\*folio theme](https://github.com/bogoli/-folio) (published by [Lia Bogoev](https://liabogoev.com) and under the MIT license). Since then, it got a full re-write of the styles and many additional cool features.
+
+## calendar-control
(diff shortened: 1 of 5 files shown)


george-gca (changes requested):

george-gca:
I like your solution, but not like it being configured in `_config.yml` to appear in the about page. Maybe this should be more like an include, like [_includes/figure.liquid](https://github.com/alshedivat/al-folio/blob/main/_includes/figure.liquid), so the user could include it in any page it likes it most, even using different calendars in different pages. Think it like a professor creating a calendar for every lecture it will teach. What do you think?

Issue #3151, opened 2025-04-25
Minor issue: rss icon display control duplicated


amithjkamath:
### Have you checked that your issue isn't already filed?

- [x] I read through [FAQ](https://github.com/alshedivat/al-folio/blob/main/FAQ.md) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [x] Yes, I have checked that this issue isn't already filed.

### Bug description

in socials.yml, rss_icon: false does not hide the RSS icon, duplicated in _config.yml

### How to reproduce the bug

Searching for rss_icon in the repository points to two locations:

line 32, _data/socials.yml:

# rss_icon: false # set to true to show the RSS icon

and line 45, _config.yml:

rss_icon: true

The control in socials.yml does not really work. Setting it to false still shows the RSS icon, while only commenting the line out hides it. 

It appears in line 57,  _includes/social.liquid, the check is:

    {% when 'rss_icon' %}
      <a href="{{ site.baseurl }}/feed.xml" title="RSS Feed"><i class="fa-solid fa-square-rss"></i></a>

which does not check really for the true/false setting. 

I do not know if the right fix is either:

1. remove rss_icon from the socials list and have it be read from _config.yml, or
2. remove the mention from _config.yml and fix the true/false setting in _socials.yml

Any ideas on how this could be fixed? My quick fix to hide the icon is to comment out the line altogether. 

Thanks for putting together this fantastic template! 

### Error messages and logs

```
The error message you got, with the full traceback if available. Please paste it between these triple backticks.
```

### What operating system are you using?

Not applicable (e.g. you're using GitHub Pages or other hosting)

### Where are you seeing the problem on?

Running locally with Docker (devcontainer)

CalaW:
Should be closed by #3154

PR #3154, opened 2025-05-01
Remove unnecessary rss_icon line from _config.yml

CalaW:
Also make document clearer for rss_icon.


CalaW:
Oh I just noticed there's a existing issue #3151

PR #3158, opened 2025-05-08
add new hugging_face icon in Social (_includes/social.liquid)

rezwanh001:


_includes/social.liquid
```diff
@@ -18,6 +18,8 @@
       <a href="https://www.flickr.com/{{ social[1] }}" title="Flickr"><i class="fa-brands fa-flickr"></i></a>
     {% when 'github_username' %}
       <a href="https://github.com/{{ social[1] }}" title="GitHub"><i class="fa-brands fa-github"></i></a>
+    {% when 'huggingface_username' %}
```
george-gca, line 21: This should be added in alphabetical order.

george-gca (changes requested):

george-gca:
Also there's a problem: Hugging Face icon is not available on the [free tier](https://fontawesome.com/search?q=hugging%20face&o=r&ic=free).

rezwanh001:
However, this icon closely resembles the Hugging Face logo. We can include it as the Hugging Face icon for now, as it is an important social icon in this context.

george-gca:
This IS the Hugging Face logo, and it is not available on the free tier.

_includes/social.liquid
rezwanh001: maintained in alphabetical order for custom hugging face icon

george-gca:
It would be better if the Hugging Face logo could be fetched from other free font, since most al-folio users don't use paid fonts (me included).

Issue #3150, opened 2025-04-22
google scholar citation count no longer working.


kevinhong-research:
### Have you checked that your issue isn't already filed?

- [x] I read through [FAQ](https://github.com/alshedivat/al-folio/blob/main/FAQ.md) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [x] Yes, I have checked that this issue isn't already filed.

### Bug description

google scholar citation count no longer working.

### How to reproduce the bug

Go to any page with the google scholar widget, the citation cont shows NA

For example, just go to the main demo site: https://alshedivat.github.io/al-folio/publications/

### Error messages and logs

```
The error message you got, with the full traceback if available. Please paste it between these triple backticks.
```

### What operating system are you using?

Mac

### Where are you seeing the problem on?

Deployed site

juliosolisar:
+1 noticed same problem (on my website by also on the sample site from al-folio)

systemreliability:
The same

george-gca:
The Google Scholar citation count is not a reliable feature. I added since so many people were asking for it, but I am considering removing it altogether.

It works by scraping the Google Scholar page referent to the paper, and it only fetches new information when the page is built. The problem is that if Google somehow finds the access weird, it can simply block it, then the counter fails and there is nothing we can do. There isn't a Google Scholar api afaik, so our hands are tied.

This can be seen by the error message

```
Error fetching citation count for qyhmnyLat1gC: OpenURI::HTTPError - 403 Forbidden
```

ins the `Install and Build` step of the `Deploy site` action.

kastnerp:
I noticed the same. I wonder if [this approach](https://cse.bth.se/~fer/googlescholar-api/googlescholar.php?user=XNaNgkwAAAAJ&view_op=list_works&sortby=pubdate) could be used instead? Repo: https://github.com/fredrike/googlescholar-api

BernardoCama:
I fixed this issue on my personal site ([BernardoCama.github.io](https://bernardocama.github.io/)) by modifying the `bib.liquid` template to properly iterate through the citation data structure.

The key changes:

1. Updated how Google Scholar IDs are handled in `bib.liquid` to match the format in citations.yml
2. Added a loop to look for IDs with different prefix formats (with/without author ID)
3. Created a GitHub Action workflow that runs weekly (or can be triggered manually) to update citations

[...]

PR #3177, opened 2025-05-25
Adapt mechanism for citiation count

dfuchss:
I adapted the idea from @BernardoCama in https://github.com/alshedivat/al-folio/issues/3150

This PR consists of ... 
* an action that updates the publication count. (Be aware that this action will create a commit to the main branch if something changed regarding citiations. However, you need a PAT (to be added to the checkout action) if you want actions to be triggered by this commit (the default token go GH actions does not trigger any other action)
* _data/citations.yml contain the citation of all papers of an author
* bin/update_scholar_citations.py is the adapted update script for the citations.

## TODOs 
* proper documentation (I have to find a suitable place)
* Reviews, whether I messed up something. (I'm already using it at for my website; but still; I may have messed up something -> Draft PR)


george-gca:
What are the reasons for the Personal Access Token (PAT) to be needed? I don't think I completely understand it.

dfuchss:
Without a Personal Access Token (PAT), everything will function normally except that the commit updating citations.yml won't trigger any subsequent GitHub Actions --- meaning the update of the citation count won't be deployed. To ensure the CI pipeline runs after the commit, you need to use a PAT  (see https://github.com/orgs/community/discussions/25702)

bin/update_scholar_citations.py
```diff
@@ -0,0 +1,134 @@
...
+                    existing_data
+                    and "metadata" in existing_data
+                    and "last_updated" in existing_data["metadata"]
+                ):
+                    print(
+                        f"Last updated on: {existing_data['metadata']['last_updated']}"
+                    )
+                    if existing_data["metadata"]["last_updated"] == today:
+                        print("Citations data is already up-to-date. Skipping fetch.")
+                        return
```
george-gca, line 59: This whole code could be outside the `with` statement.

bin/update_scholar_citations.py
```diff
@@ -0,0 +1,134 @@
...
+    try:
+        with open(config_file, "r") as f:
+            config = yaml.safe_load(f)
+            scholar_user_id = config.get("scholar_userid")
+            if not scholar_user_id:
+                print(
+                    "No 'scholar_userid' found in the configuration file. Please add 'scholar_userid' to _data/socials.yml."
+                )
+                sys.exit(1)
+            return scholar_user_id
```
george-gca, line 27: This whole code could be outside the `with` statement.

bin/update_scholar_citations.py
```diff
@@ -0,0 +1,134 @@
...
+            )
+
+    # Compare new data with existing data
+    if existing_data and existing_data.get("papers") == citation_data["papers"]:
+        print("No changes in citation data. Skipping file update.")
+        return
+
+    try:
+        with open(OUTPUT_FILE, "w") as f:
+            yaml.dump(citation_data, f, width=1000, sort_keys=False)
```
george-gca, line 120: Shouldn't it be better to sort keys, to keep the file consistent between updates?

george-gca (changes requested):

george-gca:
Can you also add a section in CUSTOMIZE.md teaching how one could set up the personal key?

dfuchss:
> Can you also add a section in CUSTOMIZE.md teaching how one could set up the personal key?

Sure. I'll update the PR.

dfuchss:
@george-gca I've added the section. 

From my point of view, we have two options:

a) before merging, you should create a PAT secret for al-folio itself, otherwise the action will fail
b) by default we could comment the part in the action regarding the PAT; then users that want to use PAT, need to uncomment that part in the action.

What do you think?

requirements.txt
```diff
@@ -1 +1,3 @@
 nbconvert
+scholarly
```
george-gca, line 2: Let's keep this sorted.

george-gca (changes requested):

george-gca:
I prefer the second one, since I don't have permission to create the token myself (only @alshedivat does). I just checked, and our deploy workflow have a [manual dispatch](https://github.com/alshedivat/al-folio/blob/f05d31bf8a6b450c0cce52edaaf9c408be70af94/.github/workflows/deploy.yml#L62) option, so could you add it to the CUSTOMIZE.md the information that if you don't want to create a token, you can forcefully run the deploy action manually?

dfuchss:
@george-gca it's now adapted accordingly

dfuchss:
@george-gca do you need further changes :) ?

george-gca:
Sorry for the delay. I will try to find some time this week to take a proper look at it.

dfuchss:
Nice . As a reference , you can look into my page https://GitHub.com/dfuchss/fuchss.org . I use the code there already since before the PR :)

dfuchss:
@george-gca any updates regarding this PR. Do I need to modify something?

dfuchss:
Hi @george-gca,

Just a gentle ping on this PR! All the requested changes from July have been implemented, and it's been running smoothly on my site for months now.

Please let me know if you need anything else from me.

Thanks! 😊

PR #3183, opened 2025-05-30
updated README.md by adding  bullab page to labss

MartinBulla:


george-gca:
The lab should be added to the end of the list, not at the beginning.

MartinBulla:
a) I have not found any info about that. If documentation on this topic is missing,  should not we add that?  
b) Shall I thus open a new PR?

george-gca:
a) It is more like a good practice instead. Keeping the timeline of users and labs created is nice. But I will add that to the docs.
b) No, just modify your file and update the PR.

PR #3196, opened 2025-06-05
Fix Prettier Feature DevContainer Link

### LINKED ISSUE #3193: Prettier package in devcontainer.json no longer resolves (causing build fail)

### Bug description

Building a local DevContainer in VSCode fails due to "ghcr.io/devcontainers-contrib/features/prettier:1" not resolving—please see full error message below. Updating the line to "ghcr.io/devcontainers-extra/features/prettier:1" as per https://containers.dev/features appears to fix the issue. This bug may apply to other installation methods as well.

### How to reproduce the bug

Build (or rebuild) the DevContainer locally in VSCode. In under a minute, it may fail with the error attached below or similar.

### Error messages and logs

```
[9302 ms] Resolving Feature dependencies for 'ghcr.io/devcontainers-contrib/features/prettier:1'...
[9302 ms] * Processing feature: ghcr.io/devcontainers-contrib/features/prettier:1
[9380 ms] Could not resolve Feature manifest for 'ghcr.io/devcontainers-contrib/features/prettier:1'.  If necessary, provide registry credentials with 'docker login <registry>'.
[9380 ms] Github feature.
[9380 ms] Could not resolve Feature 'ghcr.io/devcontainers-contrib/features/prettier:1'.  Ensure the Feature is published and accessible from your current environment.
[9190 ms] Error: ERR: Feature 'ghcr.io/devcontainers-contrib/features/prettier:1' could not be processed.  You may not have permission to access this Feature, or may not be logged in.  If the issue persists, report this to the Feature author.
```

Windows

Running locally with Docker (devcontainer)

### More info

According to [this commit](https://github.com/devcontainers-extra/features/commit/c4313dddf394c3640b85b003a5ea166ed8e32ab7) in the devcontainers-extra repository, usage of "contrib" was being phased out in favor of "extra".

It is possible that this bug is only with my instance of al-folio. However, I am unable to resolve that URL even in my web browser. Any input about whether this is a real bug or not would be greatly appreciated. ~Jason

rinehartj:
Fixes #3193

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -9,7 +9,7 @@
     "ghcr.io/rocker-org/devcontainer-features/apt-packages:1": {
       "packages": "build-essential,imagemagick,inotify-tools,jupyter-nbconvert,procps,ruby-full,zlib1g-dev"
     },
-    "ghcr.io/devcontainers-contrib/features/prettier:1": {}
+    "ghcr.io/devcontainers-extra/features/prettier:1.0.2": {}
   },
 
   // Optionally: run jekyll serve automatically on container entering using the Docker entrypoint


CheariX:
LGTM. Tested + works as expected.

CheariX:
I tested with a clean checkout.
For some reason, I needed to run `yarn add --dev prettier @shopify/prettier-plugin-liquid` now.

george-gca:
This is weird, since [this file](https://github.com/alshedivat/al-folio/blob/f1406c47f7e841638f70248df71079d308b55b27/.prettierrc) should resolve this `@shopify/prettier-plugin-liquid` issue. Maybe the plugin should also be added somewhere to the devcontainer file?

george-gca (approved):

PR #3230, opened 2025-07-07
Update .prettierignore to include assets/js/distillpub/transforms.v2.js

ProfessorShroom:
I've had pretty checks fail due to this file, despite not changing it at all. Running a check on it yourself shows no issues so it's best to just ignore that file.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -10,3 +10,4 @@ _posts/2015-10-20-math.md
 _sass/font-awesome/*.scss
 _sass/tabler-icons/*.scss
 _scripts/*
+assets/js/distillpub/transforms.v2.js


ProfessorShroom:
Added Buy me a coffee (and it's 3 variations, beer, pizza, and book. And also Thingiverse, to socials

george-gca:
You should create a different PR for the socials part. Also, we can't ignore this file, since it is needed by the distill template.

ProfessorShroom:
I've changed it to just the social edits so you can merge them now if you like.

Still would like to know why that file caused me issues though?

george-gca:
> Still would like to know why that file caused me issues though?

Tbh I have no idea. Try checking out that file, maybe it will stop displaying as an issue?

george-gca (changes requested):
Can you add this information in alphabetical order in both files? Maybe keep the ones that link to `coff.ee` together, but also in alphabetical order between them. Also, you need to change more files. Check the newly introduced section in [CONTRIBUTING.md](https://github.com/alshedivat/al-folio/blob/main/CONTRIBUTING.md#adding-new-social-media-information).

PR #3233, opened 2025-07-08
keep dark mode inside the (site.enable_darkmode) conditional

jvgemert:
Hi,
I noticed some weird behavior when I turned dark mode off (some of the text seemed to disappear, it became white :) )

I traced it to: {% if site.enable_darkmode %} in head.liquid
where some darkmode features are place before the if statement.

When putting it inside the if statement then the weird behavior was solved; see the pull request here :)

Cheers,
Jan

### DIFF
--- a/_includes/head.liquid
+++ b/_includes/head.liquid
@@ -69,9 +69,9 @@
 <link rel="stylesheet" href="{{ '/assets/css/main.css' | relative_url | bust_css_cache }}">
 <link rel="canonical" href="{{ page.url | replace:'index.html','' | absolute_url }}">
 
+{% if site.enable_darkmode %}
 <!-- Dark Mode -->
 <script src="{{ '/assets/js/theme.js' | relative_url | bust_file_cache }}"></script>
-{% if site.enable_darkmode %}
   <link
     defer
     rel="stylesheet"


george-gca (approved):

george-gca:
Just missing the prettier complaints.

jvgemert:
> Just missing the prettier complaints.

Yes! But, I don't know exactly what the goal of that "prettier" action is (?). I thought it was a linter or something, or is it the intention that it fails so that the humans make their code prettier themselves? :). 

(ie: any tips on how I can fix those complaints? :) )

george-gca:
Yes, it is a linter, but I couldn't make it change the code automatically with every PR. This is a limitation (rather security feature) of the GitHub actions itself, since it would involve an action in our repo make changes in your code. Probably a better idea would be to create a pre-commit script, but that wouldn't run if someone make changes using only GitHub web interface.

Currently what the action does is try to validate the code, and if it fails, it creates a zip with an HTML inside pointing the needed changes to make the code compliant. You can check this file inside the failed action. For more on this, check our [FAQ](https://github.com/alshedivat/al-folio/blob/main/FAQ.md#my-code-runs-fine-locally-but-when-i-create-a-commit-and-submit-it-it-fails-with-prettier-code-formatter-workflow-run-failed-for-main-branch-how-do-i-fix-that).

jvgemert:
Ah, sorry for missing it in the FAQ. 

I think that I now fixed it :)

george-gca (approved):

PR #3267, opened 2025-08-16
Moved PurgeCSS to Jekyll Build

### LINKED ISSUE #3266: PurgeCSS not working

### Bug description

PurgeCSS installs and runs, but doesn't do anything

### How to reproduce the bug

Image

This is already happening with your repo. Here is a screenshot from your deploy.yml workflow run. PurgeCSS is empty. I didn't notice at first, but I modeled my own Jekyll site based on al-folio where I realized that PurgeCSS is not working.

### Error messages and logs

```
```

Windows, Linux

Deployed site

### More info

After multiple debugging I have determined that it's not using the config properly at all. My solution is to inline the config when running purgeCSS.

```cmd
purgecss --content './_site/**/*.html' './_site/**/*.js' --css './_site/assets/styles/*.css' --output './_site/assets/styles/'
```

MUmarShahbaz:
Fixes #3266 

# Purpose

Issue: PurgeCSS was not working with `purgecss.config.js`, only with inline config

Solution: Created a Ruby Gem to run PurgeCSS using Configuration defined inside `_config.yml`.

Now, all assets optimization configs (minifier, terser, purgecss) are together allowing ease of use.

# Example of use

## Before

Opera Snapshot_2025-08-16_095904_github com

## After

Opera Snapshot_2025-08-16_100042_github com

## New Logs - Helps in debugging

Opera Snapshot_2025-08-16_095717_github com

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -247,8 +247,8 @@ terser:
 # -----------------------------------------------------------------------------
 
 PurgeCSS:
-  content: ['_site/**/*.html', '_site/**/*.js']
-  css: ['_site/assets/css/*.css']
+  content: ["_site/**/*.html", "_site/**/*.js"]
+  css: ["_site/assets/css/*.css"]
   output: _site/assets/css/
 
 # -----------------------------------------------------------------------------


MUmarShahbaz:
@george-gca here's another one!

MUmarShahbaz:
Note: The Netlify Errors occur because you are installing purgecss post-build to run with `purgecss.config.js`. With this gem you will need it to be installed pre-build.

Size Stats:
| Stat | Before | After |
|:-:|:-:|:-:|
| CSS Directory Size | 3.80 MB | 2.81 MB |
| CSS Directory URL | [here](https://github.com/alshedivat/al-folio/tree/gh-pages/assets/css) | [here](https://github.com/MUmarShahbaz/al-folio/tree/gh-pages/assets/css) |
| Workflow run | - | [here](https://github.com/MUmarShahbaz/al-folio/actions/runs/17004683801) |

george-gca:
Will definitely took a look at this.

george-gca:
@MUmarShahbaz sorry for the (VERY LONG) delay on checking this. Question: we have in our purge css config this line:

```
skippedContentGlobs: ["_site/assets/**/*.html"]
```

This is used to avoid purging inline css, if I am not wrong, for htmls generated by jupyter notebooks. How this is being handled in your case? Does it displays correctly the notebook from the sample, both light and dark modes?

MUmarShahbaz:
@george-gca that's what I thought earlier too. For some reason, it has just a ghost presence. It doesn't do anything. You can see from the huge difference in size when running my plugin as compared to the current setup.

Right now, I'm trying to mimic the customization options through manual logic in ruby and there are some kinks, I admit, but overall it's working.

I spent hours, using your repo for reference, to add purgecss to mine. After failing many times, it turned out that your's isn't working either.

MUmarShahbaz:
As for the skipped content blobs, it basically never touches them. Underneath, it uses logic to manually single out and purge each file one at a time. You can see them in the logs as it goes them all.

alshedivat:
Closing as superseded. Conflicts with v1 starter ownership model; runtime build ownership moved to owning gems.

If you still want to pursue this area in v1, please open a fresh issue/PR against the owning plugin repo or current starter boundary, and reference this PR for context.

PR #3270, opened 2025-08-16
Make Giscus light and dark themes configurable via _config.yml

### LINKED ISSUE #3269: Giscus theme option doesn't work

### Bug description

Currently the giscus theme is computed in `giscus.liquid` irrespective of the `site.giscus.theme`. 

### How to reproduce the bug

Change `giscus.theme` in `_config.yml` and the giscus theme does not change because the option is ignored

### Error messages and logs

```
```

Running locally with Docker (docker compose), Deployed site

### More info

It is a simple bug which can be solved by adding `dark_theme` and `light_theme` options to `_config.yml` and using these in `giscus.liquid` based on current theme. I can submit a PR if required.

neuroconvergent:
Currently, the `site.giscus.theme` option is ignored because `giscus.liquid` always computes the theme as light/dark based on the current site theme. This PR allows users to configure separate light and dark giscus themes in `_config.yml`, which will support dynamic updates when switching between light, dark, or system themes.

Fixes #3269

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -110,7 +110,8 @@ giscus:
   strict: 1 # use strict identification mode
   reactions_enabled: 1 # enable (1) or disable (0) emoji reactions
   input_position: bottom # whether to display input form below (bottom) or above (top) the comments
-  theme: preferred_color_scheme # name of the color scheme (preferred works well with al-folio light/dark mode)
+  dark_theme: dark # name of the dark color scheme (preferred works well with al-folio dark mode)
+  light_theme: light # name of the light color scheme (preferred works well with al-folio light mode)
   emit_metadata: 0
   lang: en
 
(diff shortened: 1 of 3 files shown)


george-gca:
Just out of curiosity, are there other giscus themes available? Or is this just to support custom build themes?

neuroconvergent:
There are several official [Giscus themes](https://github.com/giscus/giscus/tree/main/styles/themes) available and alternatively custom themes can be set by changing the theme name to the URL of a custom stylesheet.

george-gca:
I didn't know about that. Now, let's do another improvement since we are about to change this part. Can you make it work like it is currently done for google analytics, for example? This way, the generated js code can be properly minified.

Check:
- https://github.com/alshedivat/al-folio/blob/main/_scripts/google-analytics-setup.js
- https://github.com/alshedivat/al-folio/blob/main/_includes/scripts.liquid#L226

neuroconvergent:
I have made the changes and moved the giscus setup to a different `.js` file.

george-gca (approved):

george-gca:
Thanks for your contribution.

PR #3272, opened 2025-08-17
Fix for the CV page's date sorting of work history, education, and volunteering

### LINKED ISSUE #3271: Sorting Issues on CV Page

### Bug description

Job history, volunteering, and education sections of the CV page do not sort correctly. Currently, they sort from latest to earliest start date only.

### How to reproduce the bug

1. Go to the CV page
2. Observe the entry order for the job, education, and volunteering sections.

### Error messages and logs

N/A

Mac

Running locally with Docker (docker compose)

### More info

_No response_

profio1997:
This PR normalizes partial dates (YYYY, YYYY-MM) to full ISO strings for sorting. It replaces the Liquid date filter with string ops so year-only and year-month values display and sort correctly.

Fixes #3271

### DIFF
--- a/_includes/resume/education.liquid
+++ b/_includes/resume/education.liquid
@@ -1,14 +1,57 @@
 <ul class="card-text font-weight-light list-group list-group-flush">
-  {% assign education = data[1] | sort: 'startDate' | reverse %}
-  {% for content in education %}
+  {%- comment -%}
+  Normalize startDate/endDate:
+  - If only year is given ("YYYY"), append "-01-01"
+  - If year+month is given ("YYYY-MM"), append "-01"
+  - Otherwise assume full ISO
+  Then build sortable keys based on endDate (default far-future), then startDate.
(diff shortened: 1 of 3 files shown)


george-gca:
I like your solution, but since this code is repeated in multiple files, maybe it is a good idea to create a plugin for it. Maybe create a new liquid filter that alters the date value, something like [_plugins/cache-bust.rb](https://github.com/alshedivat/al-folio/blob/main/_plugins/cache-bust.rb) does with urls. What do you think?

alshedivat:
@codex, can you look into refactoring changes proposed in this PR to avoid code repetition (consider adding a plugin to `_plugins/` for sorting by date)?

alshedivat:
Thanks for the contribution. Under the v1 starter modularization, this change is owned by `al-org-dev/al-folio-cv`.

Tracking issue opened: https://github.com/al-org-dev/al-folio-cv/issues/3

Issue #3256, opened 2025-07-29
Sass deprecation warnings


fverdoja:
### Have you checked that your issue isn't already filed?

- [x] I read through [FAQ](https://github.com/alshedivat/al-folio/blob/main/FAQ.md) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [x] Yes, I have checked that this issue isn't already filed.

### Bug description

Dart Sass has deprecated the use of @import and global built-in functions, and now al-folio throws deprecation warnings

### How to reproduce the bug

When launching the current version of al-folio, I receive the attached deprecation warnings. It seems they are all related to [this breaking change in Dart Sass](https://sass-lang.com/d/import).

Everything seems to still work fine for the time being, but I thought it might be good to report this so it can be addressed in time before switching to 3.0.0.

I see these warnings both on my local devcontainer and on github pages.

### Error messages and logs

```
Deprecation Warning [import]: Sass @import rules are deprecated and will be removed in Dart Sass 3.0.0.

More info and automated migrator: https://sass-lang.com/d/import

  ╷
7 │   "variables",
  │   ^^^^^^^^^^^
  ╵
    /workspaces/al-folio/assets/css/main.scss 7:3  root stylesheet
Deprecation Warning [import]: Sass @import rules are deprecated and will be removed in Dart Sass 3.0.0.

More info and automated migrator: https://sass-lang.com/d/import

  ╷
8 │   "themes",
  │   ^^^^^^^^
  ╵
    /workspaces/al-folio/assets/css/main.scss 8:3  root stylesheet
Deprecation Warning [import]: Sass @import rules are deprecated and will be removed in Dart Sass 3.0.0.

More info and automated migrator: https://sass-lang.com/d/import

  ╷
9 │   "layout",
  │   ^^^^^^^^
  ╵
    /workspaces/al-folio/assets/css/main.scss 9:3  root stylesheet
Deprecation Warning [import]: Sass @import rules are deprecated and will be removed in Dart Sass 3.0.0.

More info and automated migrator: https://sass-lang.com/d/import

   ╷
10 │   "base",
   │   ^^^^^^
   ╵
    /workspaces/al-folio/assets/css/main.scss 10:3  root stylesheet
Deprecation Warning [import]: Sass @import rules are deprecated and will be removed in Dart Sass 3.0.0.

More info and automated migrator: https://sass-lang.com/d/import

   ╷
11 │   "distill",
[...]

PR #3281, opened 2025-08-27
Update SASS syntax in base (non-font) files

scottleechua:
Partially fixes issue #3256 by updating SASS syntax in main SCSS files.

To fully address the issue, FontAwesome should be updated to `v7.x` and Tabler to `v3.30` or higher (see Tabler fix [here](https://github.com/tabler/tabler-icons/pull/1256)), where the SCSS has been fixed.

### DIFF
--- a/_sass/_base.scss
+++ b/_sass/_base.scss
@@ -2,6 +2,8 @@
  * Styles for the base elements of the theme.
  ******************************************************************************/
 
+@use "variables" as v;
+
 // Typography
 
 p,
@@ -629,12 +631,12 @@ footer.sticky-bottom {
       padding: 0.75rem 1.15rem;
 
       &:hover {
-        color: $black-color;
+        color: v.$black-color;
       }
     }
 
     &.active .page-link {
-      color: $white-color;
+      color: v.$white-color;
       background-color: var(--global-theme-color);
 
       &:hover {
@@ -817,7 +819,7 @@ footer.sticky-bottom {
         padding-bottom: 0.5rem;
         span {
           display: inline-block;
-          color: $black-color;
+          color: v.$black-color;
           height: 100%;
           padding-right: 0.5rem;
           vertical-align: middle;
--- a/_sass/_layout.scss
+++ b/_sass/_layout.scss
@@ -2,6 +2,8 @@
  * Content
  ******************************************************************************/
 
+@use "variables" as v;
+
 body {
   padding-bottom: 70px;
   color: var(--global-text-color);
@@ -28,7 +30,7 @@ body.sticky-bottom-footer {
 }
 
 .container {
-  max-width: $max-content-width;
+  max-width: v.$max-content-width;
 }
 
 // Profile
--- a/_sass/_themes.scss
+++ b/_sass/_themes.scss
@@ -1,34 +1,34 @@
 /*******************************************************************************
  * Themes
  ******************************************************************************/
-
+@use "variables" as v;
 @use "sass:color";
 
 :root {
   color-scheme: light;
-  --global-bg-color: #{$white-color};
-  --global-code-bg-color: #{$code-bg-color-light};
-  --global-text-color: #{$black-color};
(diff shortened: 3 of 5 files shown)


george-gca:
I updated FontAwesome in this PR, but Tabler Icons is not easily downloadable as a package anymore, at least I couldn't find it in their site a way do download a built package without paying. I am currently considering the options:
- remove tabler icons from code, since it is only used for 4 icons
- create something to build the package myself

scottleechua:
I'm in favor of removing Tabler and replacing the search and sun/moon with FontAwesome similar icons. I'm sure there was a good reason to opt for the Tabler version instead, but perhaps the maintenance cost is not equivalent to the benefit at this point.

george-gca:
Tbh the main reason Tabler Icons was added was because of the "follow system theme" option. Font Awesome doesn't have a free icon that translates this information. I was thinking of using the [desktop](https://fontawesome.com/icons/desktop?f=classic&s=solid) icon, but I am not sure it translates the information correctly. What do you think?

scottleechua:
I've implemented a "half and half" sun moon icon to represent the system setting, using Fontawesome sun and moon. I've seen something like this done elsewhere, though I can't remember exactly where, and it seems communicative enough.

If that works for you, the path should be clear for you now to replace the other Tabler icons and remove the package.

On hover:
Screenshot 2025-09-02 at 08 43 38

Without hover:
Screenshot 2025-09-02 at 08 43 36

george-gca:
I think I will ditch tabler icons entirely, but think a little bit more about the system icon. your suggestion is actually clipped on the right side.

scottleechua:
Good flag @george-gca  — updated the PR to make the half-and-half icon look better:

Screenshot 2025-11-18 at 09 24 21
Screenshot 2025-11-18 at 09 24 40

george-gca:
Can we keep it without these colors (blue and purple) and only the dark/light version like the others?

scottleechua:
Those colors are already implemented in the icons on hover — got the below screenshots from the [demo site](https://alshedivat.github.io/al-folio/).

The only change I made was in the "follow system" mode, I allowed the moon and sun respectively to retain their own colors.

Screenshot 2025-11-22 at 12 46 43
Screenshot 2025-11-22 at 12 46 49

Screenshot 2025-11-22 at 12 46 39

george-gca:
Sorry for the VERY LONG delay on reviewing this @scottleechua. I didn't have enough time before to properly check this PR, since it is a big one, but I am finally having some time to do so. I rebased your code with the latest changes, fixed the prettier complaints, now we are finally getting ready to finish it. Just missing a few things:
- can we make it so that both sun and moon (when displaying both at the same time) keep the same color?
- when changing theme, the delay is weird. Compare the one on the deployed site with running one locally with this PR. Think you might be able to fix that?

scottleechua:
Heya @george-gca , no worries at all!

I addressed your first comment about the color of the "system theme" half sun + half moon.
On hover, both halves of the icon are now a unified **purple**.

I unfortunately couldn't replicate the timing delay you mentioned — if anything, the local PR version is a bit faster than the demo site on my (admittedly slow) internet. It's possible FontAwesome takes slightly longer to load than Tabler? But I don't think I can address that one.

Cheers,
Scott

george-gca:
The main issue with the sun/moon is when your system theme is dark. Look at my case, when I highlight the sun and moon icon, it is becoming purple while the accent color should be blue in this case.

[Screencast_01_23_2026_12:11:46_PM.webm](https://github.com/user-attachments/assets/fd5b60dc-223c-4d06-bd6a-8995ad46e284)

scottleechua:
Ah, thanks for catching that, @george-gca !

george-gca:
Thanks for your contribution @scottleechua. This was kind of a headache for me to update by myself, but with your help we finally did this. There will be one day that we might update all of our dependencies (specially bootstrap).

Issue #3280, opened 2025-08-25
Incorrect spacing between LeetCode icon and adjacent icons; RSS icon displayed despite being disabled in configuration


onixlas:
### Have you checked that your issue isn't already filed?

- [x] I read through [FAQ](https://github.com/alshedivat/al-folio/blob/main/FAQ.md) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [x] Yes, I have checked that this issue isn't already filed.

### Bug description

Incorrect spacing between LeetCode icon and adjacent icons; RSS icon displayed despite being disabled in configuration 

### How to reproduce the bug

On the ["About" page of my al-folio-based site](https://onixlas.github.io/), two issues are observed in the social icons section: 

* **Excessive Spacing:** The LeetCode icon is visually separated from the adjacent icons (Kaggle and LinkedIn) by a larger gap compared to the spacing between other consecutive icons. This inconsistency disrupts the visual alignment and uniformity of the social links block. 
* **RSS Icon Visibility Despite Disabled Setting:** The RSS feed icon is being displayed in the social icons section even though the configuration in `_data/socials.yml` explicitly sets `rss_icon: false`.

## Steps to Reproduce:

1. Deploy an al-folio site with the following configuration in `_data/socials.yml`:

 ```yml
email: example@example.com
github_username: example
kaggle_id: example
leetcode_id: example
linkedin_username: example
rss_icon: false
telegram_username: example
```

2. Navigate to the "About" page of the site. 
3. Observe the spacing between the LeetCode icon and its neighboring icons (Kaggle and LinkedIn). 
4. Check whether the RSS icon is visible in the social icons section.  

## Expected Behavior: 

1. All social icons should be evenly spaced with consistent margins.
2. The RSS icon should not be displayed when rss_icon: false is set in _data/socials.yml.

## Actual Behavior: 

1. The LeetCode icon has a noticeably larger gap on both sides compared to other icons.
2. The RSS icon is still rendered despite being disabled in the configuration.
 

### Error messages and logs

```
The error message you got, with the full traceback if available. Please paste it between these triple backticks.
```

### What operating system are you using?

Linux

### Where are you seeing the problem on?

Deployed site

### More info

<img width="1118" height="253" alt="Image" src="https://github.com/user-attachments/assets/b1f79ab9-8007-4d69-91d9-ad74539b696a" />

onixlas:
[...]

PR #3282, opened 2025-08-27
Socials icons: Check if rss_icon: false

scottleechua:
Partially addresses issue #3280.

Currently, the RSS feed icon in the socials section will show whenever `rss_icon` in `data/_socials.yml` has a value --- **even if that value is `false`**.

Unlike the other social settings, which are empty by default, for `rss_icon` the default setting is `true`.

So it makes sense that a user might expect `rss_icon: false` to disable said icon.

This PR just adds a check if the value is `false` and hides the icon if so.


george-gca:
Thanks for the fix.

Issue #3095, opened 2025-03-24
Google analytics not working


Randl:
### Have you checked that your issue isn't already filed?

- [x] I read through [FAQ](https://github.com/alshedivat/al-folio/blob/main/FAQ.md) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [x] Yes, I have checked that this issue isn't already filed.

### Bug description

Google analytics data is not flowing

### How to reproduce the bug

1. Enable google analytics
2. It doesn't work

### What operating system are you using?

Linux, Not applicable (e.g. you're using GitHub Pages or other hosting)

### Where are you seeing the problem on?

Deployed site

### More info

I think it was broken in https://github.com/alshedivat/al-folio/pull/2571

Specifically I think in 
https://github.com/alshedivat/al-folio/blob/main/assets/js/google-analytics-setup.js#L6
the value is not substituted, at least when I check the JS in deployed version.

george-gca:
Thanks for reporting this. You are right. I'll fix this.

PR #2962, opened 2025-01-17
Updated scripts for correctly using gtag.

sr-dash:
Implementation of Google tag ID to correctly getting tracked requires further modification in addition to the one line script in the `_includes/scripts.liquid` file as
```
<script async src="https://www.googletagmanager.com/gtag/js?id={{ site.google_analytics }}"></script>
  <script>
    window.dataLayer = window.dataLayer || [];
    function gtag() {
      dataLayer.push(arguments);
    }
    gtag('js', new Date());

    gtag('config', '{{ site.google_analytics }}');
  </script>
```

I have made the changes and tested it out on my personal webpage to ensure the tracking in analytics dashboard.

PR #3287, opened 2025-08-30
Revert duplicate Google Analytics

sim642:
This reverts PR #2962. The inline script here is exactly the same as the `/assets/js/google-analytics-setup.js` sourced on the following line, so Google Analytics is being triggered twice.

It seems that #2962 originally added this because liquid templating of the setup script didn't work (#3095). However, this was then properly fixed in #3117.


george-gca:
Thanks for your contribution.

PR #3289, opened 2025-08-30
Revert to new jekyll-minifier for terser support

### LINKED ISSUE #3286: Remove jekyll-terser dependency
`jekyll-minifier` has recently been updated to also use terser itself, so this should be reverted to some extent at least.

Because al-folio still depends on jekyll-minifier as well, so now it has two different plugins running terser, which is completely redundant.

_Originally posted by @sim642 in https://github.com/alshedivat/al-folio/issues/2571#issuecomment-3237239010_

sim642:
Closes #3286.

I have only briefly tested this in my own gem-based fork of the theme and it seemed to work. But this probably should be tested more thoroughly, especially to see if this fixes the problems that forced the switch away from jekyll-minifier JS minification in the first place (I don't think I ever encountered an issue to begin with).

### TODO
- [ ] Test thoroughly.
- [ ] Update Gemfile.lock after merge. I haven't done it here to avoid creating conflicts.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,7 +11,7 @@ group :jekyll_plugins do
     gem 'jekyll-imagemagick'
     gem 'jekyll-jupyter-notebook'
     gem 'jekyll-link-attributes'
-    gem 'jekyll-minifier', '>= 0.2.0' # 0.2.0 replaces Uglifier with Terser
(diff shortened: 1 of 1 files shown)


george-gca:
I can do the update to the Gemfile.lock later. It will also require a new release from our side.

alshedivat:
Closing as superseded. Closing for now and tracking dependency policy separately for post-v1 follow-up.

If you still want to pursue this area in v1, please open a fresh issue/PR against the owning plugin repo or current starter boundary, and reference this PR for context.

PR #3293, opened 2025-09-01
fix(workflows): restrict permissions to contents: read

onixlas:
This PR updates the `.github/workflows/broken-links-site.yml` workflow to explicitly set minimal permissions.

- Added: 
  ```yaml 
  permissions: 
    contents: read
  ```
- This adheres to the principle of least privilege.
- Prevents potential write access to the repository that is not required.

### DIFF
--- a/.github/workflows/broken-links-site.yml
+++ b/.github/workflows/broken-links-site.yml
@@ -4,7 +4,7 @@ on:
   workflow_run:
     workflows: [Deploy site]
     types: [completed]
-    
+
 permissions:
   contents: read
 


george-gca:
I believe this can also be set to [.github/workflows/broken-links.yml](https://github.com/alshedivat/al-folio/blob/main/.github/workflows/broken-links.yml).

onixlas:
> I believe this can also be set to [.github/workflows/broken-links.yml](https://github.com/alshedivat/al-folio/blob/main/.github/workflows/broken-links.yml).

Yes, sure. I've made changes only to `.github/workflows/broken-links-site.yml`, because it's my first pull request to this project and I wanted to keep it small. In the second commit I've applied the same changes to `.github/workflows/broken-links.yml`

george-gca (approved):

PR #3329, opened 2025-10-11
[MINOR] Update bib_search.liquid to use input type `search`

### LINKED ISSUE #3328: Can't clear search field in bib search

### Bug description

The input field uses type text instead of search so it's missing the usual `X` to clear the field when text is input.

### How to reproduce the bug

N/A

### Error messages and logs

N/A

Deployed site

### More info

_No response_

cakiki:
Switches the input type from text to search to add an X to clear the search box from text.

Fixes #3328

Sorry if I'm missing something obvious with this. I don't have much webdev experience.

### DIFF
--- a/_includes/bib_search.liquid
+++ b/_includes/bib_search.liquid
@@ -1,4 +1,4 @@
 {% if site.bib_search %}
   <script src="{{ '/assets/js/bibsearch.js' | relative_url | bust_file_cache }}" type="module"></script>
-  <input type="text" id="bibsearch" spellcheck="false" autocomplete="off" class="search bibsearch-form-input" placeholder="Type to filter">
+  <input type="search" id="bibsearch" spellcheck="false" autocomplete="off" class="search bibsearch-form-input" placeholder="Type to filter">
 {% endif %}


george-gca:
The only issue is now with the color of the X, which is barely appearing on dark mode. Can you fix that @cakiki?

cakiki:
@george-gca Sorry for the ping, but I wanted to make sure I was looking at the correct way to do this. Is this what I should be using to style the `X`: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/::-webkit-search-cancel-button ?

Found it here: https://stackoverflow.com/questions/47041672/change-the-color-of-the-clear-button-x-in-input-type-search

george-gca:
@cakiki sorry for the late reply. Not really sure, but it appears that this might be it.

alshedivat:
Thanks for the contribution. Under the v1 starter modularization, this change is owned by `al-org-dev/al-citations`.

Tracking issue opened: https://github.com/al-org-dev/al-citations/issues/3

PR #3350, opened 2025-10-31
Fixed/publications fields

priyapathak39:
The issue where publication entries displayed only a limited number of fields has been resolved by modifying the publication layout template.

Key Change: The fields for abstract, award, and video are now rendered directly and visibly within the main publication entry.

Result: Users no longer need to click a separate button to see this important information, ensuring a more complete and visible subset of publication data is immediately available to the reader.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -1,13 +1,8 @@
-# TODO for Issue #3328: Can't clear search field in bib search
+# TODO: Show More Fields in Publications
 
-## Tasks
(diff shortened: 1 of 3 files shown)


CheariX:
Thank you for this idea.
For me, showing all this information is too overwhelming, especially if you have many publications (all my entries include the abstract).
 I really like the button-toggle approach.

alshedivat:
judging by the preview, I don't think this change makes sense. hiding unnecessary information was an intentional design decision. closing.

Issue #3252, opened 2025-07-22
Make quote blocks dimmer in dark mode


zhuyuhui97:
### Have you checked that your feature request isn't already filed?

- [x] I read through [FAQ](https://github.com/alshedivat/al-folio/blob/main/FAQ.md) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my feature request.
- [x] Yes, I have checked that this feature request isn't already filed.

### Description & Motivation

The quote blocks are always dazzling in dark mode. They have same colors in light mode and dark mode. It would be great to have a separated definition of colors.

Example in [blog/2023/custom-blockquotes](https://alshedivat.github.io/al-folio/blog/2023/custom-blockquotes/)

<img width="2560" height="1390" alt="Image" src="https://github.com/user-attachments/assets/e2e66e4e-5c9d-4b20-bab9-216ad365e38d" />

It can be like this:

<img width="2560" height="1390" alt="Image" src="https://github.com/user-attachments/assets/22aa30ae-26e9-4f80-8ce1-b90fc29f7d2a" />

### Pitch

Change the color definitions of the quote blocks in dark mode, so these blocks can get darker and more friendly to eyes.

divyam-r25:
Hii @zhuyuhui97, I would like to work on this issue,can you please assign me this issue

arorasneha08:
Hii @zhuyuhui97 , Could you please assign this issue to me?

zhuyuhui97:
Hi @@divyam-r25 @arorasneha08  , I already have a patch and waiting for maintainer's reply.

PR #3357, opened 2025-11-07
Fix quoteblk color in dark mode

zhuyuhui97:
This PR fixes the issue #3252 which makes the `quoteblk` dazzling in dark mode 🙈✨🥹. This is implemented by defining seperated color values of `quoteblk`s in dark mode thus does not affect the colors in light mode.


george-gca:
Thanks for the fix.

PR #3384, opened 2025-11-29
Added support for wiki-links and obsidian like graph view for blog posts [Feature] 

HrishiDhondge:
Hi, 
This PR introduces two key features to make your blog more interactive:
1. The integration of wiki-links, and
2. The powerful visualization of relationships/connections between posts through the generated graph view

For more details on implementation and utilization, I have created a sample post as well.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -99,6 +99,11 @@ related_blog_posts:
   enabled: true
   max_related: 5
 
+# Graph view
+enable_graph_plugin: true
+graph_field: ["tags", "categories"]
+graph_include_wiki_links: true
+
 # Giscus comments (RECOMMENDED)
 # Follow instructions on https://giscus.app/ to setup for your repo to fill out the information below.
 giscus:
@@ -192,6 +197,7 @@ exclude:
   - README.md
   - readme_preview/
   - vendor
+  - _includes/posts_graph.json
 keep_files:
   - CNAME
   - .nojekyll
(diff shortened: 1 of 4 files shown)


HrishiDhondge:
@george-gca Could you please take a look and let me know in case of any changes needed?
Thanks!

george-gca:
I am curious, what are the use cases you think of to this graph visualization? The graph structure and visuals is nice, but I am not sure if it brings any benefit to the user.

HrishiDhondge:
I agree that by default, a global graph view provides little direct value to readers beyond offering a high-level sense of how posts are interconnected.
The main idea behind this is to render a local graph below each post, something similar to related posts, but visualized as a graph showing their connections. This is particularly useful when an author is writing a series of posts around a shared topic.

george-gca:
I believe this would be more useful if created as a tool that allowed the user to easily create its own graph based on its own data. For example, think that a user might want to display some concepts using a graph, like relationship between animals, he could write some input data in some way (csv, dot from graphviz, idk) and it would create this beautiful visualization. Not sure about related topics though.

alshedivat:
Thanks for the contribution. Under v1 modularization, this feature should be developed as a standalone plugin rather than in the starter repo.

Tracking proposal issue opened: https://github.com/alshedivat/al-folio/issues/3542

I am closing this PR in favor of the plugin proposal path.

PR #3387, opened 2025-11-30
Fix bib / code layout bug

dfuchss:
Resolve #3370
Resolve #3380

Based on the [comment](https://github.com/alshedivat/al-folio/issues/3370#issuecomment-3565962058) of @MarkusThill, I created a bugfix that solved the issue on my website. Hope that fixes it in general; thus the PR :) However, I'm not an expert for scss .. so handle with care :)

PR #3391, opened 2025-12-01
Fix: bibliography month persists to subsequent entries without month

### LINKED ISSUE #3386: Publication page is broken and shows wrong dates

### Bug description

There is a bug (actually two) in the publications part. 

a) ~the bibtex is not properly rendered, if you click on the bib button.~ (EDIT: Found the related issue #3380; so deleted that from this issue)
b) as soon as there was an entry with a set month, all entries below (earlier in time) get this month if they do not have a month in their entry. See entries below "Can Quantum-Mechanical Description of Physical Reality Be Considered Complete?".

### How to reproduce the bug

Go to https://alshedivat.github.io/al-folio/publications/ and check ...

#### ~the layout of the bibtex (basically #3380):~ (EDIT: Found the related issue; so deleted that from this issue)

Image

#### the months of the publications

Image

Image

### Error messages and logs

N/A

Deployed site

### More info

_No response_

dfuchss:
Fixes #3386 (the part about the wrong months; the other part (layouting of bibtex) is fixed in #3387)

## Summary
- Fixes incorrect month shown for bibliography entries that do not specify a `month` in BibTeX.
- Ensures `month`/`year` do not persist between entries in `_layouts/bib.liquid`.

## Root Cause
- In Liquid, captured variables persist across iterations if not reinitialized.
- `entrymonth` and `entryyear` were only set when `entry.month`/`entry.year` existed, causing values from a previous entry to leak into the next one when those fields were missing.

## Changes
- Reinitialize date-related variables per entry before conditional captures:

## Behavior
- Before: Entries without a `month` could display the previous entry's month.
- After: Entries without a `month` display only the `year`; entries with a `month` display the correct month and year.

image


george-gca:
Thanks for the fix.

PR #3403, opened 2025-12-25
add filtered BibTeX keywords

jucheval:
add `dimensions, eprint` and `pmid` to the filtered keywords


george-gca (changes requested):
You should add these sorted alphabetically to the list.

george-gca:
Thanks for your contribution.

PR #3409, opened 2025-12-26
Add auto-orient flag to Imagemagick config

Michaelmvh:
Currently, images generated by Imagemagick do not respect the orientation of the source image. By adding the auto-orient flag, the images are the correct orientation.

Before
image

After
image


george-gca (approved):

george-gca:
Thanks for the fix.

PR #3410, opened 2025-12-26
Add flag to config to disable copyright footer

Michaelmvh:


george-gca:
Tbh I have mixed feelings about this.

I am pro facilitating the user to modify the theme, but I don't think the footer is that much of a burden. This footer is small and gives credit to this repository and also helps other users that thinks that your site looks great and would like to have their own site like yours find a way to do so.

This exact footer is what brought me to find al-folio many years ago. Since then I started contributing to this template and learning a lot about ruby and web dev, so much so that I became a maintainer to this repo. I think I will leave this one out to the original creator @alshedivat.

Michaelmvh:
Thanks for the added context @george-gca . Im fine with closing this PR then

PR #3413, opened 2025-12-27
[Fix] cache bust plugin not loading new scss on deployments

andreyrisukhin:
## The Bug
I created page with custom scss, and noticed my local docker page looked different than my deployed page (no bounding rectangles around tables and UI elements). 

### Development
image

### Deployed GH Page
image

## The Fix
Updating the cache bust plugin to the current dir of scss fixed my issue. Perhaps this plugin was not updated when scss dir was moved? 

### Deployed GH Page
image

### DIFF
--- a/_plugins/cache-bust.rb
+++ b/_plugins/cache-bust.rb
@@ -43,7 +43,7 @@ module Jekyll
     end
 
     def bust_css_cache(file_name)
-      CacheDigester.new(file_name: file_name, directory: 'assets/_sass').digest!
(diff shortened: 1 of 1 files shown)


george-gca:
Funny enough, the `_sass` dir never moved. Just checked, and it is currently working in our repository:

image

What was happening in your case?

alshedivat:
Thanks for the contribution. Under the v1 starter modularization, this change is owned by `al-org-dev/al-folio-core`.

Tracking issue opened: https://github.com/al-org-dev/al-folio-core/issues/10

PR #3421, opened 2026-01-04
Fix typo in Bootstrap small column class usage

caiopetruccirosa:
This PR fixes a recurring typo in multiple Liquid templates where the Bootstrap class `cl-sm` is used instead of `col-sm`.

Because of this typo, the `col-sm` styling was not being applied. All instances have been corrected.


george-gca:
Thanks for the fix.

PR #3462, opened 2026-01-19
Unify jsonresume and RenderCV CVs solutions

### LINKED ISSUE #2787: RenderCV integration

george-gca:
Fixes #2787 as an alternative to #2969. It was getting too cumbersome to have 2 different data sources for CV and also a lot of different layout files, so I decided to unify them all.
Main changes:
- synchronized the information inside RenderCV (`_data/cv.yml`) and JSONResume (`assets/json/resume.json`)
- unified layout files for CV information
- added the option to set the "data source" in the CV page

### DIFF
--- a/assets/rendercv/settings.yaml
+++ b/assets/rendercv/settings.yaml
@@ -1,7 +1,5 @@
 settings:
   render_command:
-    design:
-    locale:
     # Paths must be relative to _data/cv.yaml
     typst_path: ../assets/rendercv/rendercv_output/NAME_IN_SNAKE_CASE_CV.typ
     pdf_path: ../assets/rendercv/rendercv_output/NAME_IN_SNAKE_CASE_CV.pdf


george-gca:
I'd like your comments on this @sinaatalay, @neuroconvergent, @jiuntian.

neuroconvergent:
I think this is probably the best way to integrate rendercv. Only disadvantage I can see is that rendercv tries to be more flexible with the section titles by providing general purpose entry formats which is lost here as the sections are predefined. This was the original suggestion by @jiuntian, but I am not sure how it can be integrated in a unified renderer.

As for actions workflow to automatically generate the CV from `cv.yml`, I have created a fork with the actions at https://github.com/neuroconvergent/al-folio/tree/rendercv. I am attaching the patches to apply the changes to your branch. I think it is best to specify the design, locale and settings for rendercv separately. 
[0001-Update-_includes-cv-awards.liquid.patch](https://github.com/user-attachments/files/24740033/0001-Update-_includes-cv-awards.liquid.patch)
[0002-add-RenderCV-PDF-generation-workflow.patch](https://github.com/user-attachments/files/24740034/0002-add-RenderCV-PDF-generation-workflow.patch)
[0003-fix-rendercv-workflow-paths.patch](https://github.com/user-attachments/files/24740035/0003-fix-rendercv-workflow-paths.patch)

george-gca:
I couldn't apply your patches to my branch. Are they these commits: https://github.com/alshedivat/al-folio/commit/473f84ee4f3978fea2f76d4926bb13feeee988d5 and https://github.com/alshedivat/al-folio/commit/f45a252c8e61260ae963f32cd99ed48e1db96eba?

neuroconvergent:
Yes those are the commits. 

It seems I accidentally generated the patches against this PR instead of your branch. These patches should work.
[0001-add-RenderCV-PDF-generation-workflow.patch](https://github.com/user-attachments/files/24741770/0001-add-RenderCV-PDF-generation-workflow.patch)
[0002-fix-rendercv-workflow-paths.patch](https://github.com/user-attachments/files/24741771/0002-fix-rendercv-workflow-paths.patch)

george-gca:
Just tested with your changes, and it isn't generating a PDF file. Maybe there's something wrong with the settings file? I got the code that is currently in the new action, set up the variables values, and run:

```bash
rendercv render $cv_file --design $design --locale-catalog $locale --settings $settings
╭─ Your CV is ready ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ Rendering...                                                                                                                                               │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

If I run without the rest of the params, it runs correctly:

```bash
rendercv render $cv_file
╭─ Your CV is ready ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ ✓ 80 ms    Generated Typst:           ./_data/rendercv_output/Albert_Einstein_CV.typ                                                                       │
│ ✓ 240 ms   Generated PDF:             ./_data/rendercv_output/Albert_Einstein_CV.pdf                                                                       │
│ ✓ 39 ms    Generated PNGs:            ./_data/rendercv_output/Albert_Einstein_CV_1.png; ./_data/rendercv_output/Albert_Einstein_CV_2.png;                  │
│ ./_data/rendercv_output/Albert_Einstein_CV_3.png                                                                                                           │
│ ✓ 17 ms    Generated Markdown:        ./_data/rendercv_output/Albert_Einstein_CV.md                                                                        │
│ ✓ 6 ms     Generated HTML:            ./_data/rendercv_output/Albert_Einstein_CV.html                                                                      │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

neuroconvergent:
I just tested it and the problem seems to be the `dont_generate_typst` option as the pdf is generated using the `typst` output. The following seems to fix it.
```diff
index 1a26dea8..a6a518a9 100644
--- a/assets/rendercv/settings.yaml
+++ b/assets/rendercv/settings.yaml
@@ -10,7 +10,7 @@ settings:
     png_path: ../assets/rendercv/rendercv_output/NAME_IN_SNAKE_CASE_CV.png
     dont_generate_markdown: true
     dont_generate_html: true
-    dont_generate_typst: true
+    dont_generate_typst: false
     dont_generate_pdf: false
     dont_generate_png: true
   bold_keywords: []
```
Now the output is:
```bash
╭─ Your CV is ready ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ ✓ 62 ms    Generated Typst:           ./_data/../assets/rendercv/rendercv_output/Albert_Einstein_CV.typ                                                                 │
│ ✓ 166 ms   Generated PDF:             ./_data/../assets/rendercv/rendercv_output/Albert_Einstein_CV.pdf                                                                 │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

```

jiuntian:
Hi @george-gca,

Thank you for compiling and refining these ideas. Unifying both options is definitely the right approach; this PR looks great.

Previously, I handled CV PDF rendering during the build stage using a [custom Jekyll plugin](https://github.com/jiuntian/al-folio/blob/main/_plugins/compile-cv.rb), so that the build always have the latest CV, it also re-render when the cv.yml have its content updated, which would be useful for local development. 

However, I agree with @neuroconvergent that using a GitHub Action workflow would be better for easier maintenance. To ensure the latest PDF is always included in the deployment, I suggest making this workflow a pre-requisite for the deployment step.

george-gca:
Thanks for your support @neuroconvergent, now it is generating the PDF correctly. I tried changing the design so it doesn't create `last updated in` and the pagination, but it didn't work. But I think we can tinker with the design a little bit more, for example there's a url that's trespassing the page limits.

@jiuntian there is no need for that. @neuroconvergent added the GitHub action to be triggered by a change in the cv.yml file. Check the latest additions to this PR.

george-gca:
Currently the action builds the cv and adds to the git repo, but I am thinking that a user might not want that, since it might already have its own version of their cv as pdf. What do you think, maybe only add a section in README about that, telling the user to remove or comment the action?

jiuntian:
> @jiuntian there is no need for that. @neuroconvergent added the GitHub action to be triggered by a change in the cv.yml file. Check the latest additions to this PR.

Ok. I missed that commiting a PDF from workflow would also trigger the deployment workflow. 🤣 



> maybe only add a section in README about that, telling the user to remove or comment the action?

It might accidentally overwrite existing CV if the user has the one with same name? Or we could made it an opt-in and add a section in README?

neuroconvergent:
In that case removing the action would be best way. 

However, all this adds is the computational overhead of running the action. If the user has a different PDF file, they can just have the other file in a different location and link it to the CV page and ignore the rendercv PDF. That is the reason why I set the default output path to `assets/rendercv/rendercv_output` instead of `assets/pdf/`.

george-gca:
Just added this information to [CUSTOMIZE.md](https://github.com/george-gca/multi-language-al-folio/blob/rendercv/CUSTOMIZE.md#modifying-the-cv-information). Any other thoughts on this PR? I think it is good enough to be merged into main.

I just really wanted to remove the pagination and last updated in from the generated CV, but it is not working by setting [here](https://github.com/george-gca/multi-language-al-folio/blob/0a24c5bb86b96614602f764e49dd29393e0c597a/assets/rendercv/design.yaml#L9):

```yaml
show_footer: false
show_top_note: false
```

neuroconvergent:
The design is not being applied as there is a problem in the `settings.yaml`. The fields `settings.render_command.design` and `settings.render_command.locale` need to be removed because they are overwriting the design and locale specified in the actions workflow. 

We need to choose between defining the locale and design files either through command  options or through the settings file. If using settings file to specify this, the paths need to be relative to the `_data` directory where `cv.yml` exists.

Aside from that the PR looks good to me.

george-gca:
Removed, but still not working.

george-gca:
This seems to be a bug in rendercv. I opened a bug there, and will accept this PR.

PR #3479, opened 2026-01-25
fix: CV rendering failure without Volunteer field

### LINKED ISSUE #3478: CV rendering fails when `volunteer` field is missing. (Both JSONResume and RenderCV)

### Bug description

The `cv.liquid` concats the work and volunteer field while rendering the CV from both formats, which throws up an error when the fields are missing. 

### How to reproduce the bug

1. Create a `resume.json` or `cv.yml`  without the section `Volunteer`
2. Try to serve the site with Jekyll or Docker compose.

### Error messages and logs

```
jekyll-1  |   Liquid Exception: Liquid error (line 122): concat filter requires an array argument in /srv/jekyll/_layouts/cv.liquid
jekyll-1  | bundler: failed to load command: jekyll (/usr/local/bundle/bin/jekyll)
jekyll-1  | /usr/local/bundle/gems/liquid-4.0.4/lib/liquid/standardfilters.rb:279:in 'Liquid::StandardFilters#concat': Liquid error (line 122): concat filter requires an array argument (Liquid::ArgumentError)

```

Linux, Mac, Windows

Running locally with Docker (docker compose), Running locally with Docker (devcontainer), Running locally without Docker, Deployed site

### More info

A more robust implementation is needed for the concat operation.

I can think of two ways:
1. Performing the concat only if `Volunteer` passes an if condition to ensure it exists and is not empty. 
2. Assigning defaults to the fields to ensure concat doesn't fail. Something like:
```liquid
{% assign empty_array = "" | split: "," %}
{% assign exp = cv.sections.Experience | default: empty_array %}
{% assign vol = cv.sections.Volunteer | default: empty_array %}
{% assign combined_experience = exp | concat: vol %}
```

neuroconvergent:
Fixes #3478

### DIFF
--- a/_layouts/cv.liquid
+++ b/_layouts/cv.liquid
@@ -122,7 +122,7 @@ layout: default
 
         <!-- RenderCV Sections - Merge Experience and Volunteer -->
         {% comment %} First, combine Experience and Volunteer entries {% endcomment %}
-        {% assign empty_array = "" | split: "," %}
+        {% assign empty_array = '' | split: ',' %}
         {% assign exp = cv.sections.Experience | default: empty_array %}
         {% assign vol = cv.sections.Volunteer | default: empty_array %}
         {% assign combined_experience = exp | concat: vol %}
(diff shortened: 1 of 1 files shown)


george-gca:
Thanks for the fix.

PR #3493, opened 2026-01-31
Fix #2598, shift to last_modified_at instead of last_updated.

### LINKED ISSUE #2598: `last_updated` vs `last_modified_at` for posts

### Description & Motivation

#2341 added the `last_updated` field to posts. However, jekyll-sitemap, which is included here, uses `last_modified_at` instead. This requires duplicating the date in two fields.

### Pitch

Transition to `last_modified_at` field instead:
1. Use `last_modified_at` to output what `last_updated` is used for right now.
2. Direct new users to use `last_modified_at` for that purpose.

For backwards-compatibility `last_updated` should be used as fallback.

### Alternatives

_No response_

### Additional context

_No response_

jsmaskeen:
Use `last_modified_at` rather than `last_updated`. So that the jekyll-sitemap plugin automatically looks for the standard `last_modified_at` front matter field to generate the `<lastmod>` tag in your `sitemap.xml`.

It fixes [#2598](https://github.com/alshedivat/al-folio/issues/2598) 

Also, it preserves backward compatibility, as the updated metadata logic prioritizes the `page.last_modified_at`, and defaults to `page.last_updated`.

### DIFF
--- a/_layouts/post.liquid
+++ b/_layouts/post.liquid
@@ -19,7 +19,8 @@ layout: default
     <p class="post-meta">
       Created on {{ page.date | date: '%B %d, %Y' }}
       {% if page.author %}by {{ page.author }}{% endif %}
-      {% if page.last_updated %}, last updated on {{ page.last_updated | date: '%B %d, %Y' }}{% endif %}
+      {% assign last_date = page.last_modified_at | default: page.last_updated %}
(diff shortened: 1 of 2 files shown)


george-gca:
Thanks for your contribution.

PR #3501, opened 2026-02-05
Add support for Simple Analytics

daniel-gomm:
This PR adds support for [Simple Analytics](https://www.simpleanalytics.com), an EU-based, privacy-first alternative to google analytics that is compliant with GDPR, PECR, and CCPA without requiring consent to cookies.

Implementation includes a new field in the configuration (`site.enable_simple_analytics`), the insertion of the analytics script when this field is set to `true`, and an update to the documentation.

To follow the implementation of the other analytics providers, I have added support for cookie-consent, though this is not strictly necessary for Simple Analytics under GDPR (https://www.simpleanalytics.com/#why) since it does not use cookies or collect any PII data.

### DIFF
--- a/ANALYTICS.md
+++ b/ANALYTICS.md
@@ -12,6 +12,7 @@ This guide helps you add website analytics to track visitor statistics and behav
   - [Privacy-Friendly Alternatives](#privacy-friendly-alternatives)
     - [Pirsch Analytics](#pirsch-analytics)
     - [Openpanel Analytics](#openpanel-analytics)
+    - [Simple Analytics](#simple-analytics)
   - [Monitoring &amp; Performance](#monitoring--performance)
     - [Cronitor](#cronitor)
   - [GDPR and Privacy Considerations](#gdpr-and-privacy-considerations)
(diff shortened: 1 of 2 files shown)


daniel-gomm:
Following-up on this PR:
Given that, to the best of my knowledge, [Simple Analytics](https://www.simpleanalytics.com/#why), [Pirsch](https://pirsch.io/privacy-friendly-analytics), and [OpenPanel](https://openpanel.dev/), do not use tracking cookies and thus do not require cookie-consent (under GDPR and other privacy policies), it seems to me that the documentation should reflect this to provide users of the template the means to make a more informed choice in how to configure their website.

george-gca:
Recently, I added support for the GDPR cookie consent in #3492, but to be honest I don't have enough knowledge about how these cookies work and how to enable/disable then correctly. For example, I have no idea who uses tracking cookies and who doesn't. If you have this information and could add to the documentation (preferably pointing to the source of the information) would be much appreciated.

daniel-gomm:
> Recently, I added support for the GDPR cookie consent in #3492, but to be honest I don't have enough knowledge about how these cookies work and how to enable/disable then correctly. For example, I have no idea who uses tracking cookies and who doesn't. If you have this information and could add to the documentation (preferably pointing to the source of the information) would be much appreciated.

I've read into the topic a bit and updated the documentation accordingly, linking to the respective parts of the documentation of the different analytics providers. `ANALYTICS.md` did already specify which analytics providers require cookie consent. I've also included a note on the requirement to ask for cookie consent if third-party contents that themselves require consent (e.g., YouTube videos,...) are embedded on the site. This is covered, for instance, by [this article](https://dr-dsgvo.de/youtube-videos-und-die-dsgvo-en/).

alshedivat:
Thanks for the contribution. Under the v1 starter modularization, this change is owned by `al-org-dev/al-analytics`.

Tracking issue opened: https://github.com/al-org-dev/al-analytics/issues/2

PR #3503, opened 2026-02-07
[unbundle] extract search wiring into al_search plugin

alshedivat:
## Summary
- add `al_search` gem dependency from `al-org-dev/al-search`
- register `al_search` in `_config.yml` plugin list
- replace hardcoded search asset block in both script includes with `{% al_search_assets %}`

## Notes
- plugin repo: https://github.com/al-org-dev/al-search
- this PR focuses on integration wiring and keeps current theme files in place to avoid disruptive file deletions

### DIFF
--- a/_includes/distill_scripts.liquid
+++ b/_includes/distill_scripts.liquid
@@ -216,7 +216,15 @@
   </script>
 {% endif %}
 
-{% if site.plugins contains 'al_search' %}
+{% if site.search_enabled %}
   <!-- Search -->
-  {% al_search_assets %}
+  {% if site.plugins contains 'al_search' %}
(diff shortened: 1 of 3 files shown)


alshedivat:
Addressed the review feedback.\n\n- Restored  gating in both script includes.\n- Added a parse-safe wrapper include for  plus legacy fallback markup when the plugin is not enabled.\n\nFix commit: \nValidation run: https://github.com/alshedivat/al-folio/actions/runs/21776379894

alshedivat:
Correction: fixed in commit a6975fae. Restored search_enabled gating and added parse-safe wrapper/fallback for al_search_assets. Validation run: https://github.com/alshedivat/al-folio/actions/runs/21776379894

PR #3504, opened 2026-02-07
[unbundle] extract chart and diagram wiring into al_charts plugin

alshedivat:
## Summary
- add `al_charts` gem dependency from `al-org-dev/al-charts`
- register `al_charts` in `_config.yml` plugin list
- replace chart/diagram script-loading blocks in:
  - `_includes/scripts.liquid`
  - `_includes/distill_scripts.liquid`
  with plugin tag `{% al_charts_scripts %}`

## Notes
- plugin repo: https://github.com/al-org-dev/al-charts
- this PR migrates integration wiring while preserving existing theme asset files for compatibility

### DIFF
--- a/_includes/distill_scripts.liquid
+++ b/_includes/distill_scripts.liquid
@@ -18,7 +18,9 @@
 
 {% if site.plugins contains 'al_charts' %}
   <!-- Charts and Diagrams -->
-  {% al_charts_scripts %}
+  {% include plugins/al_charts_scripts.liquid %}
+{% else %}
+  {% include plugins/al_charts_scripts_legacy.liquid %}
 {% endif %}
 
 {% if page.tikzjax %}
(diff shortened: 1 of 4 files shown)


alshedivat:
Addressed the review feedback.\n\n- Replaced direct  calls with a parse-safe include wrapper.\n- Restored the previous inline chart/diagram asset block as a legacy fallback when the plugin is not enabled.\n\nFix commit: \nValidation run: https://github.com/alshedivat/al-folio/actions/runs/21776380200

alshedivat:
Correction: fixed in commit 109effbc. Replaced direct al_charts_scripts usage with parse-safe wrapper and restored legacy fallback scripts. Validation run: https://github.com/alshedivat/al-folio/actions/runs/21776380200

PR #3505, opened 2026-02-07
[unbundle] extract math and tikzjax wiring into al_math plugin

alshedivat:
## Summary
- add `al_math` gem dependency from `al-org-dev/al-math`
- register `al_math` in `_config.yml` plugin list
- replace math/tikzjax asset loading with plugin tags:
  - `{% al_math_styles %}` in `_includes/head.liquid`
  - `{% al_math_scripts %}` in `_includes/scripts.liquid`
  - `{% al_math_scripts %}` in `_includes/distill_scripts.liquid`

## Notes
- plugin repo: https://github.com/al-org-dev/al-math
- this PR focuses on integration wiring while preserving existing theme-side assets for compatibility

### DIFF
--- a/_includes/distill_scripts.liquid
+++ b/_includes/distill_scripts.liquid
@@ -25,7 +25,9 @@
 
 {% if site.plugins contains 'al_math' %}
   <!-- Math and Tikzjax -->
-  {% al_math_scripts %}
+  {% include plugins/al_math_scripts.liquid %}
+{% else %}
+  {% include plugins/al_math_distill_scripts_legacy.liquid %}
 {% endif %}
 
 {% if page.typograms %}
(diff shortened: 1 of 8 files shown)


alshedivat:
Addressed the review feedback.\n\n- Replaced direct  tag calls with parse-safe include wrappers.\n- Added legacy fallbacks in head/scripts/distill scripts, including the original pseudocode runtime path for non-plugin mode.\n\nFix commit: \nValidation run: https://github.com/alshedivat/al-folio/actions/runs/21776380477

alshedivat:
Correction: fixed in commit 13ed64c3. Replaced direct al_math tag usage with parse-safe wrappers and added legacy fallbacks (including pseudocode path in non-plugin mode). Validation run: https://github.com/alshedivat/al-folio/actions/runs/21776380477

PR #3506, opened 2026-02-07
[unbundle] extract comments rendering into al_comments plugin

alshedivat:
## Summary
- add `al_comments` gem dependency from `al-org-dev/al-comments`
- register `al_comments` in `_config.yml` plugin list
- replace layout-level Disqus/Giscus includes with plugin tag `{% al_comments %}` in:
  - `_layouts/post.liquid`
  - `_layouts/page.liquid`
  - `_layouts/distill.liquid`
  - `_layouts/book-review.liquid`

## Notes
- plugin repo: https://github.com/al-org-dev/al-comments
- PR keeps legacy include files in-place for backward compatibility while switching primary rendering path to plugin

### DIFF
--- a/_includes/plugins/al_comments.liquid
+++ b/_includes/plugins/al_comments.liquid
@@ -1,0 +1,4 @@
+{% comment %}
+  Parse-safe wrapper for al_comments custom tag.
+{% endcomment %}
+{% al_comments %}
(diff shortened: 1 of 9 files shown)


alshedivat:
Addressed the review feedback.\n\n- Replaced direct  tag calls with a parse-safe include wrapper.\n- Restored per-layout legacy comment includes (/) when the plugin is not enabled.\n\nFix commit: \nValidation run: https://github.com/alshedivat/al-folio/actions/runs/21776397691

alshedivat:
Correction: fixed in commit 846ab00f. Replaced direct al_comments usage with parse-safe wrapper and restored legacy disqus/giscus fallbacks per layout. Validation run: https://github.com/alshedivat/al-folio/actions/runs/21776397691

PR #3507, opened 2026-02-07
[unbundle] extract newsletter rendering into al_newsletter plugin

alshedivat:
## Summary
- add `al_newsletter` gem dependency from `al-org-dev/al-newsletter`
- register `al_newsletter` in `_config.yml` plugin list
- replace newsletter include/script usage with plugin tags:
  - `_includes/footer.liquid`
  - `_includes/related_posts.liquid`
  - `_layouts/about.liquid`
  - `_includes/scripts.liquid`

## Notes
- plugin repo: https://github.com/al-org-dev/al-newsletter
- PR switches rendering and JS loading to plugin-provided tags while preserving existing theme include file for compatibility

### DIFF
--- a/_includes/footer.liquid
+++ b/_includes/footer.liquid
@@ -19,8 +19,12 @@
   </footer>
 {% else %}
   <footer class="sticky-bottom mt-5" role="contentinfo">
-    {% if site.plugins contains 'al_newsletter' and site.newsletter.enabled %}
-      {% al_newsletter_form align=center %}
+    {% if site.newsletter.enabled %}
+      {% if site.plugins contains 'al_newsletter' %}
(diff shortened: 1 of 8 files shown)


alshedivat:
Addressed the review feedback.\n\n- Replaced direct newsletter custom tag calls with parse-safe include wrappers.\n- Restored legacy newsletter includes/scripts as fallback when the plugin is not enabled.\n\nFix commit: \nValidation run: https://github.com/alshedivat/al-folio/actions/runs/21776412600

alshedivat:
Correction: fixed in commit dccfefc1. Replaced direct al_newsletter tag usage with parse-safe wrappers and restored legacy newsletter fallbacks. Validation run: https://github.com/alshedivat/al-folio/actions/runs/21776412600

Issue #3583, opened 2026-03-30
Fix invalid CSS display value on line 220 and 245 in _layouts/book-review.liquid


maxstridde:
### Have you checked that your issue isn't already filed?

- [x] I read through [FAQ](https://github.com/alshedivat/al-folio/blob/main/FAQ.md) and searched through the [past issues](https://github.com/alshedivat/al-folio/issues), none of which addressed my issue.
- [x] Yes, I have checked that this issue isn't already filed.

### Bug description

Wrong CSS

### How to reproduce the bug

Look at the Css, there are typos.

### Error messages and logs

not applicable.

### What operating system are you using?

Not applicable (e.g. you're using GitHub Pages or other hosting)

### Where are you seeing the problem on?

Deployed site

## Problem 1
Line 220 in `_layouts/book-review.liquid` contains invalid CSS:
```css
display: in-line block; /* in-line block; */
```

should be `inline-block`

## Problem 2
Line 245 in `_layouts/book-review.liquid` contains a typo in a CSS property:
```css
marign: 0px;
```

sidenote: in line 198 is written

```
display: block; /* in-line block; */
```

here the comment is not coherent with the argument.

maxstridde:
I also had a problem with the image reaching into the footer, aka some kind of overflow. 
[Fix on my own Blog](https://github.com/maxstridde/maxstridde.github.io/pull/32) can be seen in this Pull request in my repository. Dont know, if this belongs here or in another issue :).

PR #3598, opened 2026-04-23
Fix CSS display properties and margin typo

maxstridde:
The image would overflow into the footer, these css changes fix the problem. The did only appear, when the text was smaller than the height of the image, so on the al folio example, this issue could not be seen. Its a solution for my raised Issue #3583

Atteched images of the problem on my own site (I did not screenshot the wole page):

before:
overflow

after:
fixed overflow

### DIFF
--- a/_layouts/book-review.liquid
+++ b/_layouts/book-review.liquid
@@ -167,6 +167,9 @@ layout: default
 </div>
 
 <style>
+      .post article {
+        display: flow-root;
+      }
       figure {
         float: none;
         width: auto;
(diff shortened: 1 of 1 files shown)


alshedivat:
Thanks for catching these CSS issues. In v1 the `book-review` layout is owned by the extracted `al_folio_core` plugin, so I ported the fix there instead of merging this PR into the old starter layout path.

The port fixes the invalid `display: in-line block`, the `marign` typo, and contains floated cover figures with `display: flow-root`. It is released as `al_folio_core` 1.0.8 and PR #2968 now pins that version.

Closing this PR as superseded by the plugin release.

PR #3603, opened 2026-05-01
Fix: correctly process external posts to display tags and categories

niebles:
The create_document() function was not getting the right input to correctly parse each post's tags/categories and other content.

With this fix, each external post entry in blog.md will now correctly display tags and categories. Indexing should also improve now.

### DIFF
--- a/_data/citations.yml
+++ b/_data/citations.yml
@@ -1,5 +1,5 @@
 metadata:
-  last_updated: '2026-05-01'
+  last_updated: '2026-05-13'
 papers:
   qc6CJjYAAAAJ:-1WLWRmjvKAC:
     citations: 4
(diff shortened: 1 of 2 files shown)


niebles:
For the issue, see https://alshedivat.github.io/al-folio/blog/

External posts do not display tags/categories even though they are set in _config.yml per commit 81c0556

alshedivat:
Thanks for the fix. For v1 this code is now owned by the extracted `al_ext_posts` plugin rather than the starter repository, so I ported the behavior there instead of merging this PR into the old `_plugins/external-posts.rb` path.

The port preserves source-level defaults while allowing individual RSS entries / explicit posts to override `categories` and `tags`, and includes regression coverage. It is released as `al_ext_posts` 1.0.1 and PR #2968 now pins that version.

Closing this PR as superseded by the plugin release.

PR #3681, opened 2026-07-27
docs: correct factual errors, make AGENTS.md authoritative, and move the showcase out of the README

alshedivat:
A documentation overhaul in four commits. Every claim was verified against the repo, the workflows, or the installed gem before being changed — where a reported defect turned out to be inaccurate, it was skipped and is called out below.

**Docs-only.** No runtime, no `test/style_contract.js`, no CI gating changes.

---

## 1. Factual errors in user-facing guides (`59eecedd`)

| Fix | Detail |
| --- | --- |
| `QUICKSTART.md` | Told users to edit a `theme_color` key in `_config.yml`. No such key exists — in v1.x the palette is a Sass token owned by `al_folio_core` (`--global-theme-color` / `$purple-color`). Now points at `CUSTOMIZE.md#changing-theme-color`. |
| `CUSTOMIZE.md` / `ANALYTICS.md` | Analytics schema corrected. `_config.yml` ships `analytics.{google,cronitor,pirsch,openpanel}` ID values; a provider activates when its **ID** is non-empty. |
| `CUSTOMIZE.md` | `assets/pdfs/` → `assets/pdf/` (contradicted the same section 50 lines later). |
| `CUSTOMIZE.md` | `bin/update_citations.py` → `bin/update_scholar_citations.py`, plus the `scholarly` dependency. |
| `CUSTOMIZE.md` | Three default collections → four (`news`, `projects`, `books`, `teachings`), matching `collections:` in `_config.yml`. |
| `INSTALL.md` | `purgecss` is not a devDependency, so `purgecss -c purgecss.config.js` is not runnable after `npm ci`. Documented installing it (CI installs it globally). |
| `FAQ.md` | Two dead deployment anchors (`#deployment` on FAQ itself, and on the README) → `INSTALL.md#deployment`. |
| `FAQ.md` | Workflow list regenerated: it documented 10 of the 19 workflows. Now complete and grouped by purpose. |
| `CONTRIBUTING.md` | Corrected the claim that `CLAUDE.md` "simply contains `@AGENTS.md`". |

## 2. Agent-facing documentation (`7bb7e5e3`)

The ownership rule was restated in ~9 places with none declaring itself authoritative, while the highest-value operational knowledge lived **only** in `CLAUDE.md` — which non-Claude agents never read.

- **New `docs/ARCHITECTURE.md`** — single home for cross-repo architecture. Leads with *"failure modes that produce no error message"*: silent feature gating, the two-list `Gemfile`/`_config.yml` sync, and the `--baseurl /al-folio` requirement. Also holds the wrapper→tag→gem delegation table (moved out of `CLAUDE.md`, re-verified against each installed gem's `register_tag` calls).
- **`AGENTS.md` rewritten** as the authoritative, tool-neutral entry point: a change-type routing table up top, and an explicit **stop sign** naming the paths that must not exist here.
- **`.github/copilot-instructions.md`** reduced from a verbatim duplicate to a pointer, so the two cannot drift.
- **`CLAUDE.md`** trimmed to Claude-specific content. Three defects fixed along the way: a stale `al_folio_core '= 1.0.9'` pin (Gemfile says `1.0.11` — now tells you to read the Gemfile rather than quoting a version), a hardcoded machine-specific path, and an incorrect description of `bin/setup-python-deps` (it only pip-installs jupyter/nbconvert and never reads `requirements.txt`).
- **Integration test count corrected** from "five" to six everywhere — `test/integration_css_minify.sh` exists and is gated by `unit-tests.yml`. Added the missing `bundle install` to the validated command set.

## 3. Showcase moved out of the README (`ea6473c7`)

The table was 199 lines and ~34% of the README, sitting above both the TOC and "Getting started".

- Extracted verbatim to **`docs/SHOWCASE.md`**. All **190 links diffed before and after: none removed, reordered, or added.** No liveness audit ran in this environment, so nothing was pruned.
- "Getting started" now sits directly under the badges, above the fold.
- **Add-your-site-by-PR is retired.** README, `CONTRIBUTING.md`, the PR template, and the issue chooser now point to the *Showcase* category of GitHub Discussions, worded to read correctly once a dedicated category exists.
- README opening reframed from "a Jekyll theme for academics" to the thin-starter/plugin identity.
- ⚠️ **`docs/SHOWCASE.md` added to the lychee excludes.** Those 190 community links were only unchecked because `README.md` is excluded; moving them to a checked path would have started failing `broken-links.yml`, which runs with `fail: true`.

## 4. Repo hygiene (`082faac9`)

- **`update-tocs.yml`**: `paths:`/`files:` globs were `*.md` / `./*.md`, which never match `docs/*.md` — the root cause of every stale TOC under `docs/`. Now matches both; stale TOCs regenerated.
- **`_config.yml` `exclude:`**: dropped four entries for files that now live in `docs/`; added `AGENTS.md` and `CLAUDE.md`, which were being **copied into `_site/` on every build**.
- **`.prettierignore`**: pruned five entries for gem-owned paths, each verified absent.
- **`_pages/plugins.md`**: `nav: false` was hiding the plugin catalog page the README promotes.

---

## 🔶 Maintainer decision needed: style contract vs. documented overrides

`BOUNDARIES.md` and the FAQ tell **users** that local `_layouts`/`_includes`/`_sass` overrides are supported — but `test/style_contract.js` fails when those directories exist, and that check **ships to every site created from this template**. A user who adds a perfectly legal override sees the starter's own contract check fail in their fork.

Per instructions I **documented the distinction but changed no code**: overrides are legal in a user's own site; the starter repo itself must not contain them. Whether to re-scope the check to the upstream repo only is flagged as an open decision in `ARCHITECTURE.md` and `BOUNDARIES.md`.

## Survey claims I found inaccurate

- **"`enable_google_analytics: true` is a nonexistent key."** Not quite — `al_analytics` **does** read it, as an optional *off-switch*. The real problem is that the flag doesn't enable anything (the ID does), and setting it alone is a no-op. Documented accurately rather than as "nonexistent".
- **"The repo has ~22 workflows."** It has **19** `.yml` workflows plus `schedule-posts.txt`, which is a disabled template, not a workflow.
- **"Showcase PR instructions in README, CONTRIBUTING, PR template, and issue templates."** Only `README.md:34` actually had one. The other locations were updated proactively so the new policy is visible where people would look.

## Validation

| Check | Result |
| --- | --- |
| `npm run lint:prettier` | ✅ pass (after `npx prettier . --write`) |
| `npm run lint:style-contract` | ✅ pass |
| `bundle exec jekyll build --baseurl /al-folio` | ✅ builds (see caveat) |
| `bundle exec al-folio upgrade audit` | ✅ Blocking: 0 |
| `bundle exec al-folio upgrade overrides audit` | ✅ no overrides |
| `bash test/integration_upgrade_cli.sh` | ✅ pass |
| Internal doc links/anchors | ✅ all resolve (scripted check across 16 files) |
| Showcase links | ✅ 190/190 byte-identical |

**Build caveat:** in this sandbox a full build fails inside `al_ext_posts` fetching `medium.com`/`blog.google` through the HTTPS proxy. **Confirmed pre-existing** — clean `origin/main` fails identically. With `external_sources: []` and ImageMagick disabled (binary absent here), the build completes in ~12s. `integration_plugin_toggles.sh` fails for the same network reason. CI has network and should be unaffected.

Verified in the built output: `AGENTS.md`/`CLAUDE.md` no longer leak into `_site/`, and `/plugins/` builds and appears in the navbar.

---


alshedivat:
CI status on `c08957ab`:

**Prettier — fixed.** The failure on `e30ff286` was mine (an em-dash-adjacent `*effective*` that Prettier normalises to `_effective_`). Fixed in `c08957ab`; `npm run lint:prettier` is clean across the whole repo locally.

**Codex review comments — both were right, both addressed in `e30ff286`:**

1. *baseurl.* `_config.yml:22` already sets `baseurl: /al-folio`, and `deploy.yml`, `broken-links-site.yml` and `axe.yml` all run a plain `bundle exec jekyll build`. My claim that omitting the flag breaks the site was simply false. Reframed around the **effective** baseurl in both `AGENTS.md` and `docs/ARCHITECTURE.md`, with the actual failure mode (blanking the baseurl out) called out instead.
2. *`bundle update`.* Also correct — every `al_*` gem is pinned to an exact version, so `bundle update` cannot pick up a newer plugin release. The README now describes bumping the pin. This is not hypothetical: several fixes are currently sitting on gem `main` branches unreleased, and no site can reach them until the gem is published *and* the pin moves.

**link-checker — external, pre-existing, not from this PR.** The only error out of 543 links checked (520 ✅, 20 excluded) is:

```
### Errors in docs/CUSTOMIZE.md
* [502] https://www.altmetric.com/ | Failed: Network error: Bad Gateway
```

That link is on `main` at `docs/CUSTOMIZE.md:802` and this PR does not touch that line — the only `altmetric` text in the diff is the word appearing in a BibTeX field list on the separate line where I corrected `assets/pdfs/` → `assets/pdf/`. It's a 502 from Altmetric's own server, reproduced on two consecutive runs, so it's an upstream outage rather than a dead link.

I've deliberately **not** added it to `.lycheeignore`: if Altmetric is genuinely gone that's a real broken link worth fixing properly, and silencing it here would hide that behind an unrelated docs PR. It should clear on a re-run once their server recovers.

---

PR #3682, opened 2026-07-27
deps: adopt the released plugin gems, disabling the remote Distill loader

alshedivat:
All eight gems that had unreleased work sitting on `main` are now published, so the starter can move off the pins it has been stuck on. They bump together here because one of them changes behaviour that `_config.yml` gates.

| gem | from | to |
| --- | --- | --- |
| `al_folio_core` | 1.0.11 | **1.0.12** |
| `al_folio_cv` | 1.0.0 | **1.0.1** |
| `al_folio_distill` | 1.0.2 | **1.0.3** |
| `al_analytics` | 1.0.0 | **1.0.1** |
| `al_ext_posts` | 1.0.1 | **1.0.3** (1.0.2 was never published either) |
| `al_img_tools` | 1.0.2 | **1.0.3** |
| `al_search` | 1.0.2 | **1.0.3** |
| `al_math` | 1.0.1 | **1.0.2** |

## The Distill loader, and why it is in this PR

`_config.yml` shipped `al_folio.distill.allow_remote_loader: true`. Under 1.0.2 that flag only gated a log line, so it was inert. **1.0.3 turns it into a real switch**: `true` opts the site out of the vendored, hash-pinned Distill runtime and back onto `https://distill.pub/template.v2.js`, fetched with **no subresource integrity**.

Bumping the pin without flipping the flag would ship that gem's entire hardening with its protection turned off — for this site, and for every site generated from this template. So the flag flips here, in the same change as the pin.

This is not a stylistic preference: `al-folio upgrade audit` enforces the pairing. `check_distill_runtime` returns early while the flag is `true`, so splitting the two produces a **blocking** finding. That is also the correction to what I said earlier about #3682 being mergeable on its own — it was not.

Verified on the rendered post at `/blog/2021/distill/`:

```html
<script src="/al-folio/assets/js/distillpub/template.v2.js"   integrity="sha256-R5CDHO0C98TyAJss32l4ztqDUfAGDTtZ3Zs6qxMuJxo=">
<script src="/al-folio/assets/js/distillpub/transforms.v2.js" integrity="sha256-XYVZD1ZSuRCrJBEBl0nIPvWlo/u5tzmtyStFV7a/PTk=">
```

Both local, both integrity-pinned, and no page in the built site references `distill.pub/template`.

## Also here

- **`test/integration_distill.sh`** asserted the 1.0.2 transforms runtime hash. Rehashed against the *published* 1.0.3 gem rather than a computed guess: `70e3f488…` → `5d85590f…`.
- **Dropped the dead `third_party_libraries.polyfill` entry.** `al_math` 1.0.2 stopped emitting the polyfill.io tag, and no gem reads that config key. Confirmed absent from the built site.
- **`test/style_contract.js` hard-coded `al_math '= 1.0.1'`**, so this bump would have failed the style contract. Relaxed to assert an exact released pin of *any* version — the boundary the check actually exists to protect — keeping the separate check that rejects a git-branch pin. Otherwise every routine version bump trips a gate that is meant to guard the thin-starter contract, not the version number.

Left alone deliberately: the `jquery` and `mdb` entries are also unread by any gem, but `al_folio_upgrade` carries a `legacy_jquery_usage` audit rule keyed on `third_party_libraries.jquery`, so removing those wants its own pass.

## Verification

- `bundle install` — all eight resolve from RubyGems
- all six `test/integration_*.sh` — pass
- `npm run lint:style-contract`, `npm run lint:prettier` — pass
- `bundle exec al-folio upgrade audit` — **Blocking: 0**, Non-blocking: 6
- built site smoke-checked: MathJax present, `swiper@12.1.2` (the CVE fix), repo cards on `github-stats-extended` with zero remaining `github-readme-stats` references, CV page renders all sections

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -40,21 +40,21 @@ end
 
 # Gems for al-folio plugins
 group :al_folio_plugins do
-    gem 'al_folio_core', '= 1.0.11'
+    gem 'al_folio_core', '= 1.0.12'
     gem 'al_icons', '= 1.0.0'
-    gem 'al_folio_cv', '= 1.0.0'
-    gem 'al_folio_distill', '= 1.0.2'
+    gem 'al_folio_cv', '= 1.0.1'
+    gem 'al_folio_distill', '= 1.0.3'
     gem 'al_folio_upgrade', '= 1.0.3'
     gem 'al_folio_bootstrap_compat', '= 1.0.0'
     gem 'al_cookie', '= 1.0.0'
 
-    gem 'al_analytics', '= 1.0.0'
+    gem 'al_analytics', '= 1.0.1'
     gem 'al_citations', '= 1.0.1'
-    gem 'al_ext_posts', '= 1.0.1'
-    gem 'al_img_tools', '= 1.0.2'
-    gem 'al_search', '= 1.0.2'
+    gem 'al_ext_posts', '= 1.0.3'
+    gem 'al_img_tools', '= 1.0.3'
+    gem 'al_search', '= 1.0.3'
     gem 'al_charts', '= 1.0.1'
-    gem 'al_math', '= 1.0.1'
+    gem 'al_math', '= 1.0.2'
     gem 'al_comments', '= 1.0.0'
     gem 'al_newsletter', '= 1.0.0'
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -23,7 +23,7 @@ GEM
       tzinfo (~> 2.0, >= 2.0.5)
     addressable (2.9.0)
       public_suffix (>= 2.0.2, < 8.0)
-    al_analytics (1.0.0)
+    al_analytics (1.0.1)
       jekyll (>= 3.9, < 5.0)
       liquid (>= 4.0, < 6.0)
     al_charts (1.0.1)
@@ -39,7 +39,7 @@ GEM
     al_cookie (1.0.0)
       jekyll (>= 3.9, < 5.0)
       liquid (>= 4.0, < 6.0)
-    al_ext_posts (1.0.1)
+    al_ext_posts (1.0.3)
       feedjira (>= 3.2, < 5.0)
       httparty (>= 0.18, < 1.0)
       jekyll (>= 3.9, < 5.0)
@@ -47,13 +47,13 @@ GEM
     al_folio_bootstrap_compat (1.0.0)
       jekyll (>= 3.9, < 5.0)
       liquid (>= 4.0, < 6.0)
-    al_folio_core (1.0.11)
+    al_folio_core (1.0.12)
       jekyll (>= 3.9, < 5.0)
       liquid (>= 4.0, < 6.0)
-    al_folio_cv (1.0.0)
+    al_folio_cv (1.0.1)
       jekyll (>= 3.9, < 5.0)
       liquid (>= 4.0, < 6.0)
-    al_folio_distill (1.0.2)
+    al_folio_distill (1.0.3)
       jekyll (>= 3.9, < 5.0)
       liquid (>= 4.0, < 6.0)
     al_folio_upgrade (1.0.3)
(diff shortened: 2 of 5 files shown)


alshedivat:
⚠️ **A second change must land with the `al_folio_distill 1.0.3` pin bump — deliberately *not* included here, because it would break CI today.**

Running the starter's own integration suite against every plugin's merged `main` turned up:

```
$ bash test/integration_distill.sh
unexpected distill transforms runtime hash: 5d85590f5652b910ab2411019749c83ef5a5a3fbb9b739adc92b4557b6bf3d39
exit=1
```

`test/integration_distill.sh:48` pins the vendored runtime by hash:

```bash
expected_transforms_hash="70e3f488e23ec379d33a10a60311ec60b570b3b2d5f1823e9159f661c315184e"
```

al-org-dev/al-folio-distill#3 **patched** `transforms.v2.js` — that's the whole point of it, removing the code that re-injected `https://distill.pub/template.v2.js` — so the file's hash necessarily changed:

| file | 1.0.2 (pinned today) | 1.0.3 (merged) |
| --- | --- | --- |
| `transforms.v2.js` | `70e3f488…c315184e` | `5d85590f…b6bf3d39` |
| `template.v2.js` | unchanged | `4790831c…132e271a` |

So **publishing `al_folio_distill 1.0.3` breaks `unit-tests.yml` on this repo** until that expected hash is updated. It can't be updated in advance either — the pin is still `= 1.0.2`, whose runtime genuinely does hash to `70e3f488…`, so changing it now would fail CI immediately. The two changes are coupled.

**When you publish 1.0.3, land these together:**

```diff
- gem 'al_folio_distill', '= 1.0.2'
+ gem 'al_folio_distill', '= 1.0.3'
```
```diff
- expected_transforms_hash="70e3f488e23ec379d33a10a60311ec60b570b3b2d5f1823e9159f661c315184e"
+ expected_transforms_hash="5d85590f5652b910ab2411019749c83ef5a5a3fbb9b739adc92b4557b6bf3d39"
```

The hardcoded hash is doing its job here — it's an independent pin held by the starter rather than something the gem attests about itself, so it correctly refused to accept a runtime it hadn't been told about. Worth keeping that way; it just needs updating on each distill release. Happy to open that PR the moment 1.0.3 is on RubyGems.

---

alshedivat:
**Correction to the PR description, and I've moved this to draft.**

I wrote above that merging this now was safe and would only surface a transient build *warning*. That was wrong. `upgrade-check` fails on it:

```
Upgrade audit complete. Blocking: 1, Non-blocking: 6.
Process completed with exit code 1.
```

Blocking finding — `distill_remote_loader_enabled`:

> Distill runtime still references remote template loader while allow_remote_loader is false.

triggered by `assets/js/distillpub/transforms.v2.js:14067` inside the installed gem:

```js
loaderTag.src = 'https://distill.pub/template.v2.js';
```

**Why flipping the flag surfaced it.** From `al_folio_upgrade`'s `check_distill_runtime` (`cli.rb:365`):

```ruby
allow_remote_loader = parsed.dig("al_folio", "distill", "allow_remote_loader") == true
return if allow_remote_loader     # flag true => check skipped entirely
```

While the flag was `true` the audit **short-circuited and never scanned**. Setting it to `false` activates the scan, which greps the *installed* runtime — and `Gemfile:46` still pins `al_folio_distill '= 1.0.2'`, whose vendored `transforms.v2.js` genuinely does still hardcode the remote URL.

So the audit is right, and this is not a defect in the config change: it's correctly reporting that **the config now forbids the remote loader while the pinned gem still ships it**. Worth noting the tool did exactly its job — and that the finding was invisible until now only because the flag was suppressing its own check.

**Consequence for release sequencing.** This is a stronger constraint than I described. The config flip cannot land on its own at all; it has to go out together with the gem bump. Once `al_folio_distill 1.0.3` is on RubyGems, one PR should carry all three:

```diff
- gem 'al_folio_distill', '= 1.0.2'
+ gem 'al_folio_distill', '= 1.0.3'
```
```diff
-    allow_remote_loader: true
+    allow_remote_loader: false
```
```diff
- expected_transforms_hash="70e3f488e23ec379d33a10a60311ec60b570b3b2d5f1823e9159f661c315184e"
+ expected_transforms_hash="5d85590f5652b910ab2411019749c83ef5a5a3fbb9b739adc92b4557b6bf3d39"
```

Left in draft as the placeholder for that change. I'll complete and un-draft it as soon as 1.0.3 is published — or close it in favour of a single combined release PR if you'd prefer that shape.

Everything else on this branch is unaffected: `deploy`, `visual-regression`, `link-checker`, `CodeQL` and the integration tests all pass.

---

PR #3683, opened 2026-07-28
ci: keep auto-generated TOCs Prettier-clean

alshedivat:
**`prettier.yml` is currently red on `main`.** This fixes it, and the underlying conflict that caused it.

## What happened

#3681 fixed the `update-tocs.yml` `paths`/`files` globs so they finally match `docs/*.md` — that was the right fix, since every TOC under `docs/` had been silently stale. The knock-on: the generator now rewrites those files, and it escapes markdown link text differently than Prettier does.

`gh-md-toc` emits:

```
- [2. Gemfile and _config.yml are two lists that must agree](#...)
```

Prettier requires the escaped underscore:

```
- [2. Gemfile and \_config.yml are two lists that must agree](#...)
```

So the auto-commit `ff4c43c6 "Auto update markdown TOC"` landed unformatted, and Prettier has been failing on `main` ever since — on `docs/ARCHITECTURE.md`, `docs/CUSTOMIZE.md` and `docs/FAQ.md`.

I confirmed this is pre-existing and not caused by any open PR: stashing all local changes and running `npm run lint:prettier` against clean `origin/main` reproduces the same three failures.

## Fix

Run Prettier over the changed markdown **after** the TOC pass and **before** the auto-commit, so the two tools can't disagree about the same text again. Prettier gets the last word. Also adds `actions/setup-node`, which the new step needs.

Plus the one-time reformat of the three files that are currently failing.

## Verified

- `npm run lint:prettier` → *All matched files use Prettier code style!* (was: 3 files failing)
- `npm run lint:style-contract` → passed
- `update-tocs.yml` parses; step order is checkout → setup-node → changed-files → TOC → **normalize** → commit

Worth noting this class of bug is self-inflicting by design: a workflow that commits generated content will fight any formatter that also owns that content, and the failure only shows up on `main` after the merge — never on the PR that enabled it.

---

### DIFF
--- a/.github/workflows/update-tocs.yml
+++ b/.github/workflows/update-tocs.yml
@@ -56,7 +56,12 @@ jobs:
         env:
           ALL_CHANGED_FILES: ${{ steps.changed-files.outputs.all_changed_files }}
         run: |
-          npm install --save-dev --save-exact prettier @shopify/prettier-plugin-liquid
+          # `npm ci` installs from the lockfile and does not touch package.json or
+          # package-lock.json. `npm install --save-dev` would rewrite the caret
+          # ranges to exact pins and upgrade them, and because the auto-commit step
+          # below sets no file_pattern (default `.`), those dependency changes would
+          # be committed to main under the message "Auto update markdown TOC".
+          npm ci
           for file in ${ALL_CHANGED_FILES}; do
             if [[ "$file" != *.md ]]; then
               continue
--- a/.github/workflows/update-tocs.yml
+++ b/.github/workflows/update-tocs.yml
@@ -23,6 +23,8 @@ jobs:
         with:
           fetch-depth: 0 # OR "2" -> To retrieve the preceding commit.
 
+      - uses: actions/setup-node@v4
+
       - name: Get changed files
         id: changed-files
         uses: tj-actions/changed-files@ed68ef82c095e0d48ec87eccea555d944a631a4c # v46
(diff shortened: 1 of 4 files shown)


alshedivat:
Re: the Codex review on `update-tocs.yml:59` — right, and it would have bitten on the very first TOC refresh. Fixed in `102ccb70`.

I'd copied the install line from `prettier.yml`, where it's harmless because that job never commits. Here it is not: this job commits via `git-auto-commit-action`, which sets no `file_pattern` and so defaults to staging everything.

Reproduced on a clean tree:

```
$ npm install --save-dev --save-exact prettier @shopify/prettier-plugin-liquid
$ git status --porcelain -- package.json package-lock.json
 M package-lock.json
 M package.json

"@shopify/prettier-plugin-liquid": "^1.10.0"  ->  "1.11.0"
"prettier":                        "^3.8.0"   ->  "3.9.6"
```

So every TOC refresh would have silently dropped both caret ranges **and upgraded both packages**, committed to `main` under the message "Auto update markdown TOC". Exactly the kind of change that's invisible until something breaks.

`npm ci` installs from the lockfile and leaves both manifests untouched — verified on the same tree, with `prettier 3.8.0` still resolving as pinned:

```
$ npm ci && git status --porcelain -- package.json package-lock.json
  (clean)
$ npx prettier --version
3.8.0
```

**One related hazard I did not change,** since it predates this PR and I'd rather flag it than alter untested behaviour: that auto-commit step has no `file_pattern` at all, so *any* stray file the runner leaves behind gets committed to `main`. Scoping it to `*.md docs/*.md` would close that off permanently. Happy to do it in a follow-up if you want — I left it alone here because a wrong pattern fails by silently committing nothing, which would break TOC updates without any signal.

---

alshedivat:
Closing — **superseded by #3684**, which carried this branch's commits and has now merged.

Verified rather than assumed. Comparing blobs against current `main`:

```
IDENTICAL  .github/workflows/update-tocs.yml
IDENTICAL  docs/ARCHITECTURE.md
IDENTICAL  docs/FAQ.md
DIFFERS    docs/CUSTOMIZE.md          (main has extra repo-card doc edits from #3680)
```

and a trial merge into `main` stages nothing:

```
$ git merge --no-commit --no-ff claude/fix-toc-prettier-conflict-xrj39u
Auto-merging docs/CUSTOMIZE.md
$ git diff --cached --quiet   # NO-OP: main already contains everything this adds
```

`main` is confirmed green: `npm ci` is in place in `update-tocs.yml`, and `npm run lint:prettier` reports *All matched files use Prettier code style!*

Worth recording why this mattered, since the symptom recurred twice before it landed:

```
ffee0cd6  feat: self-host the star history chart (#3684)   <- carried the fix
39562420  Auto update markdown TOC                          <- re-broke prettier
38af3ca2  fix: point repo cards ... (#3680)
ff4c43c6  Auto update markdown TOC                          <- broke it first
5678c07a  docs: ... (#3681)
```

The TOC bot fires on every markdown-touching merge, so this wasn't a one-time cleanup — `main` flipped red after each one. It should stay green now that Prettier runs before the auto-commit.

The one hazard I flagged and deliberately left alone is still open: that auto-commit step has no `file_pattern`, so any stray file a runner leaves behind gets committed to `main`. Happy to scope it to `*.md docs/*.md` in a small follow-up if you want it closed off.

---

PR #3684, opened 2026-07-28
feat: self-host the star history chart

alshedivat:
Replaces the dead `api.star-history.com` hotlink with a chart generated in-repo and refreshed automatically.

## Why

**1. The service is down for everyone.**

```
$ curl "https://api.star-history.com/svg?repos=alshedivat/al-folio&type=Date"
HTTP 500 — Repo alshedivat/al-folio: timeout of 10000ms exceeded
```

5/5 attempts failed, and it reproduces on unrelated repos including `octocat/Hello-World` (2 stars) — so it's a service-wide backend failure, not our repo or its size. Their website root returns 200; specifically the SVG endpoint is broken.

**2. It now asks visitors for a token that can push code.**

star-history.com prompts for a GitHub token with **`Contents: read and write`**, claiming "the stargazers endpoint now requires contents write". That claim is false. Unauthenticated, no headers beyond `Accept`:

```
GET /repos/alshedivat/al-folio/stargazers            -> 200
     Accept: application/vnd.github.star+json        -> 200, with starred_at timestamps
GET /repos/alshedivat/al-folio                       -> stargazers_count = 15931
```

`Contents: write` is the permission to **push code**. For a template this widely forked, that's not a dependency worth keeping even if the backend recovers.

## What this does

- **`bin/generate_star_history.py`** — fetches the public stargazer timestamps and renders light + dark SVGs. No third-party services, no runtime dependencies, stdlib only.
  - Samples ~28 pages across the history instead of fetching all ~160, so a refresh is ~29 requests.
  - Anchors the final point at the **most recent star**, not "now", so repeated runs are byte-identical rather than producing a daily no-op commit. Verified: two consecutive runs → same md5.
- **`.github/workflows/star-history.yml`** — regenerates on every push to `main`, weekly on a schedule, or on demand; commits only when the chart actually changed.
- **`README.md`** — serves the committed assets via the existing `<picture>` + `prefers-color-scheme` pattern, so it can't 500 and can't prompt anyone for anything. Links to the stargazers page instead of star-history.com.

## Loop and blast-radius guards

Two things I was careful about, since this workflow commits to the branch that triggers it:

- **No self-retrigger.** `paths-ignore` on the generated SVGs means the bot's own commit — which touches nothing else — can't retrigger the workflow. (GitHub also declines to raise `push` events for commits made with `GITHUB_TOKEN`, so there are two independent guards.) Plus `concurrency` so two runs can't race to commit.
- **Doesn't leak to template users.** Gated on `github.repository == 'alshedivat/al-folio'`, matching the existing convention in `broken-links.yml`. Without it, every site generated from this template would spend Actions minutes charting someone else's repo and committing files they never asked for.

`GITHUB_TOKEN` is passed only to lift the API rate limit from 60/hr to 5000/hr — the data itself is public, and no personal or scoped token is involved.

## Result

15,931 stars, 29 sampled points, 2016-09-21 → 2026-07-28, styled with al-folio's own palette (`#b509ac`, and a lightened `#d94fd0` on dark for contrast against `#1c1c1d`).

## Verified

- Rendered both SVGs in headless Chromium and reviewed them; fixed one real defect found that way — year ticks were derived from sampled data points, so the sparse early years never appeared and the axis jumped 2016 → 2020. Ticks are now calendar-based.
- Both files parse as valid XML.
- `npm run lint:style-contract` → passed.
- `npm run lint:prettier` → the repo currently fails on three unrelated docs files; that is **pre-existing on `main`** and fixed separately in #3683. This branch adds no new Prettier failures.

## Note

`main` is protected, so if the Actions bot can't push directly the workflow will fail loudly rather than silently. If you'd rather it opened a PR per refresh instead of committing, say so and I'll switch it.

---


alshedivat:
The Prettier failure here was **inherited from `main`**, not introduced by this PR — and it's now resolved on this branch.

Verified rather than assumed. The three failing files are:

```
docs/ARCHITECTURE.md
docs/CUSTOMIZE.md
docs/FAQ.md
```

and the six files this branch touches are:

```
.github/workflows/star-history.yml
.gitignore
README.md
assets/img/star-history-dark.svg
assets/img/star-history-light.svg
bin/generate_star_history.py
```

Zero overlap. Stashing all local changes and running `npm run lint:prettier` against clean `origin/main` reproduces the same three failures, so `main` itself is red — cause and fix are described in #3683.

Rather than make you merge in a specific order, I've merged the #3683 branch into this one, so **both PRs are now independently green and can be merged in either order**. If #3683 lands first, its commit here simply becomes redundant — no conflict.

Current state on this branch:

```
npm run lint:prettier        All matched files use Prettier code style!
npm run lint:style-contract  Starter style contract check passed.
```

---

alshedivat:
Re: the Codex review on `bin/generate_star_history.py:126` — correct, and a worse bug than it looks. Fixed in `9d868b40`.

Latent for this repo (15,931 stars ≈ 160 pages, well under the 400-page cap), but the script takes `--repo`, so it's real for anyone pointing it elsewhere. Tested deterministically by simulating an 80,000-star repository offline:

```
total=80,000  capped=True
last OBSERVABLE star : 2019-07-25  (star #40,000)
final plotted point  : 2026-07-28  count=80,000

  [old bug] would back-date 80,000 onto 2019-07-25 -> vertical cliff, x-axis frozen
  [fixed  ] final point dated today: True
  [fixed  ] subtitle discloses truncation: True
```

The old logic would have placed all 80,000 stars on **2019-07-25** — a seven-year error, with the curve jumping vertically and the axis never advancing past the cap.

**Fix.** When the cap is hit, anchor the final point at today rather than back-dating it. The current total on today's date is a fact; the segment between it and the last observed star is interpolated — the same approximation already used between every other sampled point, since the chart samples ~28 pages rather than fetching all ~160. The difference is it's now *disclosed* instead of presented as measurement; the subtitle reads:

> `owner/repo · beyond 40,000 stars the API stops paginating, so the tail is interpolated`

I considered reaching past the cap instead, but REST `/stargazers` has no reverse ordering and the 400-page limit is absolute — that would mean a GraphQL path with `orderBy: {field: STARRED_AT, direction: DESC}` and a mandatory token. Not worth the complexity for a case that doesn't arise here; being honest about the gap is the better trade.

Dates truncated to midnight so reruns within a day stay byte-identical and don't churn commits.

**No regression:** regenerating this repo's chart after the change produces byte-identical SVGs to the committed ones.

---

Mubelotix:
Hello, I forked star-history.com into https://star-history.dera.page, it provides the same experience as the OG service, no token required

